PR #165, opened 2015-01-13
v4l2decoder: do not wait on POLLOUT event

xuguangxin:
else it will wakeup by it self. Make current thread busy waiting.


halleyzhao:
+1
good finding

PR #173, opened 2015-02-03
Update Libyami for VP9 decode playback on Chrome.

skibey:
Fix resolution change handling when running VP9 on
Chrome OS with Libyami. Ported similar change from
VP8 to VP9. After these changes VP9 works fine
on Chrome OS with Libyami.

Signed-off-by: Sameer Kibey <sameer.kibey@intel.com>

### DIFF
--- a/decoder/vaapidecoder_base.cpp
+++ b/decoder/vaapidecoder_base.cpp
@@ -423,8 +423,6 @@ Decode_Status VaapiDecoderBase::terminateVA(void)
     DEBUG("surface pool is reset");
     m_context.reset();
     m_display.reset();
-    m_externalDisplay.type = NATIVE_DISPLAY_AUTO;
-    m_externalDisplay.handle = 0;
 
     m_VAStarted = false;
     return DECODE_SUCCESS;
--- a/decoder/vaapidecoder_base.cpp
+++ b/decoder/vaapidecoder_base.cpp
@@ -327,6 +327,7 @@ Decode_Status
     VaapiSurface *suf;
     Decode_Status status;
     VAStatus vaStatus = VA_STATUS_SUCCESS;
+    const char* driver_name = NULL;
 
     if (m_enableNativeBuffersFlag == true) {
         numSurface = 20;        //NATIVE_WINDOW_COUNT;
@@ -342,10 +343,22 @@ Decode_Status
     }
 
 #if __PLATFORM_BYT__
+    driver_name = "wrapper";
     if (setenv("LIBVA_DRIVER_NAME", "wrapper", 1) == 0) {
         INFO("setting LIBVA_DRIVER_NAME to wrapper for chromeos");
     }
 #endif
+
+    if (profile == VAProfileVP9Version0) {
+       driver_name = "hybrid";
+       if (setenv("LIBVA_DRIVER_NAME", "hybrid", 1) == 0) {
+           INFO("setting LIBVA_DRIVER_NAME to hybrid for VP9 profile");
+       }
+    }
+    else {
+           INFO("profile: %d, will use legacy i965 driver", profile);
+    }
+
     m_display = VaapiDisplay::create(m_externalDisplay);
 
     if (!m_display) {
@@ -392,6 +405,11 @@ Decode_Status
     if (!(m_configBuffer.flag & USE_NATIVE_GRAPHIC_BUFFER)) {
         m_videoFormatInfo.surfaceWidth = m_videoFormatInfo.width;
         m_videoFormatInfo.surfaceHeight = m_videoFormatInfo.height;
+    }
+
+    // unset the env variable here
+    if (driver_name) {
+       unsetenv("LIBVA_DRIVER_NAME");
     }
 
     m_VAStarted = true;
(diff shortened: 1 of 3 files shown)


decoder/vaapidecoder_vp9.cpp
```diff
@@ -90,23 +91,50 @@ void VaapiDecoderVP9::flush(void)
...
     // only reset va context when there is a larger frame
     if (m_configBuffer.width < hdr->width
         || m_configBuffer.height <  hdr->height) {
+        resetContext = true;
         INFO("frame size changed, reconfig codec. orig size %d x %d, new size: %d x %d",
                 m_configBuffer.width, m_configBuffer.height, hdr->width, hdr->height);
-        Decode_Status status = VaapiDecoderBase::terminateVA();
+        
+        if (m_hasContext){
+            status = VaapiDecoderBase::terminateVA();
```
chivakker: Is there a better way to do this?  I think we don't need to terminateVA but instead re-allocate surfaces should be enough.  If Vaapi has to be terminated it should be on fatal errors only.

skibey: Guangxin had sent me another email before on this. As per his email - VaapiDecoderBase::terminateVA(void) is not the same as vaTerminate. VaapiDecoderBase::terminateVA(void) will clear the surface pool in VaapiDecoderBase and reset the context. So Guangxin has a separate patch that will remove reset of the external display in the terminateVA. Then we will not lose display handle on resolution change. 

Guangxin, do you want to submit your patch separately to address Daniel's concern? What is your comment on this? Thanks.

seanvk: Okay, that makes sense.  Can we test with Guangxin's patch with yours ?  Or have you already done that?

xuguangxin: yes, this patch only fix "display lost" at beginning of playback. If the bitstream have resolution change in the middle of playback. It still lost display. So i proposed other patch to sameer. I planed to test this patch after we get my image ready. But I stucked at hang issue. So it's great if sammer can help me test this patch and push to this thread.It can help us move faster.

decoder/vaapidecoder_base.cpp
```diff
@@ -346,6 +346,16 @@ Decode_Status
         INFO("setting LIBVA_DRIVER_NAME to wrapper for chromeos");
     }
 #endif
+
+    if (profile == VAProfileVP9Version0) {
```
xuguangxin: seems not good.Nobody clean the env, it cause problem if you play vp9 and then want play vp8.  And if you want them play at same time, it will have issue too.We need find better way.

decoder/vaapidecoder_vp9.cpp
skibey: okay, I will include Guangxin's patch with mine to take care of the lost display handle.

decoder/vaapidecoder_base.cpp
skibey: Okay I will investigate more.

decoder/vaapidecoder_vp9.cpp
skibey: I uploaded patch from Guangxin which addresses Daniel's comment about vaTerminate.

decoder/vaapidecoder_base.cpp
skibey: Playing VP9 and VP8 one after the other works fine. But playing VP8 and VP9 at the same time does not work. The reason is that for first stream, the vaDisplayContext stores the display context of i965 driver in vaDisplayContext->vaDriverContext. For the second stream, it will again use the context of i965 driver, even though the driver name is set as hybrid. Thus it will return 'profile not supported' error during vaCreateConfig(). I will file a JIRA on this issue, We need to figure out another way to solve this.

decoder/vaapidecoder_vp9.cpp
xuguangxin: There is no need check you have context or not. we use smart pointer. reset context it in terminateVA directly is not harmful,No matter you have context or not. I have create a pull request #177 to simply this patch. please help review it.
 
Decode_Status VaapiDecoderBase::terminateVA(void)
{
    INFO("base: terminate VA");
    m_surfacePool.reset();
    DEBUG("surface pool is reset");
    m_context.reset();
    m_display.reset();

    m_VAStarted = false;
    return DECODE_SUCCESS;
}

decoder/vaapidecoder_vp9.cpp
```diff
@@ -90,23 +91,50 @@ void VaapiDecoderVP9::flush(void)
...
+        }
+        m_hasContext = false;
         if (status != DECODE_SUCCESS)
             return status;
         m_configBuffer.width = hdr->width;
         m_configBuffer.height = hdr->height;
         m_configBuffer.surfaceWidth = hdr->width;
         m_configBuffer.surfaceHeight = hdr->height;
+        DEBUG("USE_NATIVE_GRAPHIC_BUFFER: %d",
```
xuguangxin: USE_NATIVE_GRAPHIC_BUFFER  is inherit from libmix, we will drop it latter. No need set this here. see #177 for details.

YuJiankang:
+1, looks good to me

chivakker:
Is there a conclusion on this patch?  What is the next step?

xuguangxin:
3 patches here. They are merged or rejected. I will close this if sameer agree.

details:
1. fix resolution changed bug (merge as b2661283302cfd1463509df6ebf5f8b5e854508e with some modify)
2. dynamic switch h264 and vp9 driver( reject since it may have problem when you play them in same time)
3. do not reset external display after resolution change(merged as 434eafd476f6ab732197083e6b2cff99a90106a4).

skibey:
I am okay with closing this.

xuguangxin:
thanks

PR #177, opened 2015-02-09
handle vp9 resolution change

xuguangxin:
1. do not reset the external display in VaapiDecoderBase::terminateVA. else it will lost display when we change video size to larger resolution.
2. report every frame change to caller. Not just silent set the output format in getOutput.


xuguangxin:
Just for clarify. I have test it on chromeos. It's work well

YuJiankang:
+1, looks good to me

Issue #175, opened 2015-02-05
lack of const


waiter89:
in  VideoDecoder class
    virtual const VideoRenderBuffer* getOutput(bool draining = false) = 0;
it return "const " value,but
    virtual void renderDone(VideoRenderBuffer * renderBuf);
has no "const".

YuJiankang:
This issue is fixed by the the following commit, patch is merged to trunk code.
bc55f96e12e758f70f84e63eeed7ef2f278f43c0

xuguangxin:
thanks waiter89 & Jiankang

PR #179, opened 2015-02-10
Fix interface mismatch issue #175

YuJiankang:
fix the issue of lack const


xuguangxin:
Do we need change VaapiDecSurfacePool::recycle(VideoRenderBuffer * renderBuf) to const?

YuJiankang:
yes, we need change it too, I will update the commit, thanks

YuJiankang:
Changed VaapiDecSurfacePool::recycle(VideoRenderBuffer * renderBuf) to const. please help review the new commit, thanks

xuguangxin:
+1,and please push and reply #175, thanks.

PR #185, opened 2015-02-13
fix decodeoutput.cpp for the outputfile can output to file and directory

luokunrock:


tests/decodeoutput.cpp
```diff
@@ -72,8 +73,7 @@ Decode_Status DecodeOutputNull::renderOneFrame(bool drain)
     m_frame.fourcc = 0;
 
     Decode_Status status = m_decoder->getOutput(&m_frame, drain);
-    if (status == RENDER_SUCCESS)
```
xuguangxin: Why you need change this? mistake?

tests/decodeoutput.cpp
```diff
@@ -239,7 +247,7 @@ bool DecodeOutputFileDump::render(VideoFrameRawData* frame)
     if (!m_fp)
         return false;
 
-    //ASSERT(m_width == frame->width && m_height == frame->height);
```
xuguangxin: why you need change this?

tests/decodeoutput.cpp
```diff
@@ -206,10 +206,18 @@ bool DecodeOutputFileDump::config(const char* dir, const char* source, const cha
         if (s)
             baseFileName = s+1;
         assert(dir);
-        m_name<<dir<<"/"<<baseFileName;
-        m_appendSize = true;
+        struct stat buf;
+        int r = stat(dir, &buf);
+        if (r == 0 && buf.st_mode & S_IFDIR){
```
xuguangxin: why we need this judgement? source file's parent must be a directory, right?

xuguangxin:
modified, resubmit, and merged as 826274a08864385ef1f645f91328a0a6ba24136c

PR #191, opened 2015-03-05
seek or close the video window in chrome will messup the graphic system.

xuguangxin:
v4l2dec: do not recycle buffers after seek
        v4l2 client in chrome still hold the buffers after seek

### DIFF
--- a/v4l2/v4l2_decode.cpp
+++ b/v4l2/v4l2_decode.cpp
@@ -50,11 +50,7 @@ V4l2Decoder::V4l2Decoder()
     m_bufferPlaneCount[OUTPUT] = 2;
 
     m_maxBufferCount[INPUT] = 8;
-#if __ENABLE_V4L2_GLX__
-    m_maxBufferCount[OUTPUT] = 8; // Pixmap is alloc at client side, in order to save memory we have to tell them the reality. [it requires chrome change as well]
-#else
-    m_maxBufferCount[OUTPUT] = 26; // use a bigger # to satify chrome, the effective count is m_actualOutBufferCount
-#endif
+    m_maxBufferCount[OUTPUT] = 8;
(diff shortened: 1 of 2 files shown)


v4l2/v4l2_decode.cpp
```diff
@@ -50,11 +50,7 @@ V4l2Decoder::V4l2Decoder()
     m_bufferPlaneCount[OUTPUT] = 2;
 
     m_maxBufferCount[INPUT] = 8;
-#if __ENABLE_V4L2_GLX__
-    m_maxBufferCount[OUTPUT] = 8; // Pixmap is alloc at client side, in order to save memory we have to tell them the reality. [it requires chrome change as well]
-#else
-    m_maxBufferCount[OUTPUT] = 26; // use a bigger # to satify chrome, the effective count is m_actualOutBufferCount
-#endif
+    m_maxBufferCount[OUTPUT] = 8;
```
chivakker: Can this value be taken from libyami Client?  In the case I am working on, they match but it would be good to avoid the hardcoded value.

chivakker:
lgtm +1, I just wanted to point out the need to have the client tell about the correct value.

xuguangxin:
Yes.Current solution have 2 drawbacks.
1. it need dma buffer related api.
2. imagepool(inside decoder) manages VaImage life cycle. and v4l2 client manages egl image life cycle. But they are same thing, just bind together. It's hard to sync this two life cycle manager. Close, seek, open failed bug all related to this.
Best way is let client allocate and manage the egl image. we just blit nv12 to those egl image.
We will implement this later.But it need time. Currently, we better make old code work first.

chivakker:
Ok, thanks for the clarification.

I am still +1 lgtm.

YuJiankang:
+1

PR #197, opened 2015-03-20
vpp interface for review.

xuguangxin:
YAMI_VPP_SCALER  will do color space conversion and scale.


xuguangxin:
close since we have #202

PR #198, opened 2015-03-25
vaapiimagepool: Added a fix to properly release the Vaapi Image pool on VA terminate

mdnavare:
This fixes the memory leak issue during the repetitive
VP8 decode in the same process.

Signed-off-by: Manasi <manasi.d.navare@intel.com>

### DIFF
--- a/vaapi/vaapiimagepool.cpp
+++ b/vaapi/vaapiimagepool.cpp
@@ -69,6 +69,7 @@ struct VaapiImagePool::ImageRecycler
         if (!image)
             return;
         m_pool->recycleID(image->getID());
+        m_pool.reset();
     }
 private:
     ImagePoolPtr m_pool;


chivakker:
lgtm

Reviewed-by "Daniel Charles <daniel.charles@intel.com>"
Verified-by "Daniel Charles <daniel.charles@intel.com>"

vaapi/vaapiimagepool.cpp
```diff
@@ -69,6 +69,7 @@ struct VaapiImagePool::ImageRecycler
         if (!image)
             return;
         m_pool->recycleID(image->getID());
+        m_pool.reset();
```
xuguangxin: Are you sure this will help you ? Vfter void operator()(VaapiImage* image) called, the ImageRecycler will out of scope.It will reset m_pool automatically. It should have same result no matter you reset here or not.

mdnavare: We want to reset the pool for each ImageRecycler call so that when it destroys each Surface, it also releases the corresponding m_pool. Without the reset call, the use_count remains the same each time ImageRecycler is called.

xuguangxin: From my understanding, it should release the use_count automatically after void operator()(VaapiImage* image) called. But it's not meet your observation. Anyway, I will spend half day to dig more. If I did not find more useful things. I will accept your patch. Is it ok to you?

xuguangxin: Hi mdnavare:
This patch can fix VaapiDisplay leak problem, but i still see the random failed.Can you test more times?
I find the reason of this leak, I will submit other patch and explain details in it

chivakker:
use #199 instead

xuguangxin:
thanks to mdnavare and chivakker.

PR #199, opened 2015-03-26
image: do not inherit from enable_shared_from_this

xuguangxin:
wired thing happend when we do two things together
1. make VaapiImage inherit from shared_from_this (if you make map as VaapiImage's member, you need this)
2. set ImagePtr's deleter in VaapiImagePool. and hold a renference of ImagePoolPtr in ImageRecycler's member.
It will call deleter's operator(), but the deleter will never out of scope. So ImagePoolPtr will never get chance to be del.Make map a standalone function can break 1.
This patch bigger than mdnavare's #198, But it will make ImageRecycler out of scope in somewhere.
This patch have same result as mdnavare's . It fix VaapiDisplay leak problem.But it still not fix open file failed issue.


chivakker:
lgtm

PR #200, opened 2015-03-27
getOutput: while flushing some images might not be available

chivakker:
some images will not be available when flushing before
shutdown and that should not be asserted.


xuguangxin:
+1

PR #204, opened 2015-04-10
yamidecode: compile tests with X when supported

chivakker:
test application use X only when available

### DIFF
--- a/tests/v4l2decode.cpp
+++ b/tests/v4l2decode.cpp
@@ -278,7 +278,9 @@ int main(int argc, char** argv)
     renderMode = 3; // set default render mode to 3
 
     yamiTraceInit();
+#if __ENABLE_V4L2_GLX__
     XInitThreads();
+#endif
 
     if (!process_cmdline(argc, argv))
         return -1;
@@ -321,9 +323,9 @@ int main(int argc, char** argv)
     fd = YamiV4L2_Open("decoder", 0);
     ASSERT(fd!=-1);
 
+#if __ENABLE_V4L2_GLX__
     x11Display = XOpenDisplay(NULL);
     ASSERT(x11Display);
-#if __ENABLE_V4L2_GLX__
     ioctlRet = YamiV4L2_SetXDisplay(fd, x11Display);
 #endif
     // set output frame memory type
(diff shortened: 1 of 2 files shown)


tests/v4l2decode.cpp
```diff
@@ -321,9 +323,9 @@ int main(int argc, char** argv)
     fd = YamiV4L2_Open("decoder", 0);
     ASSERT(fd!=-1);
 
+#if __ENABLE_V4L2_GLX__
     x11Display = XOpenDisplay(NULL);
```
xuguangxin: eglInit need this to call eglGetDisplay. it will break VIDEO_DATA_MEMORY_TYPE_DMA_BUF & VIDEO_DATA_MEMORY_TYPE_DRM_NAME in no glx mode.

chivakker: On Apr 9, 2015 5:54 PM, "Xu Guangxin" <notifications@github.com> wrote:
>
> In tests/v4l2decode.cpp:
>
> > @@ -321,9 +323,9 @@ int main(int argc, char** argv)
> >      fd = YamiV4L2_Open("decoder", 0);
> >      ASSERT(fd!=-1);
> >
> > +#if __ENABLE_V4L2_GLX__
> >      x11Display = XOpenDisplay(NULL);
>
> eglInit need this to call eglGetDisplay. it will break
VIDEO_DATA_MEMORY_TYPE_DMA_BUF & VIDEO_DATA_MEMORY_TYPE_DRM_NAME in no glx
mode.
>

libyami will operate in auto mode which will do drm in the absence of X11

xuguangxin: but eglGetDisplay will failed, right? glGenTextures all dependent on eglDisplay, we'd better get eglDisplay from zone and eglInitialize it.

chivakker: On Apr 9, 2015 7:09 PM, "Xu Guangxin" <notifications@github.com> wrote:
>
> In tests/v4l2decode.cpp:
>
> > @@ -321,9 +323,9 @@ int main(int argc, char** argv)
> >      fd = YamiV4L2_Open("decoder", 0);
> >      ASSERT(fd!=-1);
> >
> > +#if __ENABLE_V4L2_GLX__
> >      x11Display = XOpenDisplay(NULL);
>
> but eglGetDisplay will failed, right? glGenTextures all dependent on
eglDisplay, we'd better get eglDisplay from zone and eglInitialize it.
>

It is not failing as NATIVE_DISPLAY_AUTO is in use and eglGetDisplay will
create a Libyami drmfd. This is specific to the test application and
libyami.

xuguangxin:
ok, +1

xuguangxin:
merged as 456207affc9ecfe17127bd24630a35a00727f500

Issue #206, opened 2015-04-13
vaPutSurface flags


jsorg71:
We need a way to pass a value to the last parameter of vaPutSurface.
Currently, you always hard code a zero.
See VaapiDecoderBase::getOutput.
I can do a pull request but I want to know how you prefer to do this.
Options include.

1 - add a parameter to the function calls ie VaapiDecoderBase::getOutput.  This will touch many files.
2 - add a class member to "class VaapiDecoderBase"  Someting like
public:
  unsigned int m_putSurfaceFlags;
3 - some other idea I didn't think about, let me know.

Currently, I prefer option 2.

Jay

xuguangxin:
Hi Jay:
which flag you want use, Is it gone dynamic change?

jsorg71:
VA_SRC_BT601 and VA_SRC_BT709
RemoteFX H264 uses VA_SRC_BT709 which is not default.
I currently have to patch this and ship a modified libyami.
I guess it does not need to be dynamic but I can see the possible need.

xuguangxin:
If you use option 2, you still need a function to set the m_putSurfaceFlags. and this function is not needed by other getOutput. 
Maybe, option 1 is better. you do not need modify too many file since you can use default param.
just like  frameX.

By The Way, in middle term. We want getOutput only output VaSurface(with some wrapper). You can do what you want when you get the surfacd id.

jsorg71:
Ok, I'll create a PR.

xuguangxin:
thanks

xuguangxin:
Hi Jay:
we have new interface SharedPtr<VideoFrame>IVideoDecoder::getOutput(), VideoFrame.surface is a VASurface. So you can use it do any thing you want. You can see https://github.com/01org/libyami/blob/master/examples/simpleplayer.cpp for usage.

thanks

jsorg71:
@xuguangxin Thanks, I'll check it out.

xuguangxin:
Hi @jsorg71 :
How was this goingon?

thanks

jsorg71:
@xuguangxin 

I'm using the capi.
Here is some sample code how I'm using the decode and vaPutSurface.
This works great and I have access to all the capabilities of vaPutSurface.

Jay

    memset(&ib, 0, sizeof(ib));
    ib.data = cdata;
    ib.size = cdata_bytes;
    status = g_decode(self->dec, &ib);
    if (status == 2) /* DECODE_FORMAT_CHANGE */
    {
[...]

PR #207, opened 2015-04-13
add flags to getOutput method

jsorg71:
Discussed in #206 
This add the ability to pass some flags through libyami for vaPutSurface.

### DIFF
--- a/capi/VideoDecoderCapi.cpp
+++ b/capi/VideoDecoderCapi.cpp
@@ -78,11 +78,11 @@ const VideoRenderBuffer* decodeGetOutput(DecodeHandler p, bool draining)
 #ifdef __ENABLE_X11__
 Decode_Status decodeGetOutput_x11(DecodeHandler p, Drawable draw, int64_t *timeStamp
         , int drawX, int drawY, int drawWidth, int drawHeight, bool draining
-        , int frameX, int frameY, int frameWidth, int frameHeight)
+        , int frameX, int frameY, int frameWidth, int frameHeight, unsigned int flags)
 {
     if(p)
         return ((IVideoDecoder*)p)->getOutput(draw, timeStamp, drawX, drawY, drawWidth, drawHeight, draining
-                        , frameX, frameY, frameWidth, frameHeight);
+                        , frameX, frameY, frameWidth, frameHeight, flags);
     else
         return DECODE_FAIL;
 }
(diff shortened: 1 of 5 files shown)


chivakker:
What is your use case?  shouldn't the de-interlacing flags be set as a parameter to libyami?

jsorg71:
In the referenced issue #206, I explain that sometime BT709 is used on the server.
And yes, since this passes the flags through, you can also use this for de-interlacing flags.

chivakker:
I've checked #206. I also preferred option 2 as it is a cleaner way to handle this such that libva-API is better wrapped within libyami.  

I will let xuguangxin decide if he wants to go forward with this patch instead.

xuguangxin:
hi Daniel:
I think expose more libva is better than wrap it. because following reasons:
1. it's hard to wrap libva. see getOutput for details.
we have 3 getOutput and getOutput(VideoFrameRawData* frame, bool draining = false) have 5  VideoDataMemoryType. it means we have 7 getOuputs. If we want support wayland or android. We need add more getOuput to decoder interface. It's insane. Why not just output VASurfaceID(with some wrapper and reference count, you can see https://github.com/xuguangxin/libyami/commit/f892d1f8faddb2e7964e89a875c438ec87b305e0 for detials), user can do what he want. dump, output to x,output to wayland, It's his call.

2. wrap libva is not making things easier. 
I mean call
    virtual Decode_Status getOutput(unsigned long draw, int64_t *timeStamp
        , int drawX, int drawY, int drawWidth, int drawHeight, bool draining = false
        , int frameX = -1, int frameY = -1, int frameWidth = -1, int frameHeight = -1) = 0; 
is not easier than call getOutput and  vaPutSurface.
You still need prepare all params.

3. wrap libva will lost flexibility.
what if some one still need   cliprects in vaPutSurface. Do we need more params to getOuptut?

I want keep yami decoder module more simpler. Just accept bitstream and output surface. Let other module do vpp,output, and file dump things.  I can talk to you more details about this when we meet face to face.
thanks.

jsorg71:
xuguangxin,

That is perfectly acceptable.
I think you should reject my patch and remove X11 support from libyami.
It's just confusing if there is X11 support but limited.
I can change my code to use libyami for decode and libva for render.

Jay

xuguangxin:
Hi Jay:
Sorry for late response. I am in a business trip. Only have limited access to email. 
Remove X dependency is not gone happen very quickly. Because yami decodeer/encoder and v4l2 decoder/encoder dependent  on this.
Is it ok to merge your patch to master? It can meet your current requirement. And you can change you code when we removed X in libyami?
thanks

jsorg71:
Sure, if it looks useful.

jsorg71:
I have working the va render path and only yami for decode.
I only found a small 'const' missing capi error.  I'll do a pull request for that.
I think we can delete this PR.

PR #208, opened 2015-04-14
fix the errors obtained by valgrind, or it will give the message:

luokunrock:
fix the errors obtained by valgrind, or it will give the message: Conditional jump or move depends on uninitialised value(s)


xuguangxin:
+1

chivakker:
lgtm, +1, please merge

PR #209, opened 2015-04-23
Update decodeoutput.cpp

ratin3:
m_window was uninitialized, thus preventing creation of an xwindow in the function 

bool DecodeOutputX11::setVideoSize(int width, int height)

Which was getting called for a format change in the video upon decoding  the first frame.


xuguangxin:
Hi ratin:
Thanks for patch. Seems Kun have fix this problem in https://github.com/01org/libyami/pull/208, Is it fix your problem?

BR.

ratin3:
yes

xuguangxin:
thanks ratin3

PR #210, opened 2015-05-04
when decode some 264 format files with render mode 2, valgrind reports

luokunrock:
informations that invalide write/read of size.


xuguangxin:
+1

PR #213, opened 2015-05-07
add 'return true' statement, otherwise return false necessarily when …

andyLiuChina:
…calling ensureQMatrix()


xuguangxin:
Hi andy:
the comment better following some rules:
1. better have component name on the begin of comment. You can explain the commit on first line but it should be short. eg:
vp8enc: add return for ensureQMatrix
        add 'return true' statement, otherwise return false necessarily when calling ensureQMatrix()

you can refer to old commit log for detials.

PR #217, opened 2015-05-12
Open a different kernel device for VA display

mdnavare:
A different kernel device is opened for VA display to avoid the same device being used by GBM and LIBVA. 
This change was made by Google as a resolution to Netflix hang issue

Signed-off-by: Manasi <manasi.d.navare@intel.com>


chivakker:
I think this should be handled in a different way, most likely with a configuration variable that will enable it to systems that define the new device.  When cross-compiling the device won't be available so there's no option to test for the existence of the device.

xuguangxin:
Hi manasi:
It's no /dev/dri/renderD128 in ubuntu, right? This will break libyami on ubuntu. Can we find other way to fix this?

seanvk:
This is not a good solution because it will break non-cros environment.  Please revise to evaluate either specific build config for CROS or more selectively query device node availability.  -1, please fix.

PR #220, opened 2015-05-20
Modify five patchs commit log info.

andyLiuChina:
1) vp8enc: add missed "return true" statement
 	2) vp8enc: add missed "return true" statement
 	3) tests: Write frame counts to VP8 file header
 	4) tests: set default fps to 30 for .vp8 container
 	5) codecparsers: add new files for h265 support.


xuguangxin:
+1

PR #224, opened 2015-05-28
fix resolution change failed in egl mode

xuguangxin:
root cause of this problem is we export internal VaapiImage to chrome renderer.When reoslution changed. We destroyed internal image, but it's no way to tell chrome renderer about this. So it's still use it.
jira for this bug is VIZ-5387

### DIFF
--- a/decoder/vaapidecoder_base.cpp
+++ b/decoder/vaapidecoder_base.cpp
@@ -459,4 +459,11 @@ Decode_Status VaapiDecoderBase::outputPicture(const PicturePtr& picture)
         picture->m_timeStamp)?DECODE_SUCCESS:DECODE_FAIL;
 }
 
+VADisplay VaapiDecoderBase::getDisplayID()
+{
+    if (!m_display)
+        return NULL;
+    return m_display->getID();
+}
+
 } //namespace YamiMediaCodec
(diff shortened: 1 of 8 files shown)


uartie:
applied to chromiumos and seem to be getting some occasional errors when changing resolution/quality from lower to higher res:

libyami error(egl_vaapi_image.cpp, 103): vaGetImage return 3
libyami error(egl_vaapi_image.cpp, 103): vaGetImage return 8
libyami error(egl_vaapi_image.cpp, 103): vaGetImage return 18

...this seems to point to a race condition.  I think the egl image blt is being executed on a stale EglVaapiImage in outputPulse while the decoder is in the middle of resetting the va context and the egl vaapi images.

xuguangxin:
No, the root cause of this problem is EglVaapiImage do not hold VaapiDisplay's reference. And https://github.com/01org/libyami/blob/master/decoder/vaapidecoder_vp9.cpp#L98 will destroy all reference to VaapiDisplay. So it will call vaTerminate. It makes all VADisplay and VAImage and VASurface invalidate. Currently  it hard to fix it. Because we can't export VaapiDisplay to decoder interface level. And it will automatic fixed if we vaInitialize and vaTerminate in v4l2decoder(core library's calller).
So I suggest we merge this and verify this later after we implemented external vaInitialize.

uartie:
Since those particular errors don't seem to have any dire side-effects, I'd be fine with this being merged.

chivakker:
Code looks good to me to be merged.

PR #225, opened 2015-05-29
h264dec: accpet annexb format codec data

xuguangxin:
If you read a ts file using avformat, the codec data and frame data is
annexb format.So We'd better accpet annexb format in codec data


xuguangxin:
changed as uartie's comment

YuJiankang:
+1

xuguangxin:
merged as ae02c58a6baf6fa30146ff75f1822cdaeb546d31

PR #232, opened 2015-06-10
v4l2_encode: do not deref NULL m_encoder

uartie:
If createVideoEncoder fails (i.e. returns NULL), then we need
to avoid dereferencing it and return -1 to the client so it
can choose to proceed gracefully if needed.

This can happen if we compile without the h264 encoder (i.e.
--disable-h264enc) and the client still attempts to request
an h264 encode format.

### DIFF
--- a/v4l2/v4l2_encode.cpp
+++ b/v4l2/v4l2_encode.cpp
@@ -83,7 +83,8 @@ bool V4l2Encoder::start()
 bool V4l2Encoder::stop()
 {
     Encode_Status encodeStatus = ENCODE_SUCCESS;
-    encodeStatus = m_encoder->stop();
+    if (m_encoder)
+        encodeStatus = m_encoder->stop();
     return encodeStatus == ENCODE_SUCCESS;
 }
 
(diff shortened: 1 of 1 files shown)


v4l2/v4l2_encode.cpp
```diff
@@ -297,6 +298,10 @@ int32_t V4l2Encoder::ioctl(int command, void* arg)
             switch (format->fmt.pix_mp.pixelformat) {
                 case V4L2_PIX_FMT_H264: {
                     m_encoder.reset(createVideoEncoder(YAMI_MIME_H264), releaseVideoEncoder);
+                    if (not m_encoder) {
```
xuguangxin: if (!m_encoder) is better

uartie: why? it's the same thing.

chivakker: Yes they read more naturally although the rest of this code uses the ! convention. I believe ! is widely known as "not" for most people with a CS background.

uartie: Ah yes, I'll follow the project convention... I don't have a problem with that.  I'll resubmit with that change.

uartie: Resubmitted with requested change.

chivakker:
Thank you @uartie.

+1 lgtm, please merge.

seanvk:
lgtm, +1.

Thanks,

Sean

xuguangxin:
merged as acdd4e181f21092084c4db8a5f0d40cfb0b105b1
thanks @uartie.

PR #233, opened 2015-06-11
fixed permissive warnings

andyLiuChina:
Compile passed when open the "-fpermissive" option, but errors must be occured if no this option. 
That is, these are incorrect cplusplus statements as following.


uartie:
Commit message titles are a bit long... perhaps summarize more and put most details in the message body.  Otherwise, LGTM.

xuguangxin:
+1

xuguangxin:
merged as 4999f7ce48bd436bcccba1eca3869037b33ad799

PR #237, opened 2015-06-12
warningInfo: initialize local variables

andyLiuChina:
Init some local variables, otherwise it`ll run failed under some conditions.
As following:
    1) the local var "object" uninitialized, program will be aborted if m_frameHdr.num_components > 4.
    2) "count" uninitialized, mybe zero or random value. Depended on compiler. 
    3) And others will be assigned before use it, but output warning log still. So better init it.


xuguangxin:
+1

xuguangxin:
merged as 226c2b32219a7f62257d08693d562f0980c859e6

PR #239, opened 2015-06-15
warningInfo: unused but set variables and lable

andyLiuChina:
1) vaapidecoder_h264.cpp: var "sizeChanged" unused but set
2) vaapidecoder_jpeg.cpp:lable "end" unused but set
3) vaapiencoder_h264.cpp: define a status-variable at the beginning of a function, and then should return this status-var at the end of function. Make the function structure  understand clearly.

### DIFF
--- a/encoder/vaapiencoder_h264.cpp
+++ b/encoder/vaapiencoder_h264.cpp
@@ -890,7 +890,7 @@ Encode_Status VaapiEncoderH264::getParameters(VideoParamConfigType type, Yami_PT
 
     FUNC_ENTER();
     if (!videoEncParams)
-        return ENCODE_INVALID_PARAMS;
+        return status;
     switch (type) {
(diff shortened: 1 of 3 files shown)


encoder/vaapiencoder_h264.cpp
```diff
@@ -914,7 +914,9 @@ Encode_Status VaapiEncoderH264::getParameters(VideoParamConfigType type, Yami_PT
     }
 
     // TODO, update video resolution basing on hw requirement
-    return VaapiEncoderBase::getParameters(type, videoEncParams);
```
xuguangxin: why we need call VaapiEncoderBase::getParameters, can we just return status.

xuguangxin:
+1

xuguangxin:
merged as 75002d68ef536834d85213ef46dbdf96969a90a9

PR #240, opened 2015-06-15
Add CBR support for libyami

lizhong1008:
1. Let yamiencode test case can support CBR mode and set different QP
2. Add CBR support for h264 and vp8 encoding


encoder/vaapiencoder_base.cpp
```diff
@@ -322,7 +326,7 @@ SurfacePtr VaapiEncoderBase::createSurface(const SharedPtr<VideoFrame>& frame)
 
 void VaapiEncoderBase::fill(VAEncMiscParameterHRD* hrd) const
 {
-    hrd->buffer_size = m_videoParamCommon.rcParams.bitRate * m_videoParamCommon.rcParams.windowSize/1000;
+    hrd->buffer_size = m_videoParamCommon.rcParams.bitRate * 4;
```
xuguangxin: 4 seems a magic number. Can we use a macro for this?
Why original code use windowSize, new code do not use it?

encoder/vaapiencoder_base.h
```diff
@@ -127,6 +128,11 @@ class VaapiEncoderBase : public IVideoEncoder {
     uint32_t frameRateNum() const {
         return m_videoParamCommon.frameRate.frameRateNum;
     }
+
+    uint32_t fps() const {
+        return m_videoParamCommon.frameRate.frameRateNum / m_videoParamCommon.frameRate.frameRateDenom;
```
xuguangxin: Is any possible m_videoParamCommon.frameRate.frameRateDenom == 0?
It will crash the application

encoder/vaapiencoder_vp8.cpp
```diff
@@ -47,6 +47,9 @@ class VaapiEncPictureVP8 : public VaapiEncPicture
 VaapiEncoderVP8::VaapiEncoderVP8():m_frameCount(0)
 {
     m_videoParamCommon.profile = VAProfileVP8Version0_3;
+    m_videoParamCommon.rcParams.minQP = 9;
```
xuguangxin: why it's 9. not 1 or 0.

encoder/vaapiencoder_vp8.cpp
```diff
@@ -47,6 +47,9 @@ class VaapiEncPictureVP8 : public VaapiEncPicture
...
+    m_videoParamCommon.rcParams.maxQP = 127;
```
xuguangxin: better use  macro to define those number

encoder/vaapiencoder_vp8.cpp
```diff
@@ -188,15 +188,21 @@ bool VaapiEncoderVP8::fill(VAEncPictureParameterBufferVP8* picParam, const Pictu
...
 
 bool VaapiEncoderVP8::fill(VAQMatrixBufferVP8* qMatrix) const
 {
-    //TODO: FIX THIS
-    for (int i = 0; i < 4; i++) {
-        qMatrix->quantization_index[i] = 40;
+    int i;
+
+    for (i = 0; i < 4; i++) {
+        qMatrix->quantization_index[i] = (initQP() > minQP() && initQP() < maxQP()) ? initQP() : 40;
```
xuguangxin: better calc this in start or some where.it's no need calc every-time.

encoder/vaapiencoder_vp8.cpp
```diff
@@ -188,15 +188,21 @@ bool VaapiEncoderVP8::fill(VAEncPictureParameterBufferVP8* picParam, const Pictu
...
 }
 
 bool VaapiEncoderVP8::fill(VAQMatrixBufferVP8* qMatrix) const
 {
-    //TODO: FIX THIS
-    for (int i = 0; i < 4; i++) {
-        qMatrix->quantization_index[i] = 40;
+    int i;
+
+    for (i = 0; i < 4; i++) {
```
xuguangxin: better use N_ELEMENTS(qMatrix->quantization_index) instead of 4.

tests/encodehelp.h
```diff
@@ -55,27 +58,51 @@ static void print_help(const char* app)
...
     printf("   -N <number of frames to encode(camera default 50), useful for camera>\n");
+    printf("   --qp <initial qp> optional\n");
+    printf("   --rcmode <CBR|CQP> optional\n");
+}
+
+static VideoRateControl string_to_rc_mode(char *str)
+{
+    VideoRateControl rcMode;
+
+    if (!strncmp(str, "CBR", 3))
```
xuguangxin: better use case less compare.

tests/encodehelp.h
```diff
@@ -103,6 +130,16 @@ static bool process_cmdline(int argc, char *argv[])
             if (strlen(optarg) == 4)
                 inputFourcc = VA_FOURCC(optarg[0], optarg[1], optarg[2], optarg[3]);
             break;
+        case 1:
```
xuguangxin: hard to read and maintain, and may conflict with short option. Better let getopt_long_only return 0 for long option. and check the option_index to do things.

encoder/vaapiencoder_base.cpp
lizhong1008: vaapi defines window_size to indicate driver must guarantee target bit-rate during a window time, but original code used windowSize as HDR CBP buffer size in ms.
4 just means HDB buffer_size is 4 times of bitrate, you can set another value,such as 2,6,8. And it can be changed for every IDR frame, so I don't think it's necessary to use a macro.

encoder/vaapiencoder_base.h
lizhong1008: frameRateDenom has been initialized to be 1 when class VaapiEncoderBase is constructed. 
Nobody will change it to be 0 if he understand what  frameRateDenom means.

encoder/vaapiencoder_vp8.cpp
lizhong1008: It's just a empirical value, because I think qp smaller than it will bring huge bitstream size increasing but litter quality improvement. It's ok to set to be 1, but never 0, beause QP is a numerator actually.

encoder/vaapiencoder_vp8.cpp
lizhong1008: They are just initial values, and can be change for every frame.

encoder/vaapiencoder_vp8.cpp
lizhong1008: Good suggestion. Thanks for remind.

encoder/vaapiencoder_vp8.cpp
lizhong1008: Yeah, agree. Thanks

tests/encodehelp.h
lizhong1008: Can you specify why "may conflict with short option" and "Better let getopt_long_only return 0 for long option. and check the option_index to do things" ?

xuguangxin:
+1

YuJiankang:
+1

xuguangxin:
Hi @lizhong1008:
Can we close this?

lizhong1008:
Hi Guangxing:
     Is it necessary to wait for QA test result before close it?

xuguangxin:
no need, it's a pull request not a issue. we need close it after pull request accepted.

PR #241, opened 2015-06-15
factory: self-register enc/dec/vpp with their factories

uartie:
The internal design to enable/disable a specific codec
from being compiled into the library has unnecessary
complexity and too much maintenance overhead.

Add a generic factory template that provides a mechanism for
codecs to register themselves for a specific mime type simply
by compiling the code unit into the library.  This also removes
the odd "template within a macro" factory (which only adds
complexity and frustration while debugging).

With this change, there is no need to maintain and check
for __BUILD_<CODEC>__ autoconf defines inside the sourcecode
units anymore (Note: __BUILD_FAKE_DEC__ is still the
exception, atm).  Determining whether to build a codec into
the library is purely managed via the autotool configurations
now.


seanvk:
lgtm +1.

xuguangxin:
Hi uartie:
Thanks for your patch, It's a good improvement. I planed to change this many time again, But do not get time to.Since you are changing those code. I'd like put some requirements for your patch.
1. It's better have runtime check in VA's profile. If VA do not supports VAProfileVP9Version0, we need return NULL for decoder creation.Fail in create time can help customer do better decision. such as fail back to software decoder.
2. it should have compile time check for component. ie. if it do not have va_dec_vp9.h. it should not build the vp9 decoder. 
We can postpone item 2. But I hope we can have item 1 in near future.
also I list some prototype here.

/* create video factory base on YamiNativeDisplayType */
/* we will support 3 native display in future, NATIVE_DISPLAY_VA, NATIVE_DISPLAY_VDPAU, NATIVE_DISPLAY_D3D */
VideoFactory* createVideoFactory(YamiNativeDisplayType displayType, intptr_t display);

void releaseVideoFactory(VideoFactory*);

struct VideoFactory
{
    SharedPtr<IVideoDecoder> createDecoder(const char* mime) = 0;
    SharedPtr<IVideoEncoder> createEncoder(const char* mime) = 0;
    SharedPtr<IVideoPostProcess> createPostProcesser(const* char* mime) = 0;
};

//create decoder/encoder/vpp base on display's caps.
class VaFactory : public VideoFactory
{
    
};
Do yo think it's OK?

Hi Sean:
Is it urgent? If not,  I'd like merge this patch after runtime check added.

uartie:
@xuguangxin I think it's important to note that this patch does not change functionality... rather, it only refactor's (and hopefully improves) the codec factory design to make it easier to maintain/update in the future.

What you're proposing in 1 and 2 are functional changes which may or may not necessarily relate to this change.  I can envision multiple ways to deal with the use case in 1.  Although I agree that 1 and 2 are reasonable proposals, I don't think they should hold up a patch such as this one.

I'd be happy to look into implementing 1 and 2 above, but I may not have time to do it immediately.

You're prototype seems reasonable too... but perhaps give it a different name.  "Video" implies that it only handles video.  What about "MediaPlatformFactory" or something ?

seanvk:
Hi Guangxin,

Your runtime check for codec support in general is a necessary aspect on top of the factory addition.  I would say instead, let's get the factory in place, then ensure that we have the VA Entry profile checks for runtime.   Also agree that checks on header presence is key.

But those are functional changes we are going to need to experiment with and think hard about.  Do we allow the build to simply fail gracefully or do we do more invasive checks for presence of headers, etc.

My preference is to merge this change first.   Let's also make sure before we merge this commit that we run a QA cycle.

Thanks,

Sean

uartie:
@seanvk, do you mean you want a QA cycle *with* this patch or *without*  before merging?

xuguangxin:
Hi uartie :
Yes. It's only create video component. Because createVideoFactory  only accept display handle. But you can change it to YamiFactory since our project name is yami. :)

Hi Sean:
Ok, But daily regression may be enough for this patch.It's just functional patch, do not have so many logical branch in it.
+1 for me.

chivakker:
lgtm +1, Can we try to merge this and later get the fixes for 1) and 2) suggested?

seanvk:
I just want a QA regression test with this patch, just before merge.  I already saw one revert the other day with some codec parser updates.  That's all.

chivakker:
Ok, I'll make sure the QA testing is done this week.

xuguangxin:
This passed our regression test, waiting daniel's QA test result to merge.

chivakker:
QA on our side is also passing.  Please merge

seanvk:
May need to be rebased.

uartie:
rebased.

PR #245, opened 2015-06-17
fix typo in code.

xuguangxin:
two commit here to fix typo:
1. preSandboxInitDecoder in vpp/vaapipostprocess_host.cpp will conflict with decoder/vaapidecoder_host.cpp. Halley assume preSandboxInitXXX will called by chrome, but actually it's not. Since nobody use it, we'd better remove it all.
2. fix == error in assignment


chivakker:
The two commits lgtm +1

PR #247, opened 2015-06-18
vaapidisplay: use the libva api for driver name

chivakker:
libva has introduced a new api to set the driver name, this is changed
when the default driver is not going to be used. Right now the driver
and supported profiles are handled by libyami

### DIFF
--- a/vaapi/vaapidisplay.cpp
+++ b/vaapi/vaapidisplay.cpp
@@ -54,6 +54,16 @@ using std::list;
 
 //helper function
 namespace YamiMediaCodec{
+static bool vaSelectDriver(VADisplay vaDisplay, const std::string& driverName)
+{
+    VAStatus vaStatus=VA_STATUS_SUCCESS;
+    if (!driverName.empty())
+    {
+        vaStatus = vaSetDriverName(vaDisplay, driverName.c_str());
+    }
+    return checkVaapiStatus(vaStatus, "vaSetDriverName");
+}
+
 static bool vaInit(VADisplay vaDisplay)
 {
    int majorVersion, minorVersion;
@@ -226,7 +236,7 @@ class DisplayCache
 {
 public:
     static SharedPtr<DisplayCache> getInstance();
-    DisplayPtr createDisplay(const NativeDisplay& nativeDisplay);
+    DisplayPtr createDisplay(const NativeDisplay& nativeDisplay, const std::string& driverName="");
 
     ~DisplayCache() {}
 private:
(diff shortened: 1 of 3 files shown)


vaapi/vaapidisplay.cpp
```diff
@@ -320,4 +333,14 @@ DisplayPtr VaapiDisplay::create(const NativeDisplay& display)
 {
     return DisplayCache::getInstance()->createDisplay(display);
 }
+
+DisplayPtr VaapiDisplay::create(const NativeDisplay& display, VAProfile profile)
+{
+    SharedPtr<DisplayCache> cache = DisplayCache::getInstance();
+    const std::string name="hybrid";
```
xuguangxin: It always select hybrid. What happened if we support vp9 decoder in hardware?

vaapi/vaapidisplay.h
```diff
@@ -31,6 +31,7 @@
 #include <va/va_x11.h>
 #endif
 #include <va/va_drm.h>
+#include <string>
```
uartie: This could probably be moved to vaapidisplay.cpp.

vaapi/vaapidisplay.cpp
uartie: This should be fine until we do... and refactor later.

uartie:
Minor nit... otherwise LGTM. +1

seanvk:
lgtm +1

vaapi/vaapidisplay.cpp
xuguangxin: Which driver should load is really a platform related info. Yami core library  do not need know this, and it not even possible. We'd better select driver name and init va in  in VaapiV4L2Device::Initialize() and set va display as NATIVE_DISPLAY_VA. This also apply to drm device name, open "/dev/card0" or ""/dev/dri/renderD128" should decide in VaapiV4L2Device::Initialize() not libyami core library.

uartie: @xuguangxin, Looks like VaapiV4L2Device does not even exist.  Has someone already started working on this addition?  If so, is there a topic/fork branch somewhere?  If not, what do you envision for this VaapiV4L2Device?

uartie: @last comment, nevermind I just realized you were probably talking about the VaapiV4L2Device in chromium...

chivakker: @xuguangxin Some layer has to know that there's a need to use a different driver to use another codec.

The intention with this patch is to push the driver selection as low on the stack as possible.  To be able to do the driver selection at the V4L2 level Yami has to refactor and expose vaSetDriverName and vaInitialize in two different API functions.  I would prefer to do this later as an enhancement.  Also Querying the driver for supported profiles and/or entry points can be done later.

xuguangxin:
Ok, merged as 07359e9b7900477d678e82abb131f936144c76df, thanks Daniel

PR #259, opened 2015-07-16
MD5: print MD5 under the mode -2

andyLiuChina:
Using the mode -2 to print MD5 by per frame, and the MD5 of the whole
decoded YUV at the end, do not dump YUV to file again.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -134,6 +134,10 @@ if test "$enable_baytrail" = "yes"; then
 AC_DEFINE([__PLATFORM_BYT__], [1], [Defined to 1 if --enable-baytrail="yes" ])
 fi
 
+AC_ARG_ENABLE(md5,
+    [AC_HELP_STRING([--enable-md5], [enable generate md5 by per frame@<:@default=yes@:>@])],
+    [], [enable_md5="yes"])
+
 dnl vp8 decoder
 AC_ARG_ENABLE(vp8dec,
     [AC_HELP_STRING([--enable-vp8dec], [build with vp8 decoder support @<:@default=yes@:>@])],
@@ -273,6 +277,14 @@ fi
 if test "$enable_avformat" = "yes"; then
 PKG_CHECK_MODULES(LIBAVFORMAT,  libavformat libavcodec libavutil)
 fi
+
+#check openssl
+if test "$enable_md5" = "yes"; then
+PKG_CHECK_MODULES(OPENSSL, [openssl],
+	         [AC_DEFINE([__ENABLE_MD5__], [1], [Defined to 1 if MD5 API and --enable-md5[default] are enabled])],
+	         [enable_md5="no"])
+fi
+AM_CONDITIONAL(ENABLE_MD5, test "x$enable_md5" = "xyes")
 
 # Checks for library functions.
 AC_FUNC_MALLOC
--- a/tests/decodeoutput.cpp
+++ b/tests/decodeoutput.cpp
@@ -4,6 +4,7 @@
  *  Copyright (C) 2011-2014 Intel Corporation
  *    Author: Halley Zhao<halley.zhao@intel.com>
  *    Author: Xu Guangxin <guangxin.xu@intel.com>
+ *    Author: Liu Yangbin <yangbinx.liu@intel.com>
  *
  *  This library is free software; you can redistribute it and/or
  *  modify it under the terms of the GNU Lesser General Public License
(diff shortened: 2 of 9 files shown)


tests/decodeoutput.cpp
```diff
@@ -267,6 +267,81 @@ DecodeOutputFileDump::~DecodeOutputFileDump()
...
+    //ASSERT(m_width == frame->width && m_height == frame->height);
+    if (!getPlaneResolution(frame->fourcc, frame->width, frame->height, width, height, planes))
+        return false;
+
+    for (int i = 0; i < planes; i++) {
+        uint8_t* data = reinterpret_cast<uint8_t*>(frame->handle);
+        data += frame->offset[i];
+        for (int h = 0; h < height[i]; h++) {
+            for(uint32_t j = 0; j < width[i]; j++)
+                singleFrame.push_back(data[j]);;
```
xuguangxin: it will crash if the frame is very large.
can you just call MD5Update here ?

tests/decodeoutput.cpp
```diff
@@ -267,6 +267,81 @@ DecodeOutputFileDump::~DecodeOutputFileDump()
...
+    }
+
+    frameCnt++;
+    const uint8_t* p = singleFrame.data();
+    std::string md5str = getMd5Value(p, singleFrame.size());
+    fprintf(stdout, "[frame %d] %s\n", frameCnt, md5str.c_str());
+
+    std::vector<uint8_t>::const_iterator it;
+    for(it = singleFrame.begin(); it != singleFrame.end(); it++)
+        allFrame.push_back(*it);
```
xuguangxin: do not collect all frame data. it  will very large

configure.ac
```diff
@@ -134,6 +134,15 @@ if test "$enable_baytrail" = "yes"; then
 AC_DEFINE([__PLATFORM_BYT__], [1], [Defined to 1 if --enable-baytrail="yes" ])
 fi
 
+AC_ARG_ENABLE(openssl,
```
xuguangxin: better change logical like this.
1. if --eanble-md5=no, disable -2 mode
2. automatic check openssl, if it's exist, disable the -2 mode

tests/decodeoutput.cpp
```diff
@@ -268,6 +269,90 @@ DecodeOutputFileDump::~DecodeOutputFileDump()
...
+        for (uint32_t h = 0; h < height[i]; h++) {
+            MD5_Update(&t_ctx, data, width[i]);
+            MD5_Update(&m_md5Ctx, data, width[i]);
+            data += frame->pitch[i];
+        }
+    }
+
+    frameCnt++;
+    MD5_Final(result, &t_ctx);
+    fprintf(m_fp, "[frame %d] ", frameCnt);
```
xuguangxin: print md5 is enough, do not print frame number.

tests/decodeoutput.cpp
```diff
@@ -268,6 +269,90 @@ DecodeOutputFileDump::~DecodeOutputFileDump()
...
+    fprintf(m_fp, "\n");
+
+    return true;
+}
+
+DecodeOutputExtractMD5::~DecodeOutputExtractMD5()
+{
+    uint8_t result[16] = {0};
+    MD5_Final(result, &m_md5Ctx);
+    fprintf(m_fp, "The whole frames MD5 ");
```
xuguangxin: better print this in stdout too

tests/decodeoutput.cpp
```diff
@@ -268,6 +269,90 @@ DecodeOutputFileDump::~DecodeOutputFileDump()
...
+    for(uint32_t i = 0; i < 16; i++)
```
xuguangxin: can we put this to a function? so the code will not dumplicate

testscripts/autotest.py
```diff
@@ -25,8 +25,8 @@
     psnr       = os.path.join(autotestpath, 'psnr')
     decodemodelist = ['m', '0']
     encodemodelist = ['m', '0']
-    v4l2wmodelist  = ['w', '0', '1']
-    v4l2mmodelist  = ['m', '-1', '0', '1', '2', '3', '4']
```
xuguangxin: why we need remove those mode?

testscripts/autotest.py
```diff
@@ -25,8 +25,8 @@
     psnr       = os.path.join(autotestpath, 'psnr')
     decodemodelist = ['m', '0']
```
xuguangxin: why we only have mode 0

xuguangxin:
+1

xuguangxin:
merged as ded0381311079759aaaf868684ddeaf7ffac6e45

PR #260, opened 2015-07-17
Fix encoded frame number is larger than set value

lizhong1008:


xuguangxin:
can you change checkin comment to 
"module:short comment

long comment "

for example
tests: fix xxxxx

lizhong1008:
Thanks for reminder. Comment updated.

PR #261, opened 2015-07-17
vp9dec: fix vp9 frame header parse mistake.

zhongcong:
when it's a intra_only frame, the frame size and display frame size will not be read. It's a vp9 frame header parse bug.

After fix this bug, 4 clips test pass.
Clips:
Syntax_VP9_1920x1080_106_inter_hidden_frames_1.1.vp9
Syntax_VP9_1920x1080_107_inter_compound_prediction_1.1.vp9
Syntax_VP9_432x240_106_inter_hidden_frames_1.1.vp9
Syntax_VP9_432x240_107_inter_compound_prediction_1.1.vp9 

### DIFF
--- a/codecparsers/vp9parser.c
+++ b/codecparsers/vp9parser.c
@@ -150,6 +150,8 @@ static void read_frame_size(BitReader* br,
 
 static void read_display_frame_size(Vp9FrameHdr* frame_hdr, BitReader* br)
 {
+  frame_hdr->display_width = frame_hdr->width;
+  frame_hdr->display_height = frame_hdr->height;
   frame_hdr->display_size_enabled = vp9_read_bit(br);
   if (frame_hdr->display_size_enabled) {
     read_frame_size(br, &frame_hdr->display_width, &frame_hdr->display_height);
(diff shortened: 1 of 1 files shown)


zhongcong:
updated.

xuguangxin:
why we need change so much lines. can we change this like wei's pull request?https://github.com/01org/libyami/pull/268/files

codecparsers/vp9parser.c
```diff
@@ -624,6 +626,49 @@ static Vp9ParseResult vp9_parser_update(Vp9Parser* parser, const Vp9FrameHdr* co
...
+{
+  BOOL *ss_x = &frame_hdr->subsampling_x;
+  BOOL *ss_y = &frame_hdr->subsampling_y;
+
+  if (frame_hdr->profile > VP9_PROFILE_1)
+    frame_hdr->bit_depth = vp9_read_bit(br);
+  frame_hdr->color_space = (VP9_COLOR_SPACE)vp9_read_bits(br, 3);
+
+  if (frame_hdr->color_space == VP9_SRGB) {
+    if (frame_hdr->profile == VP9_PROFILE_1 || frame_hdr->profile == VP9_PROFILE_3) {
```
xuguangxin: we do not support SRGB. no need do future read.

zhongcong:
@xuguangxin Yet for the frame widht and height issue. There are 4 issues.
1) i think that display width/height should be set to clip width and height if there is no diplay size flag in clip. 
2) if the clip is inter frame. subsamping and color space should also be set to the default.
3) The SRGB maybe should supported for future usage.
4) revise redundant code into function.

xuguangxin:
Hi Cong:
I will use Linwei's https://github.com/01org/libyami/pull/268, since it can fix the problem and much more  simpler.

PR #263, opened 2015-07-22
gst: rewrite macros and functions

andyLiuChina:
the original macros and functions in the file of gstutils.h are under
the GNU General Public License, so it must be rewrited under the LGPL.


YuJiankang:
+1

Issue #266, opened 2015-07-24
[HSW][media][Encode]error message "Compression ratio found 2.54 in Access Unit for level 4.0 ..." appeared after using VEGA tool to playback H264 cqp encode bitstream


keqiaox:
1.Testing environment:
libva: (staging) d9626b77f865ce7bdbc8c78988a86d3e4f55d67b
intel driver: (staging) a8ca5f911c7c19b5ede3f96ac6113606b9dd4d95
libyami: (master) 7730d21
omxil-core: (master) d6f6edef3119d5622f9066521fc4b53004410a27
omx-components: (master) 464135af23ac7d3ae5d3a9240f2fd7a058eb2d7c
gst-omx: (master) f178afe59e507ea7113bbc630affd6d502a2418c

2.Test command: 
    use VEGA tool to playback the bitstreams
3.Result:
   error log appeared :
   "Compression ratio found 2.54 in Access Unit for level 4.0 .allowed Min compression        ratio(MinCR)is 4. coded bits=4358856 Uncompressed bits =11059200"
4. Related bitstreams:
   OBILE_352x288_30_orig_01.yuv_7.h264
   MOBILE_352x288_30_orig_01.yuv_14.h264
   MOBILE_352x288_30_orig_01.yuv_21.h264
   V007383_Wall_walkers_176_144p_25fps_371frames.yuv_7.h264
   V007383_Wall_walkers_176_144p_25fps_371frames.yuv_14.h264

lizhong1008:
Create a commit to fix this issue: https://github.com/lizhong1008/libyami/commit/4612f42589430fdda48e8c4fadf7e6fa13946f11
Waiting for review.

Issue #265, opened 2015-07-24
[HSW][media][Encode]error message "Unexpected byte value after Emulation byte .found 30 allowed 0,1,2,3"  appeared after using VEGA tool to playback H264 cqp encode bitstream


keqiaox:
1.Testing environment:
libva: (staging) d9626b77f865ce7bdbc8c78988a86d3e4f55d67b
intel driver: (staging) a8ca5f911c7c19b5ede3f96ac6113606b9dd4d95
libyami: (master) 7730d21
omxil-core: (master) d6f6edef3119d5622f9066521fc4b53004410a27
omx-components: (master) 464135af23ac7d3ae5d3a9240f2fd7a058eb2d7c
gst-omx: (master) f178afe59e507ea7113bbc630affd6d502a2418c

2. Test command:
 use VEGA tool to playback the bitstreams

3.Result:
  error log appeared :
  "Unexpected byte value after Emulation byte .found 30 allowed 0,1,2,3"
4. Related bitstreams:
720x480p.yuv_7.h264
720x480p.yuv_14.h264
720x480p.yuv_21.h264
720x480p.yuv_35.h264
720x480p.yuv_42.h264
720x480p.yuv_49.h264
CITY_704x576_30_orig_01.yuv_21.h264
CITY_704x576_30_orig_01.yuv_28.h264
CITY_704x576_30_orig_01.yuv_35.h264
CITY_704x576_30_orig_01.yuv_49.h264
V007383_Wall_walkers_176_144p_25fps_371frames.yuv_21.h264
V007383_Wall_walkers_176_144p_25fps_371frames.yuv_28.h264
V007383_Wall_walkers_176_144p_25fps_371frames.yuv_35.h264
V007383_Wall_walkers_176_144p_25fps_371frames.yuv_42.h264
V007383_Wall_walkers_176_144p_25fps_371frames.yuv_49.h264

lizhong1008:
Create a commit to fix this issue: lizhong1008@4612f42
Waiting for review.

xuguangxin:
hi keqiao:
it's merged. Please help to verify if you have time

keqiaox:
Hi Guangxin,
 After re-testing these cases with the latest commit, did not found this issue,so close the bug.

PR #267, opened 2015-07-25
h264enc: fix h264 HDR comformance issues

lizhong1008:
1. Let h264 level equal to max level, in order to support high level
bitrate, 4K encoding for example
2. As spec, only when bitstream contains 0x000000/0x000001/0x000002/0x000003
need to insert emulation byte 0x03

This patch ties to fix https://github.com/01org/libyami/issues/266 and https://github.com/01org/libyami/issues/265


YuJiankang:
+1

PR #268, opened 2015-07-27
vp9dec: read_frame_size for intra-only frame

wlin70:
Fix the bug missing read_frame_size for intra-only frames' VP9 Hdr

Signed-off-by: Wei Lin <wei.w.lin@intel.com>


wlin70:
Close this pull request since fixed in other pull request #261 :
vp9dec: fix vp9 frame header parse mistake.

xuguangxin:
Hi Linwei:
I have merged this patch to master, since it can fix the problem and much more simpler than Cong's.
thanks.

PR #269, opened 2015-07-27
vp9dec: resolution-change error on repeated frames(show_existing_frame)

wlin70:
Repeated frames with hdr->show_existing_frame==1 are only for display purpose,
their hdr->width/height are init as ZERO and should NOT used to check resolution
change.

This fix the assertion error on playing Syntax_VP9_432x240_206_extra_repeat_existing_1.1.vp9

Signed-off-by: Wei Lin <wei.w.lin@intel.com>


chivakker:
It sounds like the fix is to report the proper width and height even on the case of show_existing_frames instead of just ignoring it in this case.  If the parser is setting this flag it should also populate the other values.

xuguangxin:
seems good to me.

Hi Daniel:
Do you have any concern about this patch?

wlin70:
Hi, Daniel: 
I cannot find a simple/easy way to read a decoded frame's size and write to hdr->width/height in vp9parser.c.   We can add surface->getWidth/Height() to get size and re-assign "hdr"  in vaapidecoder_vp9.cpp,  but it is defined as "const *Vp9FrameHdr) in parameter passing. Bad to change like this. 

To my pull request,  1) the repeated frame is  NOT decoded again and only for display.  so such frames can and should be bypass in VaapiDecoderVP9::ensureContext.   2)   since we should re-assign a existing surface with picture->setSurface,  so picture->getSurface()->resize is redundant.

chivakker:
HI, It looks good to me given that the parser was considered as a possibility and this is a better fix.  Thanks for addressing my concern.

xuguangxin:
merged as 216ba67186a895e8b000bece586af673701704a6,thanks Wei & Daniel

PR #272, opened 2015-07-30
compileOpt: adujst the location of compile options

andyLiuChina:
i set the comple options in the autogen.sh in previous commit, which is not noly a bad ideal but also just is useful for c++ compiler.


xuguangxin:
+1

xuguangxin:
merged as 7e2556efc8223b0f0289894364012e47cb6de548

PR #276, opened 2015-08-04
h264enc: fix one h264 HDR comformance issue

lizhong1008:
As spec, only when bitstream contains 0x000000/0x000001/0x000002/0x000003
need to insert emulation byte 0x03

This patch is to fix: https://github.com/01org/libyami/issues/265

### DIFF
--- a/encoder/vaapiencoder_h264.cpp
+++ b/encoder/vaapiencoder_h264.cpp
@@ -538,8 +538,15 @@ private:
             m_headers.insert(m_headers.end(), s, e);
             if (e == h.end())
                 break;
-            m_headers.insert(m_headers.end(), emulation, emulation + N_ELEMENTS(emulation));
-            s = e + N_ELEMENTS(zeros);
+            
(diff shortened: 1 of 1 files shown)


encoder/vaapiencoder_h264.cpp
```diff
@@ -538,8 +538,12 @@ class VaapiEncStreamHeaderH264
             m_headers.insert(m_headers.end(), s, e);
             if (e == h.end())
                 break;
-            m_headers.insert(m_headers.end(), emulation, emulation + N_ELEMENTS(emulation));
+            
             s = e + N_ELEMENTS(zeros);
+            if (*s <= 3)
```
xuguangxin: have some comments here, please. why it's 3 not 4, 5, 6....not every code ready will do "git blame" and see commit log.

lizhong1008: Agree to add comments here, patch updated. But commit log is important for every developer, thus means "git blame"

xuguangxin:
+1

PR #281, opened 2015-08-07
klocwork: fix klocwork complains

YuJiankang:
fix klocwork complains

### DIFF
--- a/codecparsers/h265parser.c
+++ b/codecparsers/h265parser.c
@@ -282,6 +282,8 @@ h265_parse_profile_tier_level (H265ProfileTierLevel * ptl,
   }
 
   for (i = 0; i < maxNumSubLayersMinus1; i++) {
+    if (i >= 6)
+      goto error;
     if (ptl->sub_layer_profile_present_flag[i]) {
       READ_UINT8 (nr, ptl->sub_layer_profile_space[i], 2);
       READ_UINT8 (nr, ptl->sub_layer_tier_flag[i], 1);
@@ -430,6 +432,9 @@ error:
 static bool
 h265_parse_vui_parameters (H265SPS * sps, NalReader * nr)
 {
+  if (!sps)
+    return FALSE;
+
   H265VUIParams *vui = &sps->vui_params;
 
   DEBUG ("parsing \"VUI Parameters\"");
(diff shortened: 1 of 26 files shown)


tests/egl/gles2_help.c
```diff
@@ -37,15 +37,17 @@
 #define ASSERT assert
 #endif
 
-#define EGL_CHECK_RESULT_RET(result, promptStr, ret) do {   \
+#define EGL_CHECK_RESULT_RET(result, promptStr, memNeedfree, ret) do {   \
```
xuguangxin: What happen if we have many memory to free?Can we use goto here?

testscripts/psnr.cpp
```diff
@@ -53,6 +55,7 @@ psnr_calculate(char *filename1, char *filename2, const char *eachpsnr, const cha
     {
         printf("open ref yuv fail\n");
         fprintf(fppsnrresult,"open %s fail\n",filename1);
+        fclose(fppsnrresult);
```
xuguangxin: can we use goto error; here

codecparsers/h265parser.c
```diff
@@ -282,6 +282,8 @@ h265_parse_profile_tier_level (H265ProfileTierLevel * ptl,
   }
 
   for (i = 0; i < maxNumSubLayersMinus1; i++) {
+    if (i >= 6)
```
xuguangxin: better if we just assert maxNumSubLayersMinus1 <= 6 before loop

codecparsers/h265parser.c
```diff
@@ -582,6 +587,8 @@ get_scaling_list_params (H265ScalingList * dest_scaling_list,
             dest_scaling_list->scaling_list_dc_coef_minus8_16x16;
       break;
     case H265_QUANT_MATIX_32X32:
+      if (matrixId >=2)
```
xuguangxin: use assert is better

tests/encodeinput.cpp
```diff
@@ -80,6 +84,10 @@ bool EncodeInputFile::init(const char* inputFileName, uint32_t fourcc, int width
     m_height = height;
     m_fourcc = fourcc;
 
+    if ((m_width <= 0) ||(m_height <= 0) ||(m_width > MAX_WIDTH) || (m_height > MAX_HEIGHT)) {
```
xuguangxin: why we need MAX_WIDTH & MAX_HEIGHT?

xuguangxin:
merged as 458616e0aaf06c0844ad8f975279633c89fb8bb2

PR #282, opened 2015-08-07
yamidecode: only decode the specified number of frames

andyLiuChina:
you can specify how many number of frames to be decode with -n option.

### DIFF
--- a/tests/decode.cpp
+++ b/tests/decode.cpp
@@ -109,11 +109,14 @@ int main(int argc, char** argv)
             status = decoder->decode(&inputBuffer);
         }
 
-        renderOutputFrames(output);
+        renderOutputFrames(output, frameCount);
         if (output->renderFrameCount() >= 5 && !skipFrameCount4NetFps) {
             skipFrameCount4NetFps = output->renderFrameCount();
             calcFpsNet.setAnchor();
         }
+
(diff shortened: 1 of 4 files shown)


tests/decode.cpp
```diff
@@ -124,7 +129,11 @@ int main(int argc, char** argv)
 #endif
 
     // drain the output buffer
-    renderOutputFrames(output, true);
+    if (useSpecified) {
+        if(output->renderFrameCount() < frameCount)
+            renderOutputFrames(output, true);
```
xuguangxin: renderOuputFrames will output more than one frame. so you can't make judgement here.

tests/decodehelp.h
```diff
@@ -49,7 +49,7 @@ uint32_t dumpFourcc = VA_FOURCC_I420;
 char *inputFileName = NULL;
 int renderMode = 1;
 static int32_t waitBeforeQuit = 1;
-
+static uint32_t frameCount = 0, useSpecified = 0;
```
xuguangxin: one variable is enough, you can name it frameCountMax and init it to numeric_limits<uint32_t>::max()

xuguangxin:
+1

xuguangxin:
merged as 73b2078a6b5412df2e24af430bea4c5ac9288065

PR #285, opened 2015-08-11
Revert m_maxCodedbufSize calculation and refine it

lizhong1008:


xuguangxin:
Could we merge this two commits? it will make sure all commit is build able and run able.
and better you can put gstreamer vaapi's commit id in revert log.

lizhong1008:
yeah, all these two commits can be built, but the first one may can't run well. I'll merge them. Thanks

xuguangxin:
and make sure you can encode a small and a large file before you send pull request again.
thanks.

lizhong1008:
It has beent tested. Don't worry.

lizhong1008:
patch updated.

xuguangxin:
+1

PR #294, opened 2015-08-24
yamidecode: Don't render frame before decoding frame started.

zhongcong:

### DIFF
--- a/tests/decodeoutput.cpp
+++ b/tests/decodeoutput.cpp
@@ -59,10 +59,11 @@ Decode_Status DecodeOutput::processOneFrame(bool drain)
 {
     Decode_Status status = renderOneFrame(drain);
 
-    if (status == RENDER_SUCCESS)
+    if (status == RENDER_SUCCESS) {
         m_renderFrames++;
+        m_decoder->renderDone(&m_frame);
+    }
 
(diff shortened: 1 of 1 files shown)


zhongcong:
updated.
in processOneFrame(), if renderOneFrame() return RENDER_NO_AVAILABLE_FRAME, just return, don't need to invoke renderDone() function.

tests/decodeoutput.cpp
```diff
@@ -59,6 +59,9 @@ Decode_Status DecodeOutput::processOneFrame(bool drain)
 {
     Decode_Status status = renderOneFrame(drain);
 
+    if (status == RENDER_NO_AVAILABLE_FRAME)
```
xuguangxin: we only renderDone picture when status == RENDER_SUCCESS. Can we change code to this
...
if (status == RENDER_SUCCESS) {
      m_renderFrames++;
      m_decoder->renderDone....
}
return status;
}

zhongcong:
updated.

xuguangxin:
+1

xuguangxin:
merged as 8d64174072a3611c40bcfd0c63a6fd8457779cd6

PR #295, opened 2015-08-26
add  some options to grid applicaiton

xuguangxin:
-d output to assigned display.
-g create another grid instance 
-s put decode and vpp in same thread


YuJiankang:
+1

PR #298, opened 2015-09-01
capi: remove non existing function encodeflush from header

jsorg71:
This function does not exist in the interface and should not be in the capi.


jsorg71:
opps, re-doing this PR

PR #309, opened 2015-09-08
yamidecode: window should be created if formatInfo is set in decoder-…

zhongcong:
…>start().

### DIFF
--- a/tests/decode.cpp
+++ b/tests/decode.cpp
@@ -98,8 +98,10 @@ int main(int argc, char** argv)
     }
 
     if (input->getWidth() && input->getHeight()) {
-        configBuffer.surfaceWidth = input->getWidth();
-        configBuffer.surfaceHeight = input->getHeight();
+        configBuffer.width = input->getWidth();
+        configBuffer.height = input->getHeight();
+        if (!output->setVideoSize(input->getWidth(), input->getHeight()))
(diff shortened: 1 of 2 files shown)


tests/decode.cpp
```diff
@@ -103,6 +103,13 @@ int main(int argc, char** argv)
     }
 
     status = decoder->start(&configBuffer);
+
+    formatInfo = decoder->getFormatInfo();
+    if (formatInfo->width && formatInfo->height) {
+        if (!output->setVideoSize(formatInfo->width, formatInfo->height))
```
xuguangxin: could we put setVideoSize to this if?
if (input->getWidth() && input->getHeight()) {
        configBuffer.surfaceWidth = input->getWidth();
        configBuffer.surfaceHeight = input->getHeight();
    }

zhongcong: @xuguangxin the width & height set to video size is aligned. if put setVideoSize to previous if, we should do the align for the width and height.

zhongcong:
bat test h264&vp9 pass.
vp8 regressionTest will let the QA to do it.

xuguangxin:
+1

xuguangxin:
merged as dbcd4e853855d5a38136886c49d3d377b1c692aa

PR #310, opened 2015-09-08
Add hevc encoding support and ipb mode setting

lizhong1008:


encoder/vaapiencoder_base.cpp
```diff
@@ -51,6 +51,7 @@ VaapiEncoderBase::VaapiEncoderBase():
     m_videoParamCommon.frameRate.frameRateNum = 30;
     m_videoParamCommon.frameRate.frameRateDenom = 1;
     m_videoParamCommon.intraPeriod = 15;
+    m_videoParamCommon.ipPeriod = 1;
```
xuguangxin: better change it to 2. IBBP will export more potential issue than IBP.

encoder/vaapiencoder_hevc.cpp
```diff
@@ -0,0 +1,1648 @@
...
+Encode_Status VaapiEncoderHEVC::setParameters(VideoParamConfigType type, Yami_PTR videoEncParams)
+{
+    Encode_Status status = ENCODE_INVALID_PARAMS;
+    AutoLock locker(m_paramLock);
+
+    FUNC_ENTER();
+    if (!videoEncParams)
+        return ENCODE_INVALID_PARAMS;
+    switch (type) {
+    case VideoParamsTypeAVC: {
```
xuguangxin: typo?

encoder/vaapiencoder_hevc.cpp
```diff
@@ -0,0 +1,1648 @@
...
+            VideoParamsAVC* avc = (VideoParamsAVC*)videoEncParams;
+            if (avc->size == sizeof(VideoParamsAVC)) {
+                PARAMETER_ASSIGN(*avc, m_videoParamAVC);
+                status = ENCODE_SUCCESS;
+            }
+        }
+        break;
+    case VideoConfigTypeAVCIntraPeriod: {
```
xuguangxin: typo? or we change change it to VideoConfigTypeIntraPeriod

encoder/vaapiencoder_hevc.cpp
```diff
@@ -0,0 +1,1648 @@
...
+            VideoConfigAVCIntraPeriod* intraPeriod = (VideoConfigAVCIntraPeriod*)videoEncParams;
+            if (intraPeriod->size == sizeof(VideoConfigAVCIntraPeriod)) {
+                m_videoParamAVC.idrInterval = intraPeriod->idrInterval;
+                m_videoParamCommon.intraPeriod = intraPeriod->intraPeriod;
```
xuguangxin: where is ipPeroid

encoder/vaapiencoder_base.cpp
xuguangxin: do we need change to this to another name since it means b freames between non bframe.

encoder/vaapiencoder_hevc.cpp
```diff
@@ -0,0 +1,1648 @@
...
+                m_refList0.pop_back();
+        } else {
+            /* set backward reflist */
+            m_refList1.push_front(m_refList[i]);
+            if (m_refList1.size() > m_maxRefList1Count)
+                m_refList1.pop_back();
+        }
+    }
+
+    ShortRfsUpdate(picture);
```
xuguangxin: better change it to shortRfsUpdate

encoder/vaapiencoder_hevc.h
```diff
@@ -0,0 +1,178 @@
...
+#include <deque>
+#include <pthread.h>
+#include <va/va_enc_hevc.h>
+
+namespace YamiMediaCodec{
+class VaapiEncPictureHEVC;
+class VaapiEncoderHEVCRef;
+class VaapiEncStreamHeaderHEVC;
+
+typedef struct shortRFS
```
xuguangxin: could we change this to a nested class of  VaapiEncoderHEVC. and typedef is not needed

tests/encodehelp.h
```diff
@@ -166,6 +172,13 @@ void ensureInputParameters()
     if (!fps)
         fps = 30;
 
+    if (ipbMode == 0)
```
xuguangxin: could we let user set ipPeroid instead hard code value here?

encoder/vaapiencoder_hevc.cpp
lizhong1008: not typo. HEVC reuse some avc interface right now

encoder/vaapiencoder_hevc.cpp
lizhong1008: ipPeroid is common paramters. You can chech how I set m_BframeNum

tests/encodehelp.h
lizhong1008: Aactually, my origial idea same as you, but finally I changed my mind because: 1. ipb_mode is more easy to know what it means. 2. ipb_mode is similar to avcenc's mode, it will be easier to compare yamienc and avcenc. 3. As our dirver and yami implemetaion, if user set a lager ipPerid, it will bring very obvious quality decline than IP mode, so ipPeriod eqaul to 3 is my recommend. User may refer our setting (IBBPBBP) after he read our test case. On the other hand, he can set ipPeriod to other values because we provide ipPeriod not ipbMode as yami interface.

encoder/vaapiencoder_base.cpp
lizhong1008: ipPeriod = 1 means IP mode without B frame. That's default case. This name follows common codecs as I know.

encoder/vaapiencoder_hevc.cpp
lizhong1008: Yeah, agree. Thanks

encoder/vaapiencoder_hevc.cpp
xuguangxin: could we typedef it to hevc? or rename it to VideoParamsType and VideoParams*. It may makes user confused.

PR #315, opened 2015-09-16
h264dec: tidy code

clearlylin:
As for rearranging the logic. the two lines are extra.
the code at line 127 has same effect.


xuguangxin:
merged as 6456e2383dac3a58f1983b7da8ac885b639451a7

PR #316, opened 2015-09-16
v4l2dec: fixed compiling error

andyLiuChina:
when the compile opton "strict-aliasing" is actively, gcc 3 manual
says "an object of one type is assumed never to reside at the same
address as an object of a different type, unless the types are
almost the same".


xuguangxin:
+1

xuguangxin:
merged as d9759e467682467f1b2be974e9fa2ffb850f43ed

PR #317, opened 2015-09-21
when using DRM try /dev/dri/renderD128 before /dev/dri/card0

jsorg71:
Working with QuickSync on servers, I noticed sometimes libyami does not work but vainfo does work.
These servers have more then one video card.
Looking at vainfo, first it tries to open /dev/dri/renderD128, then /dev/dri/card0
This PR does similar to libva, see libva/test/common/va_display_drm.c.

### DIFF
--- a/vaapi/vaapidisplay.cpp
+++ b/vaapi/vaapidisplay.cpp
@@ -128,7 +128,9 @@ class NativeDisplayDrm : public NativeDisplayBase{
         if (acceptValidExternalHandle(display))
             return true;
 
-        m_handle = open("/dev/dri/card0", O_RDWR);
+        m_handle = open("/dev/dri/renderD128", O_RDWR);
+        if (m_handle < 0)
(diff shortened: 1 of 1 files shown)


xuguangxin:
Hi jay:
So your card0 is nv card?

jsorg71:
No, in this case it's ASPEED.
Motherboard, SuperMicro X10SLH-F.
This is very common server motherboard.

jsorg71:
Another issue I seen with using /dev/dri/cardx is that only one app can have it open at a time.
With /dev/dri/renderD128, I can access it from many apps at one time.
I'm not sure if that is by design.

xuguangxin:
not sure it's by design or not. But your patch is good to me. thanks for contribute
+1

xuguangxin:
merged as 9505821c3fe51b8a9cc986f874711cd33e3b30e5

PR #321, opened 2015-09-22
Modify function setRotation and let decoding continue even if it failed

lizhong1008:


xuguangxin:
+1

xuguangxin:
merged as a0dd6b2a6caaae5c0d2747b7479d8d010faa2fab

PR #322, opened 2015-09-25
h264_dec: fix compilation error with datatype

chivakker:
m_prevPictStructure is always assigned enumeration values while
it is defined as boolean.  Redefining it to uint32_t to avoid
compilation issues

The error is the following

vaapidecoder_h264.cpp:360:33: error: comparison of constant '2' with boolean expression is always false [-Werror=bool-compare]
             (m_prevPicStructure == VAAPI_PICTURE_STRUCTURE_BOTTOM_FIELD ?

### DIFF
--- a/decoder/vaapidecoder_h264.h
+++ b/decoder/vaapidecoder_h264.h
@@ -304,7 +304,7 @@ class VaapiDecoderH264:public VaapiDecoderBase {
     int32_t m_prevFrameNum;     // prevFrameNum
     bool m_prevPicHasMMCO5;     // prevMMCO5Pic
     uint32_t m_progressiveSequence;
-    bool m_prevPicStructure;    // previous picture structure
+    uint32_t m_prevPicStructure;    // previous picture structure
     int32_t m_frameNumOffset;   // FrameNumOffset
 
   private:


xuguangxin:
wired. Just git blame this line.It exist from day one. We should identify this since we treat all compile warning as error.So you have more restricted compile flags?

chivakker:
I am compiling it on Fedora 22 with g++ version 5.1.1.  Per the g++ (gcc) man page:
===
-Wbool-compare
           Warn about boolean expression compared with an integer value different from "true"/"false".  For instance, the following comparison is always false:

                   int n = 5;
                   ...
                   if ((n > 1) == 2) { ... }

           This warning is enabled by -Wall.
===

Compiling it on Ubuntu 15.04 with g++ version 4.9.2 the warning is not available.  Although on both Fedora and Ubuntu the default building instruction includes -Wall.

/bin/bash ../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I.. -I.. -I../interface -I../common -I../vaapi -I../codecparsers  -I/home/$USERNAME/include  -I/home/$USERNAME/include  -fpermissive  -I/home/$USERNAME/include    -g -O2 -Wno-unused-function -Wno-sign-compare -Wno-cpp -Wall -Werror -MT libyami_decoder_la-vaapidecoder_h264.lo -MD -MP -MF .deps/libyami_decoder_la-vaapidecoder_h264.Tpo -c -o libyami_decoder_la-vaapidecoder_h264.lo `test -f 'vaapidecoder_h264.cpp' || echo './'`vaapidecoder_h264.cpp

 The patch proposed will not affect the behavior on older versions of g++ and will fix an error on newer versions.

xuguangxin:
ok, +1

xuguangxin:
merged as 37958b645953498c749c6d85be56eb6ccdf7e048, thanks daniel

PR #324, opened 2015-09-25
vaapidisplay: fixed terminate vadisplay incorrectly

andyLiuChina:
when the display type is native, we do not need to do terminate.


xuguangxin:
+1

xuguangxin:
merged as 201470202712eee3474acc0c1f778a4b560ad427

PR #329, opened 2015-09-30
h264enc: Add B frame encoding support

lizhong1008:


encoder/vaapiencoder_h264.cpp
```diff
@@ -751,8 +751,12 @@ void VaapiEncoderH264::resetParams ()
     DEBUG("resetParams, ensureCodedBufferSize");
     ensureCodedBufferSize();
 
-    //FIXME:
-    m_numBFrames = 0;
+    if (ipPeriod() == 0)
+        m_videoParamCommon.intraPeriod = 1;
+    else if (ipPeriod() >= 1)
```
xuguangxin: better use "else" instead of "else if" since ipPeroid is uint32_t

encoder/vaapiencoder_h264.cpp
```diff
@@ -1215,29 +1214,29 @@ bool VaapiEncoderH264::ensurePictureHeader(const PicturePtr& picture, const VAEn
     return true;
 }
 
-static void fillReferenceList(VAEncSliceParameterBufferH264* slice, const vector<ReferencePtr>& refList, uint32_t index)
+bool VaapiEncoderH264::fillReferenceList(VAEncSliceParameterBufferH264* slice) const
```
xuguangxin: why we need change this code. Original one seems better, it did not duplicate business logical

encoder/vaapiencoder_h264.cpp
```diff
@@ -912,38 +916,32 @@ Encode_Status VaapiEncoderH264::reorder(const SurfacePtr& surface, uint64_t time
...
-    picture->m_poc = ((m_curPresentIndex * 2) % m_maxPicOrderCnt);
 
     bool isIdr = (m_frameIndex == 0 ||m_frameIndex >= keyFramePeriod() || forceKeyFrame);
 
-    /* check key frames */
-    if (isIdr || (m_frameIndex % intraPeriod() == 0)) {
-        ++m_curFrameNum;
-        ++m_frameIndex;
-        /* b frame enabled,  check queue of reorder_frame_list */
-        if (m_numBFrames
```
xuguangxin: we need consider one corner case. suppose we have n b frame and next frame is idr. We need change nth b frame to P frame and push it to front of reorder list.

encoder/vaapiencoder_h264.h
```diff
@@ -118,9 +114,12 @@ class VaapiEncoderH264 : public VaapiEncoderBase {
     AVCStreamFormat m_streamFormat;
     uint32_t m_frameIndex;
     uint32_t m_curFrameNum;
-    uint32_t m_curPresentIndex;
+
     /* reference list */
-    std::list<ReferencePtr> m_refList;
+    std::deque<ReferencePtr> m_refList;
+    std::deque<ReferencePtr> m_refList0;
```
xuguangxin: make m_refList0 1 local is better, only few function will use it.

encoder/vaapiencoder_h264.cpp
```diff
@@ -1278,14 +1273,13 @@ bool VaapiEncoderH264::addSliceHeaders (const PicturePtr& picture,
...
         sliceParam->pic_order_cnt_lsb = picture->m_poc;
 
-        if (picture->m_type != VAAPI_PICTURE_TYPE_I && refList0.size() > 0)
-            sliceParam->num_ref_idx_l0_active_minus1 = refList0.size() - 1;
-        if (picture->m_type == VAAPI_PICTURE_TYPE_B && refList1.size() > 0)
-            sliceParam->num_ref_idx_l1_active_minus1 = refList1.size() - 1;
-
-        fillReferenceList(sliceParam, refList0, 0);
-        fillReferenceList(sliceParam, refList1, 1);
+        sliceParam->num_ref_idx_active_override_flag = 1;
```
xuguangxin: no need override it since we always clip it to m_maxRefList0Count

encoder/vaapiencoder_h264.cpp
xuguangxin: and do not  forget update reference in this case.

xuguangxin:
+1

YuJiankang:
+1

PR #331, opened 2015-10-09
serveral improvement: version,configure,installation 

clearlylin:
Several Improvement
h264dec:  addressiong dropped frames. 
version: creating libyami_version.h
configure: according to va header file, building
install: arranging the installation directory


clearlylin:
OK.
Have been modified according to requirements.

clearlylin:
OK.
having updated the pkgconfig/*.pc.

PR #332, opened 2015-10-10
tests: fix static compilation

### LINKED ISSUE #330: registry factory is not work in static link mode.
1. configure yami with this configuration: ./configure --with-pic --enable-static --disable-shared
2. run "yamidecode -i xxx.264"
you will see 
"libyami error(vaapidecoder_host.cpp, 55): Failed to create decoder for mimeType: 'video/h264'"
seem h264 decoder did not registered.

after search the web, I find it maybe related to static member's init problem. c++ do not guarantee the static member's init order. So maybe VaapiEncoderH264::s_registered init before Factory<T>::m_creators.

Hi @uartie : do you have better way to fix this? Else we may need fallback to old way.

uartie:
When linking a static library, ld strips symbols that it
thinks aren't being used.  The factory model for registering
codecs fools the linker into thinking the static registration
does not result in usage of the result.

The only way to fix this without changing the factory
registration model is to force the linker to include
the entire static library archive.

Fixes #330

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,6 +37,9 @@ LIBYAMI_LT_LDFLAGS="-version-number $LIBYAMI_LT_VERSION"
 AC_SUBST(LIBYAMI_LT_VERSION)
 AC_SUBST(LIBYAMI_LT_LDFLAGS)
 
+AM_CONDITIONAL(BUILD_STATIC,
+    [test "x$enable_static" = "xyes"])
+
 AC_ARG_ENABLE(debug,
     [AC_HELP_STRING([--enable-debug],
         [build with extra debug @<:@default=no@:>@])],
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -39,6 +39,13 @@ YAMI_DECODE_LIBS 	= \
 	$(YAMI_COMMON_LIBS)                            	\
 	$(top_builddir)/decoder/libyami_decoder.la      \
 	$(NULL)
+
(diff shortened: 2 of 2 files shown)


xuguangxin:
Hi uarite:
This means somebody what use static library need manually link *.la.Is it possible to use it in more automatic way? such as add some thing on pkg-config(I don't know)?

uartie:
Actually, I think this patch is a little bit of an overkill... I don't think we need the "--whole-archive" nonsense.  Instead we only need to specify the *.a in LDFLAGS when doing a static compile.  It seems the linker does the right thing when linking the *.a and doesn't strip out the registered codecs... it's the *.la linking that confuses the linker.

For external users of the static libraries, a simple "-lyami_decoder" works correctly.  That is, the *.a is linked automatically.

I'll abandon this change and submit a simpler patch as explained.

uartie:
Oh snap!  Please disregard my last comment... I forgot to pass "--enable-static" when I was testing that theory.  I'll look into this some more to see if there are better ways to fix this.

uartie:
Ok, yes, it looks like we can fix this up for pkg-config, too, via the Libs.private and Requires definitions along with the AX_PKG_CHECK_MODULES autoconf macro.  Then, user can do `pkg-config --cflags --libs --static libyami_decoder`.

I've started working on that and should have some patches for you sometime next week.

xuguangxin:
ok, tanks for your great help!

PR #333, opened 2015-10-12
Fix Static Library Support

### LINKED ISSUE #330: registry factory is not work in static link mode.

uartie:
These patches fix several issues when external programs try to link to static libyami_* libraries.  Also, fixes the static linking for the internal tests and examples.

External programs should specify `--static` to `pkg-config` when linking to libyami_* static libraries.  For example, `pkg-config --cflags --libs --static libyami_decoder`.

Fixes #330

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,6 +37,9 @@ LIBYAMI_LT_LDFLAGS="-version-number $LIBYAMI_LT_VERSION"
 AC_SUBST(LIBYAMI_LT_VERSION)
 AC_SUBST(LIBYAMI_LT_LDFLAGS)
 
+AM_CONDITIONAL(BUILD_STATIC,
+    [test "x$enable_static" = "xyes"])
+
 AC_ARG_ENABLE(debug,
     [AC_HELP_STRING([--enable-debug],
         [build with extra debug @<:@default=no@:>@])],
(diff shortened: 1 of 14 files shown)


xuguangxin:
Hi @uartie : 
This is only for fix registry factory? Or it will benefit other things? If It's just workaround registry factory, maybe we do too much things.
and after this applied, outer user need AX_PKG_CHECK_MODULES too?

uartie:
@xuguangxin, this fixes numerous other pkgconfig problems that I found with external linking of libyami_* static libs.  Those had to be fixed in order to get down to fixing the factory static linking issue.  External users *do not* need the AX_PKG_CHECK_MODULES macro.

xuguangxin:
OK, +1.

xuguangxin:
merged as db92db0dabde8bc9567e407e8b902aa731188f07, thanks very much uartie.

PR #335, opened 2015-10-19
GCC -fpermissive fixes

uartie:


chivakker:
I tested this for building.  I took first the patch that removes -fpermissive on Ubuntu and indeed h.265 decoder build was broken, then I applied the decoder/h265 patch.  Same happened with encoder/vp8 patch.  This is Ubuntu 15.10 with gcc version (Ubuntu 4.9.2-10ubuntu13). I did not do any functional testing on the affected codecs.

chivakker:
The patch series itself lgtm +1

xuguangxin:
+1, thanks uartie and daniel

xuguangxin:
merged as 9ff3d14702b29cb1662db17776ca6da6c8ac575a
thanks all

PR #336, opened 2015-10-20
serveral Improvement: version, configure,installation

clearlylin:


clearlylin:
OK.
update the commit message.

PR #338, opened 2015-10-29
display fps on screen & use one dedicate flip thread for all instance.

xuguangxin:


examples/grid.cpp
```diff
@@ -54,9 +55,31 @@ extern "C" {
 #include "common/common_def.h"
 #include "common/condition.h"
 #include "common/lock.h"
+#include "font.h"
 
 using namespace std;
 
+class Timer
```
uartie: It doesn't look like this class is ever used.

examples/grid.cpp
```diff
@@ -54,9 +55,31 @@ extern "C" {
...
+{
+public:
+    Timer(const char* str):m_str(str)
```
uartie: In C++, probably best to avoid raw pointers when possible.

    Timer(const std::string& str)

examples/grid.cpp
```diff
@@ -54,9 +55,31 @@ extern "C" {
...
+    {
+        timeval end, diff;
+        gettimeofday(&end, NULL);
+        timersub(&end, &start, &diff);
+        if (diff.tv_usec > 999)
+            ERROR("%s: %d ms", m_str, (int)diff.tv_usec/1000);
+    }
+private:
+    timeval start;
+    const char* m_str;
```
uartie: const std::string& m_str;

examples/grid.cpp
xuguangxin: we use this to measure time 
such as 
{
    Timer timer("operation")
    for (int i = 0; i < 30; i++) {
        operation();
    }
}

examples/grid.cpp
xuguangxin: It's intended to. we do not want wast time to construct string. Since we always send const char* to constructor.

examples/grid.cpp
uartie: ok.

examples/grid.cpp
uartie: Yes, I guess it's fine in cases like this.

seanvk:
@xuguangxin Are you going to abandon this one or keep it?

xuguangxin:
Hi @seanvk 
It's need more time to rebase and make it compile able in apache branch. But this patch is useful. It will add fps display on our video wall application. I will treat this as P3, rebase it when I have time. However, it should no late than Q4.

seanvk:
Okay, thanks for the update @xuguangxin

lizhong1008:
@xuguangxin ,any update or plan for this pull request?

xuguangxin:
i have a personal tree for this.https://github.com/xuguangxin/libyami-utils/tree/10bitoutput. It include 10 bit related things. I will sent for review next week.

xuguangxin:
close this since we have https://github.com/01org/libyami-utils/pull/49

PR #343, opened 2015-11-04
vaapicontext: check profile before create vacontext

lizhong1008:
1. Check profile compatible before create va vacontext
2. For h264, driver may declare support higher profile but don't support
lower profile. In this case, higher profile should be created

### DIFF
--- a/tests/decodeinputavformat.cpp
+++ b/tests/decodeinputavformat.cpp
@@ -103,7 +103,7 @@ bool DecodeInputAvFormat::getNextDecodeUnit(VideoDecodeBuffer &inputBuffer)
     int ret;
     while (1) {
         //free old packet
-        av_free_packet(&m_packet);
+        av_packet_unref(&m_packet);
 
         ret = av_read_frame(m_format, &m_packet);
         if (ret) {
@@ -129,7 +129,7 @@ const string& DecodeInputAvFormat::getCodecData()
 DecodeInputAvFormat::~DecodeInputAvFormat()
 {
     if (m_format) {
-        av_free_packet(&m_packet);
+        av_packet_unref(&m_packet);
         avformat_close_input(&m_format);
     }
 


vaapi/vaapicontext.cpp
```diff
@@ -25,10 +25,63 @@
...
+#include <algorithm>
+#include <vector>
 
 namespace YamiMediaCodec{
+using std::vector;
+static const VAProfile h264ProfileList[] = {VAProfileH264ConstrainedBaseline, VAProfileH264Main, VAProfileH264High};
+
+//Driver may declare support higher profile but don't support lower profile.
+//In this case, higher profile should be created
+static bool checkH264Profile(VAProfile& profile, vector<VAProfile> profileList)
```
xuguangxin: better use "const vector<VAProfile>& profileList" else it will make a copy of profileList

vaapi/vaapicontext.cpp
```diff
@@ -25,10 +25,63 @@
...
+    vector<VAProfile>h264Profiles(h264ProfileList, h264ProfileList + N_ELEMENTS(h264ProfileList));
+    vector<VAProfile>::iterator start = std::find(h264Profiles.begin(), h264Profiles.end(), profile);
+    vector<VAProfile>::iterator result = std::find_first_of(profileList.begin(), profileList.end(), start, h264Profiles.end());
+
+    if (result != profileList.end()){
+        profile = *result;
+    } else{
+        // VAProfileH264Baseline is super profile for VAProfileH264ConstrainedBaseline
+        // old i965 driver incorrectly claims supporting VAProfileH264Baseline, but not VAProfileH264ConstrainedBaseline
+        if (profile == VAProfileH264ConstrainedBaseline && std::count(profileList.begin(), profileList.end(), VAProfileH264Baseline))
```
xuguangxin: better use find, count will iterate entire profileList

vaapi/vaapicontext.cpp
```diff
@@ -38,16 +91,13 @@ ConfigPtr VaapiConfig::create(const DisplayPtr& display,
...
 
-    // VAProfileH264Baseline is super profile for VAProfileH264ConstrainedBaseline
-    // old i965 driver incorrectly claims supporting VAProfileH264Baseline, but not VAProfileH264ConstrainedBaseline
-    if (vaStatus == VA_STATUS_ERROR_UNSUPPORTED_PROFILE
-        && profile == VAProfileH264ConstrainedBaseline
-        && entry == VAEntrypointVLD)
-        vaStatus = vaCreateConfig(display->getID(),
-                                  VAProfileH264Baseline,
-                                  VAEntrypointVLD, attribList, numAttribs, &config);
+    if (!checkProfileCompatible(display, profile)){
```
xuguangxin: better check it's vld or not, since profile up cast may cause problem when you use encoder.

vaapi/vaapicontext.cpp
```diff
@@ -25,10 +25,63 @@
...
+    if (!checkVaapiStatus(vaStatus, "vaQueryConfigProfiles"))
+        return false;
+    assert((numProfiles > 0) && (numProfiles <= maxNumProfiles));
+
+    profileList.resize(numProfiles);
+
+    if (profile == VAProfileH264ConstrainedBaseline || profile == VAProfileH264Main){
+        if (!checkH264Profile(profile, profileList))
+            return false;
+    } else if (!std::count(profileList.begin(), profileList.end(), profile))
```
xuguangxin: better use std::find too.

vaapi/vaapicontext.cpp
lizhong1008: I give up checking vld in this version because I think encoding may also need set higher profile. For example, if encoding driver declares support Main Profile, of course it can support constrained baseline profile.

vaapi/vaapicontext.cpp
lizhong1008: Agree, thanks for your suggestion.

vaapi/vaapicontext.cpp
lizhong1008: I know find is more efficiency than count generally speaking. but still I use count instead of find in last verison, because profile list is short,  and using count need less code line, so there is no obvious efficiency gap.

vaapi/vaapicontext.cpp
xuguangxin: for encoder we'd better think from other aspect. if they want encode a baseline profile. We give him a main profile. the target decoder may failed on encoded file.

lizhong1008: If upper layer want to constrained baseline profile, we won't create main profile unless driver declare main profile supported but constrained baseline unsupported. On the other hand, constrained baseline profile still will be built into bitstream by yami instead of driver. It won't affect decoder.
If we add vld limitation, encoder will fail in such case.

PR #348, opened 2015-11-13
factory: segment fault.

clearlylin:
initialization of static variable m_creators of factory.h is not called.
remove the static member variable m_creators.
use private static member function.

### DIFF
--- a/common/factory.h
+++ b/common/factory.h
@@ -44,7 +44,7 @@ public:
     static bool register_(const KeyType& key)
     {
         std::pair<iterator, bool> result =
-            m_creators.insert(std::make_pair(key, create<C>));
+            getCreators().insert(std::make_pair(key, create<C>));
         return result.second;
     }
 
@@ -55,8 +55,9 @@ public:
      */
     static Type create(const KeyType& key)
     {
-        const const_iterator creator(m_creators.find(key));
-        if (creator != m_creators.end())
+        Creators& creators = getCreators();
+        const const_iterator creator(creators.find(key));
+        if (creator != creators.end())
             return creator->second();
         return NULL;
     }
@@ -67,10 +68,11 @@ private:
     {
         return new C;
     }
-
-    static Creators m_creators;
+    static Creators& getCreators()
+    {
+        static Creators creators;
+        return creators;
+    }
 };
-
-template < class T > typename Factory<T>::Creators Factory<T>::m_creators;
 
 #endif // __COMMON_FACTORY_H__


common/factory.h
```diff
@@ -55,8 +55,9 @@ class Factory {
      */
     static Type create(const KeyType& key)
     {
-        const const_iterator creator(m_creators.find(key));
-        if (creator != m_creators.end())
+        Creators creators = getCreators();
```
xuguangxin: better use reference: Creators&

xuguangxin:
better give more description details in commit log.

common/factory.h
uartie: indeed!

xuguangxin:
Hi @uartie , yes it's very wired
1. yamidecode only creash on some env.
2. we create a mini reproduce project(two file, factory.cpp & test.cpp), it will crash on my machine. most wired thing, is yamidecode will not crash on my machine.
3. It's only happen when you have static member on template class. no template do not have this problem.
Not sure it's compiler's bug or not.
HI @clearlylin , could you send you mini test project to @uartie ?

clearlylin:
Ok
having changed, using reference.
Only create a simple test.cpp file.for example:

#include<stdio.h>
#include "factory.h"
class TestClass{};
static bool isR=Factory<TestClass>::register_<TestClass>("testclas");
int main()
{
  if (isR)
	  printf("is register\n");
  printf("test.cpp\n");
}

uartie:
>#include "factory.h"
>class TestClass{};
>static bool isR=Factory::register_("testclas");
>int main()
>{
>if (isR)
>printf("is register\n");
>printf("test.cpp\n");
>}

@clearlylin, this test code doesn't compile for me... and rightfully so.  I would be surprised if it did compile since the Factory is a "template" and you can't use it directly like you are here.  What is your compiler and version?

uartie:
Ok, I was able to extrapolate the above test code like so, to get it to compile:

    #include <cstdio>
    #include "factory.h"
    class TestClass {};
    static const std::string cls("testclas");
    static bool isR = Factory<TestClass>::register_<TestClass>(cls);
    int main() {
      if (isR)
        printf("is register\n");
      printf("test.cpp\n");
    }

...I get the SIGSEGV too!  Turns out this is a problem commonly referred to as the "static initialization fiasco".  Luckily, libyami proper has not encountered this problem so far.

Great catch @clearlylin !  The way you've fixed it in these patches seems to be the right way to do it, too, and this will guarantee a proper order of initialization.

Might be good to explain this more in the commit message.  Also, would it make sense to squash the two commits into one commit?

Cheers.  +1

clearlylin:
@xuguangxin 
having changed.

xuguangxin:
+1

xuguangxin:
merged as 288028a98f4c57ef29c4403895b8d9dbbd98931f

xuguangxin:
thanks uarite and clearlin.

Issue #351, opened 2015-11-17
h264 decoder issue exposed during build


halleyzhao:
Ubuntu 15.10 + braswell
it seems to be a real bug.

vaapidecoder_h264.cpp: In member function 'void YamiMediaCodec::VaapiDecoderH264::initPicturePOC0(const PicturePtr&, const SliceHeaderPtr&)':
vaapidecoder_h264.cpp:360:33: error: comparison of constant '2' with boolean expression is always false [-Werror=bool-compare]
             (m_prevPicStructure == VAAPI_PICTURE_STRUCTURE_BOTTOM_FIELD ?
                                 ^
mv -f .deps/libyami_decoder_la-vaapidecoder_host.Tpo .deps/libyami_decoder_la-vaapidecoder_host.Plo

xuguangxin:
Hi @halleyzhao : 
Did you use a old version of libyami? It should be fixed by daniel's patch https://github.com/01org/libyami/commit/37958b645953498c749c6d85be56eb6ccdf7e048

halleyzhao:
I uses the release branch.

maybe I should use master instead.

xuguangxin:
yes. this because we do not release new version yet.

xuguangxin:
close this since it's not a problem

PR #352, opened 2015-11-17
h264dec: fix a potential complie error

lizhong1008:
this patch is to fix https://github.com/01org/libyami/issues/351


xuguangxin:
why we need this. This seems same to me

lizhong1008:
why same?

xuguangxin:
== have higher priority. So it always evaluate == then ?:

xuguangxin:
add "()" or not is same thing,I think.

lizhong1008:
sorry, I mix the priority of "=" with "==".

PR #354, opened 2015-11-17
v4l2decode: fix some build issue when there is no x11

halleyzhao:
- include VideoCommonDefs.h without prefix since the folder name
  is different in original source and installed dir

- define V4L2_EVENT_RESOLUTION_CHANGE if there is not it yet, in v4l2_wrapper.h


xuguangxin:
+1

xuguangxin:
merged as 1dddbab5bd05a09f01f7e833d8937ffa270e2fd1

xuguangxin:
thanks halley.

PR #355, opened 2015-11-17
Refactoring DecodeOutput

clearlylin:


clearlylin:
v4l2decode.cpp related to decodehelp.h, and used the global vairables.
so now can not merge.

PR #358, opened 2015-11-18
Using the new getOutput of Vaapidecoderbase

clearlylin:
use the new getOutput  to replace the old one and modify related code.
remove the global variables.

### DIFF
--- a/capi/VideoDecoderCapi.cpp
+++ b/capi/VideoDecoderCapi.cpp
@@ -26,7 +26,22 @@
 #include "VideoDecoderInterface.h"
 #include "VideoDecoderHost.h"
 
+#include <stdlib.h>
+
 using namespace YamiMediaCodec;
+
+class SharedPtrHold
+{
+public:
+    SharedPtrHold(SharedPtr<VideoFrame> frame):frame(frame){}
+private:
+    SharedPtr<VideoFrame> frame;
+};
+
+static void freeHold(VideoFrame* frame)
+{
+    delete (SharedPtrHold*)frame->user_data;
+}
 
 DecodeHandler createDecoder(const char *mimeType)
 {
@@ -69,27 +84,18 @@ Decode_Status decode(DecodeHandler p, VideoDecodeBuffer *buffer)
         return DECODE_FAIL;
 }
 
-const VideoRenderBuffer* decodeGetOutput(DecodeHandler p, bool draining)
+VideoFrame* getOutput(DecodeHandler p)
 {
-    return (p ? ((IVideoDecoder*)p)->getOutput(draining) : NULL);
-}
-
-#ifdef __ENABLE_X11__
-Decode_Status decodeGetOutput_x11(DecodeHandler p, Drawable draw, int64_t *timeStamp
-        , int drawX, int drawY, int drawWidth, int drawHeight, bool draining
-        , int frameX, int frameY, int frameWidth, int frameHeight)
-{
-    if(p)
-        return ((IVideoDecoder*)p)->getOutput(draw, timeStamp, drawX, drawY, drawWidth, drawHeight, draining
-                        , frameX, frameY, frameWidth, frameHeight);
-    else
-        return DECODE_FAIL;
-}
-#endif
-
-Decode_Status decodeGetOutputRawData(DecodeHandler p, VideoFrameRawData* frame, bool draining)
-{
-    return (p ? ((IVideoDecoder*)p)->getOutput(frame, draining) : DECODE_FAIL);
+    if (p) {
+        SharedPtr<VideoFrame> frame = ((IVideoDecoder*)p)->getOutput();
+        if (frame) {
+            SharedPtrHold *hold= new SharedPtrHold(frame);
+            frame->user_data = (intptr_t)hold;
+            frame->free = freeHold;
+            return frame.get();
+        }
+    }
+    return NULL;
 }
 
 const VideoFormatInfo* getFormatInfo(DecodeHandler p)
--- a/capi/VideoDecoderCapi.h
+++ b/capi/VideoDecoderCapi.h
@@ -26,9 +26,7 @@
 #define __ENABLE_CAPI__ 1
 #endif
 #include "VideoDecoderDefs.h"
-#ifdef __ENABLE_X11__
-#include <X11/Xlib.h>
-#endif
+#include "VideoCommonDefs.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -48,15 +46,7 @@ void decodeFlush(DecodeHandler p);
 
 Decode_Status decode(DecodeHandler p, VideoDecodeBuffer *buffer);
 
-const VideoRenderBuffer* decodeGetOutput(DecodeHandler p, bool draining);
-
-#ifdef __ENABLE_X11__
-Decode_Status decodeGetOutput_x11(DecodeHandler p, Drawable draw, int64_t *timeStamp
-        , int drawX, int drawY, int drawWidth, int drawHeight, bool draining
-        , int frameX, int frameY, int frameWidth, int frameHeight);
-#endif
-
-Decode_Status decodeGetOutputRawData(DecodeHandler p, VideoFrameRawData* frame, bool draining);
+VideoFrame* getOutput(DecodeHandler p);
 
 const VideoFormatInfo* getFormatInfo(DecodeHandler p);
 
--- a/tests/decodeoutput.cpp
+++ b/tests/decodeoutput.cpp
@@ -3,8 +3,9 @@
  *
  *  Copyright (C) 2011-2014 Intel Corporation
  *    Author: Halley Zhao<halley.zhao@intel.com>
- *    Author: Xu Guangxin <guangxin.xu@intel.com>
- *    Author: Liu Yangbin <yangbinx.liu@intel.com>
+ *            Xu Guangxin<guangxin.xu@intel.com>
+ *            Liu Yangbin<yangbinx.liu@intel.com>
+ *            Lin Hai<hai1.lin@intel.com>
  *
  *  This library is free software; you can redistribute it and/or
  *  modify it under the terms of the GNU Lesser General Public License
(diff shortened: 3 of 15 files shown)


clearlylin:
@xuguangxin 
have changed code according to tips

seanvk:
-1

I need more information on background of these requests?  What target project or use case?

interface/VideoCommonDefs.h
```diff
@@ -25,6 +25,10 @@
 
 #include <stdint.h>
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
```
uartie: It is generally not a good idea to include config.h in headers.  There's a comment about this 4 lines above.

capi/VideoDecoderCapi.h
```diff
@@ -19,16 +19,10 @@
  *  Boston, MA 02110-1301 USA
  */
 
-#ifndef __VIDEO_DECODER_CAPI_H__
-#define __VIDEO_DECODER_CAPI_H__
+#ifndef videodecodercapi_h
+#define videodecodercapi_h
```
uartie: I prefer the previous header guard naming syntax... but it appears both styles are used throughout this project.  Question is, which style is preferred in yami?  It would be better to enforce only one policy.

capi/VideoDecoderCapi.h
```diff
@@ -36,50 +30,28 @@ extern "C" {
...
-
-void releaseDecoder(DecodeHandler p);
+void releaseDecoder(DecodeHandler);
 
 #ifdef __cplusplus
 };
 #endif
 
-#endif
+#endif
```
uartie: need newline

tests/decodeoutput.cpp
```diff
@@ -29,411 +30,299 @@
...
+        ERROR("fail to open input file: %s", m_outputFile);
         return false;
     }
-
+    MD5_Init(&m_fileMd5);
     return true;
 }
 
-bool DecodeOutputExtractMD5::render(VideoFrameRawData* frame)
+bool DecoudeOutputMD5::write(SharedPtr<VideoFrame>& frame)
```
uartie: misspelled Decode... Decoude -> Decode

tests/decodeoutput.cpp
```diff
@@ -555,115 +377,127 @@ DecodeOutputPixelMap::~DecodeOutputPixelMap()
...
+        return false;
     }
-    return ret;
+    EGLImageKHR eglImage = EGL_NO_IMAGE_KHR;
+    createEGLImage(eglImage, dest);
+
+    return draw2D(eglImage);
 }
+#endif //__ENABLE_TESTS_GLES__
+#endif //__ENABLE_X11__
```
uartie: newline

tests/decodeoutput.h
```diff
@@ -25,114 +26,218 @@
...
+    ~DecodeOutputNull() {}
+    bool init();
+    bool output(SharedPtr<VideoFrame>& frame);
+};
 
+
+class ColorConvert
+{
+public:
+    uint8_t* convert(SharedPtr<VideoFrame>& frame, SharedPtr<VADisplay> display, uint32_t pitches[3], uint32_t offsets[3])
```
uartie: use `const`when passing shared pointers by reference, unless you plan to modify the shared pointer.

seanvk:
In addition to @uartie comments, it's clear some of this was not tested.  This has to be verified as well for all configurations not just the Android use case.

capi/VideoDecoderCapi.h
uartie: https://webkit.org/code-style-guidelines/#names-header-guards

interface/VideoCommonDefs.h
uartie: https://webkit.org/code-style-guidelines/#include-config-h

tests/decodeoutput.cpp
clearlylin: really bad.
Thanks.

interface/VideoCommonDefs.h
clearlylin: But there need to use the macro, '__ENABLE_CAPI__'.
I am not sure that that is available.

tests/decodeoutput.h
clearlylin: Thanks

interface/VideoCommonDefs.h
clearlylin: OK. 
I understand it

capi/VideoDecoderCapi.h
xuguangxin: Yes , we should use webkit style guards. but some code from codec parser use capital letters as guards. It's not gone be problem any more. 
But we do not need change it in this patch. Hi @clearlylin , could we only change need line?

capi/VideoDecoderCapi.cpp
```diff
@@ -22,135 +22,88 @@
...
 #endif
-#include "VideoDecoderCapi.h"
+
 #include "VideoDecoderInterface.h"
 #include "VideoDecoderHost.h"
+#include "VideoDecoderCapi.h"
 
-using namespace YamiMediaCodec;
-
-DecodeHandler createDecoder(const char *mimeType)
```
xuguangxin: Could we change code using "diff friendly" way? Seems you changed all file, but actually only few lines added.

xuguangxin:
HI @seanvk:
This patch set will remove all getOuputs in current code, excpet SharedPtr<VideoFrame> getOutput().
just like I mentioned in this comment https://github.com/01org/libyami/pull/207#issuecomment-93186729. It's not for android, just code refact for api simplify.
It's just early code review make sure @clearlylin walk in right way. Before this code committed. we will do full regression test. Maybe @clearlylin  need mention it's a early version in pull request.

tests/decodeoutput.cpp
xuguangxin: is this compile able?

capi/VideoDecoderCapi.h
clearlylin: OK

tests/decodeoutput.cpp
clearlylin: Yes.
All name is wrong.

capi/VideoDecoderCapi.cpp
clearlylin: OK

clearlylin:
have changed.

halleyzhao:
-1
I don't think it is the right way to go.

codec lib exposes interface to other module to less the low level dependency, it should share common data; with camera by dma_buf (kernel generic concept), with opengl by drm_name/dma_buf, with common media pipeline by raw data; and maybe other generic concept data in the future.
libva/vaSurface is vendor specific driver interface, it is inconvenient to use and incorrect to be shared with external module. you should NOT expect camera/opengl/opencl module to depend on libva.

halleyzhao:
there are three getOutput(), only the following one is the direction to go:
getOutput(VideoFrameRawData* frame, bool draining = false)

the other two exists because of historical reason.
- VideoRenderBuffer *getOutput(bool draining) exposes implements details to support omx (in an ugly way)
- getOutput(unsigned long draw, int64_t *timeStamp
        , int drawX, int drawY, int drawWidth, int drawHeight, bool draining = false
        , int frameX = -1, int frameY = -1, int frameWidth = -1, int frameHeight = -1)
 this is convenient for X11 only, with additional gpu copy comparing to exporting buffer handles.

xuguangxin:
Hi Halley:
current we have 4 getOutputs in IVideoDecoder
* virtual SharedPtr\<VideoFrame\> getOutput() = 0;
* virtual const VideoRenderBuffer* getOutput(bool draining) = 0;
* virtual Decode_Status getOutput(unsigned long draw, int64_t *timeStamp
        , int drawX, int drawY, int drawWidth, int drawHeight, bool draining = false
        , int frameX = -1, int frameY = -1, int frameWidth = -1, int frameHeight = -1) = 0;
* virtual Decode_Status getOutput(VideoFrameRawData* frame, bool draining = false) = 0;


I prefer use *virtual SharedPtr\<VideoFrame\> getOutput() = 0;* to replace other 3.because

1. It makes api more simpler
   We do not need compile any X/android/wayland related thing in decoder. we just decode bitstream and output VASurfaceID. With this VASurfaceID customer can do any thing they want. They can output it to X (even in different thread). They can convert it to drm_name/dma_buf. They can map and and read surface. All just few lines libva code. not difficult to do.

2. It makes our api orthogonal  with libva's. 
If you take a close look at VideoFrameRawData or X version of *getOutput*, it just a wrapper of libva's function. It do not provide value added things.  Why not let caller call libva api directly? 

3.  It make internal  implemention and  external user much more easy. 
If we want support RGB in VideoFrameRawData. We must keep a internal RGB VAImage pool in our decoder. But how many RGB image we should keep? What if customer not want use our internal VAImage. They want  use they owned RGB meory?Key point here is, we put two different things in this api  *decode* and *vpp*. Split them to two functional block will makes us more flexible.  

4. camera/opengl/opencl module never directly operate with libyami.
 It always have some proxy,  such like v4l2decode. we can map the egl image or android native window buffer to VASurfaceID in v4l2decoder. use vpp api do color space conversion.

halleyzhao:
1. I prefer getOutput(VideoFrameRawData* frame, bool draining = false).
it is simple, it gets desired data in one call, it doesn't introduce additional dependency.

2. it is not good choice to force yami client call libva directly.
libva is vendor specific driver interface, yami's client usually are MMF/plugin or application. they needn't  touch driver interface.
the situation is similar to: opengl/opencl clients needn't call directly to libdrm, gtk application needn't call directly to x11.

3. RGB frame data is required because of hw limitation(usually). such work around should be enclosed inside vendor specific module, not pollute upper layer.
most hw doesn't have such limitation (intel will eliminate such limitation in future platform as well), and the limitation can be worked around in gfx/client side ( draw nv12 as two texture unit) with better performance.
getOutput(VideoFrameRawData* frame, ...) supports VASurface as well, if client requires additional feature, it is possible.

4. proxy is not must.
we have already seen camera with libva through dma_buf, libyami can go there directly (camera dma_buf example).
we have already seen opengl with yami through drm name (weston).
we talk about v4l2codec recently, the major reason is to standardized the interface; in technical, the proxy is not must.
even when there is a proxy, it's part of codec lib (libyami); not proxy by the client itself.

it is ok to separate color-space conversion to a separate API to avoid implicitly VAImage pool; but vaSurface is not value in else cases.

xuguangxin:
I see. Big difference between us is you think we should hide libva under libyami. But I think add libva dependent to yami caller is not problem. mediasdk did [this](https://github.com/FFmpeg/FFmpeg/blob/979572365f2133f969f3f49ec6a99cc8739d2eba/doc/examples/qsvdec.c#L90), ffmpeg did [this too]( https://github.com/videolan/vlc/blob/master/modules/codec/avcodec/vaapi.c#L429)
Because libyami's caller(the proxy i mentined) is not media framework. It should be the media framework plugin. More specifically, it's  gstreamer or v4l2 or omx plugin. Handle drm_name/dma_buf/android_native_window/malloc memory to VASurfaceID mapping in plugin level is more flexible to customer and it makes libyami more simpler.

halleyzhao:
the reference code you mentioned does vaGetDisplay/vaInitialize() as well; and many other libva details. it's not what yami's client to position.
they shrink core codec lib but provide total solution in the APP. it's the way to support app one by one, not a way to create portable lib.
context and handle are passed by opaque data with internal convention; they are limited by the data structure defined for pure sw codec. I don‘t like such tricks.

you can discuss with more team members to decide go or not.

halleyzhao:
how about separate libyami to two parts, the core part uses your proposed interface, the utility part helps the input/output variant.
take "V4l2Decoder::createVaSurface(const ANativeWindowBuffer* buf)" as example. it is not only required by v4l2codec, but also other wrappers like omx.

xuguangxin:
we can talk more in meeting

clearlylin:
@xuguangxin 

have run regression test. only h264 ,vp9

tests/decodeoutput.cpp
```diff
@@ -29,411 +30,253 @@
...
-    m_frame.memoryType = VIDEO_DATA_MEMORY_TYPE_SURFACE_ID;
-    m_frame.width = 0;
-    m_frame.height = 0;
-    m_frame.fourcc = 0;
-
-    Decode_Status status = m_decoder->getOutput(&m_frame, drain);
-    if (status == RENDER_SUCCESS)
-        vaSyncSurface((void*)(m_frame.handle), m_frame.internalID);
-    return status;
+    m_vaDisplay  = createVADisplay();
```
xuguangxin: need check return value

tests/decodeoutput.cpp
```diff
@@ -29,411 +30,253 @@
...
-void DecodeOutputRaw::setFourcc(uint32_t fourcc)
+bool DecodeOutputFile::init()
 {
-    delete m_convert;
-    m_destFourcc = fourcc;
-    if (m_destFourcc == VA_FOURCC_I420 && m_enableSoftI420Convert) {
-        m_srcFourcc = VA_FOURCC_NV12;
-    } else {
-        m_srcFourcc = m_destFourcc;
+    m_vaDisplay  = createVADisplay();
```
xuguangxin: need check return value

tests/decodeoutput.h
```diff
@@ -25,114 +26,299 @@
...
+    bool output(SharedPtr<VideoFrame>& frame);
+private:
+    XID m_pixmap;
 };
-#endif
 
-//help functions
-bool renderOutputFrames(DecodeOutput* output, uint32_t maxframes, bool drain = false);
-bool configDecodeOutput(DecodeOutput* output);
+class DecodeOutputDmabuf: public DecodeOutputEgl
```
xuguangxin: why you need add this to header?

tests/decodeoutput.h
```diff
@@ -25,114 +26,299 @@
...
+                index = ++col * 2;
+            }
+            u_src += srcImage.pitches[1];
+            u_dst += dstImage.pitches[u];
+            v_dst += dstImage.pitches[v];
+        }
+    }
+    bool initAllocator(uint32_t width, uint32_t height)
+    {
+        width = ALIGN8(width);
```
xuguangxin: why we need this align?

tests/decodeoutput.h
```diff
@@ -25,114 +26,299 @@
...
+        vaUnmapBuffer(*m_display, srcImage.buf);
+        vaDestroyImage(*m_display, srcImage.image_id);
+
+        return destFrame;
+    }
+
+    //support i420,yv12
+    void copy(uint32_t width, uint32_t height, char *src, VAImage& srcImage, char *dest, VAImage& dstImage)
+    {
+        uint32_t  planes, byteWidth[3], byteHeight[3];
```
xuguangxin: could we check src and dest format here?

tests/vppinputoutput.h
```diff
@@ -126,10 +127,15 @@ class PooledFrameAllocator : public FrameAllocator
     {
         return m_pool->alloc();
     }
-    ~PooledFrameAllocator()
+    void destroySurfaces()
     {
         if (m_surfaces.size())
             vaDestroySurfaces(*m_display, &m_surfaces[0], m_surfaces.size());
+        m_surfaces.clear();
```
xuguangxin: why we need care about this since entire object destroyed

clearlylin:
@xuguangxin 
have edited it according to your advise

clearlylin:
several vp9 videos have problem.
pull request 401 describe the problem.

xuguangxin:
this merged to #450, so close it

PR #359, opened 2015-11-19
Integrate GTest Framework

uartie:
GTest is a C++ test framework with a rich set of features that allow you to easily write tests for your project (https://code.google.com/p/googletest/).  This set of patches adds gtest as a dependency (currently optional, but required for unit test compilation).  Each libyami library will compile its own `unittest` executable, which can be individually executed directly or all of them through `make check`.  The unit tests do not require any user interaction and provides multiple types of result output formats. Therefore, they can be easily integrated into an automated testing framework.

On Fedora, the `gtest` and `gtest-devel` packages can be installed via 'yum' or 'dnf'.  I have not checked if other distros provide pre-built packages for gtest.  If they don't, then gtest can be compiled and installed from source easily.

The recommendation is that developers create unit tests to test new code they add, as appropriate.  However, this does not mean that everything can be tested with a unit test, so that judgement will need to be made on a case-by-case basis.

I also recommend that the policy be that all unit tests pass before submitting/merging code to libyami (i.e. no unit test regressions allowed).

Once this patchset is merged, we can start filling in the unit tests to cover the current code.  I've already started some of this in the following branches [unittest-common](https://github.com/uartie/libyami/commits/unittest-common), [unittest-decoder](https://github.com/uartie/libyami/commits/unittest-decoder), and [unittest-encoder](https://github.com/uartie/libyami/commits/unittest-encoder)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,6 +74,12 @@ fi
 AM_CONDITIONAL(ENABLE_TESTS_GLES,
     [test "x$enable_tests_gles" = "xyes"])
 
+HAVE_GTEST="no"
+ifdef([GTEST_LIB_CHECK],
+    [GTEST_LIB_CHECK([1.7.0], [:], [:])])
+AM_CONDITIONAL([ENABLE_UNITTESTS],
+    [test "x$HAVE_GTEST" = "xyes"])
+
 AC_ARG_ENABLE(dmabuf,
     [AC_HELP_STRING([--enable-dmabuf],
         [support dma_buf buffer sharing @<:@default=no@:>@])],
--- a/capi/Makefile.am
+++ b/capi/Makefile.am
@@ -36,6 +36,34 @@ libyami_capi_la_SOURCES     = $(libyami_capi_source_c)
 libyami_capi_la_LDFLAGS     = $(libyami_capi_ldflags)
 libyami_capi_la_CPPFLAGS    = $(libyami_capi_cppflags)
 
+if ENABLE_UNITTESTS
+noinst_PROGRAMS = unittest
+
+unittest_SOURCES = \
+	$(top_builddir)/common/unittest_main.cpp \
+	$(NULL)
+
+unittest_LDFLAGS = \
+	$(GTEST_LDFLAGS) \
+	$(NULL)
+
+unittest_LDADD = \
+	libyami_capi.la \
+	$(GTEST_LIBS) \
+	$(NULL)
+
+unittest_CPPFLAGS = \
+	$(GTEST_CPPFLAGS) \
+	$(NULL)
+
+unittest_CXXFLAGS = \
+	$(GTEST_CXXFLAGS) \
+	$(NULL)
+
+check-local: unittest
+	$(builddir)/unittest
+endif
+
 DISTCLEANFILES = \
 	Makefile.in
 
--- a/codecparsers/Makefile.am
+++ b/codecparsers/Makefile.am
@@ -69,6 +69,34 @@ libyami_codecparser_la_SOURCES     = $(libyami_codecparser_source_c)
 libyami_codecparser_la_LDFLAGS     = $(libyami_codecparser_ldflags)
 libyami_codecparser_la_CPPFLAGS    = $(libyami_codecparser_cppflags)
 
+if ENABLE_UNITTESTS
+noinst_PROGRAMS = unittest
+
+unittest_SOURCES = \
+	$(top_builddir)/common/unittest_main.cpp \
+	$(NULL)
+
+unittest_LDFLAGS = \
+	$(GTEST_LDFLAGS) \
+	$(NULL)
+
+unittest_LDADD = \
+	libyami_codecparser.la \
+	$(GTEST_LIBS) \
+	$(NULL)
+
+unittest_CPPFLAGS = \
+	$(GTEST_CPPFLAGS) \
+	$(NULL)
+
+unittest_CXXFLAGS = \
+	$(GTEST_CXXFLAGS) \
+	$(NULL)
+
+check-local: unittest
+	$(builddir)/unittest
+endif
+
 DISTCLEANFILES = \
 	Makefile.in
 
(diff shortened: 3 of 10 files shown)


configure.ac
```diff
@@ -74,6 +74,12 @@ fi
 AM_CONDITIONAL(ENABLE_TESTS_GLES,
     [test "x$enable_tests_gles" = "xyes"])
 
+HAVE_GTEST="no"
+ifdef([GTEST_LIB_CHECK],
```
xuguangxin: who will define GTEST_LIB_CHECK macro?

uartie: On Fedora, the gtest-devel package installs /usr/share/aclocal/gtest.m4... the GTEST_LIB_CHECK macro is defined in that file.  If gtest.m4 is not installed from a source install or other distros don't include it in a package, then we may need to provide it in libyami's m4 directory.  Are you able to check if other distros provide this?

xuguangxin: here is all package from:

autopkgtest - automatic as-installed testing for Debian packages
gtester2xunit - Helper for converting gtester xml output to xunit.
libgtest-dev - Google's framework for writing C++ tests - header files
libxorg-gtest-data - X.Org dummy testing environment for Google Test - data
libxorg-gtest-dev - X.Org dummy testing environment for Google Test - headers
libxorg-gtest-doc - X.org dummy testing environment for Google Test - documentation
autopkgtest-xenlvm - autopkgtest Xen/LVM2 based testbed snapshot system
gnome-desktop-testing - runner for GNOME installed tests
libdist-zilla-plugin-podspellingtests-perl - release tests for POD spelling
gnome-desktop-testing-dbgsym - debug symbols for package gnome-desktop-testing

I installed libgtest-dev, i can see all /usr/include/gtest/gtest.h, but i did find gtest.m4 in my system.
And do we have license issue if we include this .m4 in yami?

xuguangxin: Is it possible just use pkg-config to check gtest?

seanvk:
Please note @uartie that we test primarily on Ubuntu and that is what QA is using.  So you will need to also do your verification of these patches and their behavior on Ubuntu.

configure.ac
uartie: gtest doesn't provide a gtest.pc file.

Interestingly, I just found this:

https://code.google.com/p/googletest/wiki/FAQ#Why_is_it_not_recommended_to_install_a_pre-compiled_copy_of_Goog

...which implies that we should compile gtest during libyami compilation.  I'm not sure how we should manage that, however.  Would it be better to handle gtest as a git submodule, a tarball, a download-on-first-compile, or some other way?

uartie:
> Please note @uartie that we test primarily on Ubuntu and that is what QA is using. So you will need to also do your verification of these patches and their behavior on Ubuntu.

Ok, I'll make sure I get an Ubuntu system set up, too, for verifying changes.

configure.ac
chivakker: I believe we should not add gtest to the build of libyami.  It would be better if we add a README file with instructions on how to install gtest for the user and make sure that it is only available when debug is enabled.  If debug is not enabled we should not bother.

seanvk: Agree with you @chivakker

uartie: @chivakker that method makes it harder to enforce a policy to have developers write and execute unit tests for their patches.

chivakker: @uartie if I am adding a patch I am hoping the gtest is already present and I, as a developer, should make sure that such patch was tested with gtest (which can be done by QA through an automated process).  If I want to add a feature, part of my feature is to add gtest unittests, I should have discussed my feature with others before trying to add it.

uartie: Agree.  And we should make that clear in the main README for this project if that is what we expect from contributing developers.  Of course, it will also require full support from the project maintainers/gatekeepers to ensure this guideline is followed.  Hopefully that won't be an issue :smile:

uartie: I'll will come up with a revised patchset that takes all this discussion into consideration.

uartie: > make sure that it is only available when debug is enabled. If debug is not enabled we should not bother.

Actually, I disagree with this... we will want to be able to compile and execute the unit tests with both debug and non-debug since it is possible for developers to accidentally place code inside the __ENABLE_DEBUG__ guards that otherwise shouldn't be.

xuguangxin: yes.agree on this.

xuguangxin: but we should have a option to --disable gtest. to make sure no run time overhead from gtests.

uartie: I've worked on projects where the unit tests were defined directly within the same source files that were being tested.  But that can get messy real quick and requires extra care (e.g. with enable/disable flags, macros, and such) to ensure they don't pollute and add runtime overhead in the final product.

As long as the unit tests are defined in separate source files and compiled into their own binaries, there will be no additional runtime overhead or pollution in the libyami libraries, regardless of enabling gtest or not.  If you examine the makefile changes here and my WIP unit test branches (linked in the pull request summary), you'll notice there is a clear separation of the unit tests from the libyami libraries.

chivakker: On Thu, Nov 19, 2015 at 11:52 PM, U. Artie Eoff
<notifications@github.com> wrote:
>
> In configure.ac:
>
> > @@ -74,6 +74,12 @@ fi
> >  AM_CONDITIONAL(ENABLE_TESTS_GLES,
> >      [test "x$enable_tests_gles" = "xyes"])
> >
> > +HAVE_GTEST="no"
> > +ifdef([GTEST_LIB_CHECK],
>
> I've worked on projects where the unit tests were defined directly within the same source files that were being tested. But that can get messy real quick and requires extra care (e.g. with enable/disable flags, macros, and such) to ensure they don't pollute and add runtime overhead in the final product.
>
> As long as the unit tests are defined in separate source files and compiled into their own binaries, there will be no additional runtime overhead or pollution in the libyami libraries, regardless of enabling gtest or not. If you examine the makefile changes here and my WIP unit test branches (linked in the pull request summary), you'll notice there is a clear separation of the unit tests from the libyami libraries.

I meant to say this when saying that gtest should be enabled only in
debug mode.  If it is going to be necessary we should create a gtest
mode as @xuguangxin said.

uartie:
I have a local revision ready for this, but pushing it up to github will have to wait for a couple of weeks.

xuguangxin:
great! waiting for your patches.

uartie:
I am closing this because there are still some details that we are working out for gtest support that will take more time.

xuguangxin:
ok, thanks

PR #361, opened 2015-11-19
log: prompt when log isn't enabled at compile time

halleyzhao:


xuguangxin:
+1

seanvk:
lgtm +1

uartie:
What about the ERROR and ASSERT macros defined in the log.h header file?  This patch silences those macros, too, when configured with --disable-debug.  Is that intentional?

-1

halleyzhao:
seems inline comments are overwrite after a new push. 
thanks uartie, i have fixed the mentioned issue.
and add another patch to improve 64bit support

common/log.cpp
```diff
@@ -76,4 +79,7 @@ void yamiTraceInit()
         }
         isInit = 1;
     }
+#ifdef __ENABLE_DEBUG__
```
uartie: This should be `#ifndef` not `#ifdef`

common/log.cpp
```diff
@@ -76,4 +79,7 @@ void yamiTraceInit()
...
+     fprintf(stderr, "yami log isn't enabled (--enable-debug)\n");
```
uartie: I think it'd be better to only print this if the yamiLogFlag > YAMI_LOG_ERROR since YAMI_LOG_ERROR always gets logged regardless of debug or not.

common/log.h
```diff
@@ -65,7 +65,7 @@ extern int isIni;
 #define ASSERT(expr) do {                                                                                               \
         if (!(expr))                                                                                                    \
             ERROR();                                                                                                    \
-        assert(expr);                                                                                                   \
+            assert(0 && expr);                                                                                                   \
```
uartie: Did you intend to move this into the conditional body?  If so, you forgot to enclose the body in braces.

uartie:
> seems inline comments are overwrite after a new push.

Yes, sorry... I mistakenly made my first comments on the patch and not from the pull request.

common/log.cpp
halleyzhao: not catch your meaning, do you mean to below setence?

 ERROR("yami log isn't enabled (--enable-debug)\n");

common/log.h
halleyzhao: yes, it should be in the conditional body.
let me add the {} and (expr).

common/log.cpp
uartie: No, I mean:

    if (yamiLogFlag > YAMI_LOG_ERROR)
        fprintf(stderr, "yami log isn't enabled (--enable-debug)\n");

halleyzhao:
thanks uartie. updated

seanvk:
Thanks Artie.  lgtm +1

uartie:
lgtm :+1:

xuguangxin:
log: prompt when log isn't enabled at compile time …
merged as 3e0fa493697b29a51ec3d4a5757667ac6177aab8.
thanks all

halleyzhao:
hi guangxin:
here are two commits, but you merged just one

another one is "v4l2: improve 64bit platform support …"

xuguangxin:
sorry, I think you may accidentally push the commit since main title of this pull request is "log: prompt when log isn't enabled at compile time ".

PR #362, opened 2015-11-20
v4l2: fix pthread linkage

uartie:
Don't use -lpthread since it is less portable then -pthread.
Use -pthread instead with GCC.

On Ubuntu 14.04, attempting to link a program with two libraries,
where one was linked with -lpthread and one that linked with
-pthread, can cause the link phase to fail with undefined
references to various pthread functions.


xuguangxin:
+1

xuguangxin:
merged as 8c2795c29593034fc01d0f27703288ad7f441eb6.
thanks uartie

PR #363, opened 2015-11-21
improve 64bit platform support, make a copy of codec data

halleyzhao:


v4l2/v4l2_codecbase.cpp
```diff
@@ -279,12 +281,13 @@ const char* V4l2CodecBase::IoctlCommandString(int command)
 }
 #endif
 
-int32_t V4l2CodecBase::ioctl(int command, void* arg)
+int32_t V4l2CodecBase::ioctl(uint64_t command, void* arg)
```
xuguangxin: do we have problem when we compile this in chrome. since chrome may define command type to int

halleyzhao: 1. you can change vaapi_v4l2_video_device.h/cc slightly to be compatible with it:
typedef int32 (*VaapiV4L2Ioctl)(int32 fd, unsigned long cmd, ...); 
change the 'unsigned long' to 'uint64_t'.
similar some other places
2. more over, you can submit another patch to chrome on v4l2_video_device.h on
  virtual int Ioctl(int request, void* arg) = 0;

we are lucky that vaapi_v4l2 isn't broken for chrome because:
though compiler treat the ioctrl command (kernel defined macro) as 'unsigned long', its value is still inside the range of uint32_t. the conversion between 'long int' and int32_t still keeps the value part.

i found the issue when I tried printf.

xuguangxin: Since it only cause problem when you use printf. Could we make some workaround in printf format? We'd better not change v4l2_device.h,since we chrome authors is owner for this file

halleyzhao: step 2 is not must, but nice to have.
you can update intel code only in vaapi_V4L2_video_device.h/cc, since it's safe to cast chrome's   'int' to intel's 'int64_t'.

halleyzhao:
std::vector::assign() looks not so good. it is similar to memset() to init the vector with unique value.
we still have to copy the data manually.

xuguangxin:
std::vector::assign have two definitions. http://www.cplusplus.com/reference/vector/vector/assign/.
you can use this one:

template <class InputIterator>
  void assign (InputIterator first, InputIterator last);

it will make the code much more smaller

halleyzhao:
InputIterator should come from another container (vector), but we don't have it in this case.

halleyzhao:
wait a moment to merge this patch, let me ask my boss whether i can change the mail addr.

xuguangxin:
no, c array is ok.

"uint8_t *ptr = format->fmt.raw_data;
ptr += sizeof(uint32_t);
m_codecData.assign(ptr, ptr+m_configBuffer.size);"

it will takcare of all allocate and free things.

halleyzhao:
it's not about 'int', but 'long'

cat v4l2-macro-test.c 
#include <linux/videodev2.h>
#include <stdio.h>
#define __STDC_FORMAT_MACROS
#include <inttypes.h>

int main()
{
    printf("long: %d, sizeof(VIDIOC_DQBUF): %d\n", sizeof(long int), sizeof(VIDIOC_DQBUF));
    printf("VIDIOC_STREAMON:%d, %ud, %x\n", VIDIOC_STREAMON, VIDIOC_STREAMON, VIDIOC_STREAMON);
    printf("VIDIOC_DQBUF:  %d, %ud, %x\n", VIDIOC_DQBUF, VIDIOC_DQBUF, VIDIOC_DQBUF);
}

long: 8, sizeof(VIDIOC_DQBUF): 8
VIDIOC_STREAMON:1074026002, 1074026002d, 40045612
VIDIOC_DQBUF:  -1067952623, 3227014673d, c0585611

halleyzhao:
update codecdata for the vector

halleyzhao:
update the patch, if it is ok, please merge.

xuguangxin:
+1

xuguangxin:
merged as 3203d6a785dd89172c09c268860e358fd2a0e728, thanks halley

PR #365, opened 2015-11-29
fix v4l2 mem leak and define/implement generic v4l2codec operation interface

halleyzhao:


v4l2/v4l2_wrapper.cpp
```diff
@@ -206,3 +207,66 @@ int32_t YamiV4L2_SetDrmFd(int32_t fd, int drm_fd)
...
+
+    if (!key || !value) {
+        ERROR("invalue parameter\n");
+        return -1;
+    }
+
+    V4l2CodecPtr v4l2Codec = _findCodecFromFd(fd);
+    ASSERT(v4l2Codec);
+
+    if (!strcmp(key, "frame-memory-type")) {
```
xuguangxin: Do we really need so much output type for v4l2. Could we just keep one for real usage?

halleyzhao: 目前看，都有使用价值的。
1. raw-pointer类似shm，如果用户dump数据到文件，直接拿去写就可以了，避免拷贝。
2. raw-copy类似usr_ptr或者说copy，yami把数据copy到user给定的binary；这样就不用renderDone了，生命周期管理方便。
3. drm-name, 这个handle跟drm的其他模块配合比较方便，mesa/opencl
4. dma_buf,这个跟camera配合方便，camera不认drm
5. surface-id，这个兼容omxil目前的实现，虽然不建议使用。

halleyzhao: sorry, i'd better use english.
each mode has its value for now:
1. raw-pointer is similar to shm. if client dump video frames, it can used the data pointer to write output file. no additional copy.
2. raw-copy is similar to usr_ptr (copy indeed in our case), yami copy video frame data to user specific memory. then it's simple to manage the output buffer life cycle. not necessary for renderDone()
3. drm-name, it is friendly to other drm based modules, like mesa/opencl
4. dma_buf, it is friendly to camera, camera module has no knowledge on drm
5. surface-id, legacy support for the mode used by omxil, though i hate it. it brings vaapi complexity to omxil leve, is better be replaced by raw-pointer/row-copy

xuguangxin:
raw-pointer and raw-copy will force gpu/cpu sync. It's totally different  schedule mode to drm-name or dma_buf,
And drm-name dma_buf surfac-id are same thing, user can conversion by them self, We do not need provide this to customer.
add so many output method will introduce complex to test and interface.
maybe we can talk this in meeting.

interface/v4l2codec_device_ops.h
```diff
@@ -0,0 +1,83 @@
...
+typedef bool (*V4l2codecOperationInitFunc)(struct V4l2CodecOps *OpFuncs);
+// fill all the func ptrs implemented by platform
+bool v4l2codecOperationInit(struct V4l2CodecOps *OpFuncs);
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
+
+#endif // v4l2codec_device_ops_h
+
```
chivakker: I would prefer to add a new interface with a few examples on how to use it before just adding a new API that has no usage.

interface/v4l2codec_device_ops.h
```diff
@@ -0,0 +1,83 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *   Created on: 2015-11-26
+ *       Author:aihua.zah@alibaba-inc.com
+ */
```
uartie: libyami is LGPL... this license looks like BSD.  Is it ok to mix licenses like this?

interface/v4l2codec_device_ops.h
uartie: Should this file be in the interfaces directory?  It doesn't follow filename convention with the rest of the interfaces nor does it appear to be an installed header.  It also seems very specific to v4l2.

interface/v4l2codec_device_ops.h
xuguangxin: it's ok LGPL compatible with BSD.

v4l2/v4l2_wrapper.h
```diff
@@ -27,6 +27,7 @@
 #else
 #include <EGL/egl.h>
 #endif
+#include "v4l2codec_device_ops.h"
```
uartie: v4l2codec_device_ops.h is not an installed header, but v4l2_wrapper.h is.

interface/v4l2codec_device_ops.h
halleyzhao: to be honest, this file should be an interface file defined by OS(and installed by the OS's media framework), like the android's OMXPluginBase.h.
so, i don't assume libyami install this header file.
anyway, the header file itself is generic enough, if other OS (like chromOS) would like to pick it up to use; we can consider to install it.

v4l2/v4l2_wrapper.h
halleyzhao: my mistake, removed it in v4l2_wrapper.h.
only include it in v4l2_wrapper.cpp is necessary

halleyzhao:
add build option "--enable-v4l2-ops"
update v4l2decode test to use v4l2codec_device_ops

xuguangxin:
+1

xuguangxin:
merged as 302d09eaa71b580f148eef3c7d10e3b78fb6f185, thanks Halley,U.Artie & Daniel.

Issue #347, opened 2015-11-13
[VP9][decode][Regression] found 13 new failed cases with test suite 3


keqiaox:
1.Testing environment:
libva(master)d3b24b76dfde4daef9d463743971c7522a307092
intel_driver(master)4640d06a8ccb0a12e91bbd00cc3e2423d17ee243
cmrt(master)a4b0f2aefabfdbaf020f617deee40bc07b5992c9
intel_hybrid_driver(master)0375a7287e556fedaf7628a85b143c1233a9ef84
ffmpeg(master)9ac61e73d0843ec4b83f4e3d47eded73234e406e
libyami(master)8fe2149fa5ddba8a47fb43c0f371dc7681186e9c

2.Test command:
./yamidecode -i /media/test_suits3/$filename -m -2

3.Failed cases:
 1---g2_decoder_case_15028.vp9.webm 
 2---g2_decoder_case_15029.vp9.webm 
 3---Stress_VP9_3840x2160_250_extra_stress_1.1.vp9.webm
 4---Stress_VP9_432x240_210_extra_stress_1.0.vp9.webm 
 5---Stress_VP9_432x240_250_extra_stress_1.1.vp9.webm 
 6---Syntax_VP9_1920x1080_205_extra_onfly_scaling_1.1.vp9.webm
 7---Syntax_VP9_432x240_205_extra_onfly_scaling_1.1.vp9.webm 
 8---vp9_sf_hybrid_tile_container_1016x288_1032x296_1112x400.vp9.webM 
 9---vp9_sf_hybrid_tile_CrowdRun_72x80_260x280_782x896.vp9.webM
 10---vp9_sf_hybrid_tile_CrowdRun_88x88_136x128_340x420.vp9.webM 
 11---vp9_sf_hybrid_tile_riverbed_64x64_128x128_1024x1024.vp9.webM
 12---vp9_sf_hybrid_tile_seg_CrowdRun_64_512_128_1024.vp9.webM 
 13---vp9_sf_hybrid_tile_seg_riverbed_32x32_512x512.vp9.webM  

4.error message
 1.  1-4,7,12,13:(Segmentation fault (core dumped)
 2.   5,6,8-11:MD5 mismatch with the reference vaule

lizhong1008:
Hi Keqiao:
     Are you sure all these bitstreams are regression? As I know, at least these videos can't decoded in last release vesion(And it is driver issue).
Stress_VP9_3840x2160_250_extra_stress_1.1.vp9
[...]

PR #372, opened 2015-12-08
vp9dec: fix regression failure, issue #347

clearlylin:
The issue results from the update of avformat.

### DIFF
--- a/tests/decodeinputavformat.cpp
+++ b/tests/decodeinputavformat.cpp
@@ -38,13 +38,24 @@ DecodeInputAvFormat::DecodeInputAvFormat()
 
 bool DecodeInputAvFormat::initInput(const char* fileName)
 {
+    uint16_t width = 0, height = 0;
     int ret = avformat_open_input(&m_format, fileName, NULL, NULL);
     if (ret)
         goto error;
+    int i;
+    for (i = 0; i < m_format->nb_streams; ++i) {
+        AVCodecContext* ctx = m_format->streams[i]->codec;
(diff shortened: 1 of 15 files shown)


tests/decodeinputavformat.cpp
```diff
@@ -53,7 +59,11 @@ bool DecodeInputAvFormat::initInput(const char* fileName)
             if (ctx->extradata && ctx->extradata_size)
                 m_codecData.append((char*)ctx->extradata, ctx->extradata_size);
             m_videoId = i;
-            setResolution(ctx->coded_width, ctx->coded_height);
+            if (ctx->coded_width > width || ctx->height > height) {
```
lizhong1008: using "||" may set a larger width and a smaller height. Would better judge height and width independent.

tests/decodeinputavformat.cpp
```diff
@@ -38,9 +38,15 @@ DecodeInputAvFormat::DecodeInputAvFormat()
 
 bool DecodeInputAvFormat::initInput(const char* fileName)
 {
+    uint16_t width = 0, height = 0;
     int ret = avformat_open_input(&m_format, fileName, NULL, NULL);
     if (ret)
         goto error;
+    if (m_format->nb_streams) {
+        AVCodecContext* ctx = m_format->streams[0]->codec;
```
xuguangxin: why we need this? streams[0] maybe a audio track

xuguangxin:
please give some detailed description in your commit log. why we need this fix?

clearlylin:
edited it

PR #373, opened 2015-12-09
[v4l2] capture device node starts as late as possible

halleyzhao:
- in order to adopt all the buffer provided by client before decoding,
  we start capture device node after all output buffer has been enqued.
  then v4l2codec can setup hw related context(VASurface/VAContext etc)
  upon this STREAMON ioctl
- decode input (OUTPUT node) starts as early as possible to parser codec_data
  but frame decoding waits until the above is done.
- add buffer memory type for dma_buf and android native buffer


xuguangxin:
+1

xuguangxin:
merged as d654f94125eaa71126a2b310d5bd6415cc1ec821. 
thanks halley

PR #381, opened 2015-12-11
encoder/hevc: fix format specifier for size_t in DEBUG

### LINKED ISSUE #380: Compile error with --enable-debug --enable-h265enc
```
vaapiencoder_hevc.cpp: In member function 'virtual Encode_Status YamiMediaCodec::VaapiEncoderHEVC::doEncode(const SurfacePtr&, uint64_t, bool)':
vaapiencoder_hevc.cpp:1031:282: error: format '%d' expects argument of type 'int', but argument 6 has type 'std::list<std::tr1::shared_ptr<YamiMediaCodec::VaapiEncPictureHEVC> >::size_type {aka long unsigned int}' [-Werror=format=]
DEBUG("m_reorderFrameList size: %d\n", m_reorderFrameList.size());
^
vaapiencoder_hevc.cpp: In member function 'void YamiMediaCodec::VaapiEncoderHEVC::shortRfsUpdate(const PicturePtr&)':
vaapiencoder_hevc.cpp:1185:272: error: format '%d' expects argument of type 'int', but argument 6 has type 'std::deque<std::tr1::shared_ptr<YamiMediaCodec::VaapiEncoderHEVCRef> >::size_type {aka long unsigned int}' [-Werror=format=]
DEBUG("m_refList1_size is %d\n", m_refList1.size());
^
cc1plus: all warnings being treated as errors

```

uartie:
With --enable-debug and --enable-h265enc, the compilation
fails because of an incorrect format specifier for the size_t
type used in the DEBUG macro.  The correct GNU portable
specifier for size_t is %zu.

Fixes #380

### DIFF
--- a/encoder/vaapiencoder_hevc.cpp
+++ b/encoder/vaapiencoder_hevc.cpp
@@ -1028,7 +1028,7 @@ Encode_Status VaapiEncoderHEVC::doEncode(const SurfacePtr& surface, uint64_t tim
         CodedBufferPtr codedBuffer = VaapiCodedBuffer::create(m_context, m_maxCodedbufSize);
         if (!codedBuffer)
             return ENCODE_NO_MEMORY;
-        DEBUG("m_reorderFrameList size: %d\n", m_reorderFrameList.size());
+        DEBUG("m_reorderFrameList size: %zu\n", m_reorderFrameList.size());
         PicturePtr picture = m_reorderFrameList.front();
         m_reorderFrameList.pop_front();
         picture->m_codedBuffer = codedBuffer;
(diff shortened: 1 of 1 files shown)


xuguangxin:
Hi @uartie 
In long term, yami maybe support windows. %zu is only portable in GNU compiler.

Do you think it's ok to change it to:
DEBUG("m_reorderFrameList size: %d\n", (int)m_reorderFrameList.size());
it's more portable than %zu.

seanvk:
I disagree, this is a fundamental issue in build configuration.  Let's correct.  OS support is a policy decision that has a wide range of implications. Let's table that for further discussion.
+1

xuguangxin:
ok, we can consider this later. +1

xuguangxin:
merged as 971155f5df31937b92761e6275c2ed978e48ced4, thanks U.Artie and Sean.

PR #393, opened 2015-12-16
format: remove forced formatting in pre-commit

### LINKED ISSUE #390: The pre-commit hook does not follow coding-style
https://github.com/01org/libyami/wiki/Coding-Style states that libyami uses the Webkit coding style.  However, the `pre-commit` hook does a very poor job at formatting the code according to the Webkit coding-style guidelines.  We should either fix the pre-commit hook to follow the coding style or stop asking users to use the hook and remove it from the source tree.

xuguangxin:
kr & gst for formatter is bad to to c++ file. so we will not force do it
It fixes #390


uartie:
lgtm :+1: 

We should also remove any mention of it at https://github.com/01org/libyami/wiki/Coding-Style

xuguangxin:
merge and removed.

PR #396, opened 2015-12-17
reduce the number of dynamic library

clearlylin:
merge libyami_codecparser.so, libyami_encoder.so, libyami_common.so,
libyami_decoder.so, libyami_vaapi.so, libyami_vpp.so and libyami_capi.so
to single dynamic library,libyami.so.


clearlylin:
Now , the include headers exists problem.
merging @344 pull request will be OK.

clearlylin:
team is editing the codecparser now.

xuguangxin:
Hi Halley:
It should no problem, since output license is LGPL too.What is your concern?

configure.ac
```diff
@@ -415,6 +415,7 @@ AC_OUTPUT([
          pkgconfig/libyami_encoder.pc
          pkgconfig/libyami_vaapi.pc
          pkgconfig/libyami_vpp.pc
+         pkgconfig/libyami.pc
```
xuguangxin: do we need remove other pc if we just export single so?

clearlylin: yes. 
will remove other pc after confirm the modification.

Makefile.am
```diff
@@ -14,16 +14,49 @@ if ENABLE_DOCS
...
+	$(NULL)
+
+libyami_la_LIBADD = common/libyami_common.la \
+	codecparsers/libyami_codecparser.la \
+	vaapi/libyami_vaapi.la \
+	decoder/libyami_decoder.la \
+	encoder/libyami_encoder.la \
+	vpp/libyami_vpp.la \
+	$(NULL)
+
```
uartie: Do we want libyami_v4l2.la, too?

Makefile.am
```diff
@@ -14,16 +14,49 @@ if ENABLE_DOCS
...
+endif
+
 if ENABLE_TESTS
 SUBDIRS += tests
 SUBDIRS += testscripts
 SUBDIRS += examples
 endif
+
+lib_LTLIBRARIES = libyami.la
+libyami_la_SOURCES =
```
uartie: Please add `nodist_EXTRA_libyami_la_SOURCES = dummy.cpp` to force C++ linking...

https://www.gnu.org/software/automake/manual/html_node/Libtool-Convenience-Libraries.html

examples/Makefile.am
```diff
@@ -49,7 +48,6 @@ endif
 
 YAMI_DECODE_LIBS = \
 	$(YAMI_COMMON_LIBS) \
-	$(top_builddir)/decoder/libyami_decoder.la \
 	$(NULL)
 
 if BUILD_STATIC
```
uartie: All of the whole-archive libraries specified by LDFLAGS inside the BUILD_STATIC conditionals should be changed to `$(top_builddir)/.libs/libyami.a` now.  I believe BUILD_STATIC is used in examples/Makefile.am and tests/Makefile.am.

uartie: Please note, however, that for each bin_PROGRAM you'll want the `-Wl,--whole-archive $(top_builddir)/.libs/libyami.la -Wl,--no-whole-archive` to show up only once per link step, otherwise it'll fail to compile.  That is, watch out for some of the LDFLAGS that build upon others.

Makefile.am
uartie: ...or does that have to stay separate?

clearlylin: libyami_v4l2.so will be single one.

Makefile.am
clearlylin: OK

clearlylin:
There is a problem with linking static library.and the problem is similar to  initialization of static variable factory.h，such as 
const bool VaapiDecoderH265::s_registered =
VaapiDecoderFactory::register_<VaapiDecoderH265>(YAMI_MIME_H265);

The initialization of s_registered is not called,.

configure.ac
```diff
@@ -419,13 +419,8 @@ AC_CONFIG_FILES([Makefile
...
 
 AC_OUTPUT([
-         pkgconfig/libyami_codecparser.pc
-         pkgconfig/libyami_common.pc
-         pkgconfig/libyami_decoder.pc
-         pkgconfig/libyami_encoder.pc
-         pkgconfig/libyami_vaapi.pc
-         pkgconfig/libyami_vpp.pc
-         pkgconfig/libyami_capi.pc
+		 pkgconfig/libyami.pc
```
xuguangxin: indent

xuguangxin:
Hi @uartie :  I am not familiar with build system. Do yo think this is ok for commit?
thanks.

examples/Makefile.am
```diff
@@ -44,23 +44,25 @@ VPP_INPUT_SOURCES = \
...
-	$(top_builddir)/vaapi/libyami_vaapi.la \
+	$(top_builddir)/libyami.la \
 	$(NULL)
 
 if ENABLE_X11
 YAMI_COMMON_LIBS += $(LIBVA_X11_LIBS) -lX11
 endif
 
+if BUILD_STATIC
+YAMI_COMMON_LIBS   += $(LIBVA_DRM_LIBS) $(LIBVA_X11_LIBS) -lX11
```
uartie: $(LIBVA_DRM_LIBS) should go into YAMI_COMMON_LIBS unconditionally.  That is, it does not depend on whether we BUILD_STATIC or not.  Also, $(LIBVA_X11_LIBS) and -lX11 are already captured in the ENABLE_X11 conditional appropriately.  The only thing that should go in a BUILD_STATIC conditional is the LDFLAGS for our internal libs (i.e. libyami.a) to force the linker to capture them as a "whole archive".  That is, it prevents the Ubuntu version of GCC linker from stripping out our static initializers for the factories and such.

tests/Makefile.am
```diff
@@ -29,24 +29,28 @@ AM_CPPFLAGS = $(AM_CFLAGS)
...
-	$(top_builddir)/vaapi/libyami_vaapi.la \
+	$(top_builddir)/.libs/libyami.la \
 	$(NULL)
 
 if ENABLE_X11
 YAMI_COMMON_LIBS += $(LIBVA_X11_LIBS) -lX11
 endif
 
+if BUILD_STATIC
+YAMI_COMMON_LIBS   += $(LIBVA_DRM_LIBS) $(LIBVA_X11_LIBS) -lX11
```
uartie: Remove this line... see comment above.

tests/Makefile.am
```diff
@@ -180,13 +168,5 @@ yamitranscode_SOURCES  = vppinputdecode.cpp vppinputoutput.cpp vppoutputencode.c
...
-	$(top_builddir)/encoder/libyami_encoder.la \
-	$(top_builddir)/vpp/libyami_vpp.la \
-	$(NULL)
-yamiinfo_LDFLAGS = -Wl,--no-as-needed \
-	$(YAMI_DECODE_LDFLAGS) \
-	$(YAMI_ENCODE_LDFLAGS) \
-	$(YAMI_VPP_LDFLAGS) \
-	$(NULL)
+yamiinfo_LDADD = $(YAMI_COMMON_LIBS)
+yamiinfo_LDFLAGS = $(YAMI_DECODE_LDFLAGS)
```
uartie: We still might need `-Wl,--no-as-needed` for Ubuntu's linker.  I will try this later when I'm near my Ubuntu system.

uartie: Ok, I just checked on Ubuntu and we still need `-Wl,--no-as-needed` for yamiinfo, unconditionally.

uartie:
Also, please shorten the title summary on the first commit message to no more than 50 chars.  Most of it can go into the description body.  Here's a good article on commit messages: http://chris.beams.io/posts/git-commit/#seven-rules

clearlylin:
@uartie  
have changed it according to your advise.

uartie:
lgtm :+1:

xuguangxin:
merged as 0528aacda8df660974a1898ef30b38f64c5565a7

xuguangxin:
thanks U.Arite

Issue #409, opened 2016-01-05
decoder should support external allocated frames


xuguangxin:
It's core part of zero copy. With this we can allocate NV12 surface directly from display, decode and display it.

Issue #407, opened 2016-01-05
encoder should pass all reconstuct frames to vaCreateContext


xuguangxin:
it's required for va api.

PR #411, opened 2016-01-05
efforts to enable encoder on android and support external allocator for decoder

xuguangxin:


xuguangxin:
this fix #407, fix #408, fix #409 , fix #410

xuguangxin:
@lizhong1008 @YuJiankang please help to review.thanks

lizhong1008:
Has this pull request passed regression test?

xuguangxin:
thanks for reminder. Just h264 tested. I will test other formats. and give you feed back.

common/surfacepool.cpp
```diff
@@ -0,0 +1,87 @@
...
+    m_params.height = height;
+    m_params.size = size;
+    YamiStatus status = alloc->alloc(alloc.get(), &m_params);
+    if (status != YAMI_SUCCESS)
+        return status;
+
+    //prepare surfaces for pool
+    std::deque<SurfacePtr> surfaces;
+    for (uint32_t i = 0; i < m_params.size; i++) {
+        SurfacePtr s(new VaapiSurface(m_params.surfaces[i], width, height));
```
lizhong1008: Can we init m_display of VaapiSurface here? It may more convenient in some case, surface mapping for example.

xuguangxin: We are move forwards to keep yam's core api more simpler. We accept bit stream and output Surfaces. It no need map VaImage. We will remove all VaImage map in libyami.Let user map and retrieve VaImages.

vaapi/vaapisurfaceallocator.h
```diff
@@ -0,0 +1,49 @@
...
+#define vaapisurfaceallocator_h
+#include "common/basesurfaceallocator.h"
+#include <va/va.h>
+
+namespace YamiMediaCodec{
+
+class VaapiSurfaceAllocator : public BaseSurfaceAllocator
+{
+    //extra buffer size for performance
+    static const uint32_t EXTRA_BUFFER_SIZE = 5;
```
lizhong1008: 5 extra buffers seems a little larger for 4k or 8k videos. Can we reduce it to 2 or 3, or change it depends on resolution? Every codec can set a larger extar_buffer_size by itself?

xuguangxin: In real world case user should set external allocator. Every codec should set the buffer strictly follow the spec. And the external allocator will decide how many buffer should extrated. Suppose decoder need 3 frame for dpb, and the queue   between decoder and sink have size 3, and display will hold 1 frame for on screen display.  it should should return 3+3+1 frame in alloctor. 
We use 5 extra buffers here just for some one do not care about the size. Give them a better performance and acceptable memory overhead.

lizhong1008: I agree we should set cache buffers. But in fact, we set it at two places, one is here, the other is in every code. Take h264 for example, DPBSize + H264_EXTRA_SURFACE_NUMBER. At least one is redundant.

common/surfacepool.cpp
lizhong1008: My suggestion to init m_display is not only for encoder surfaces, also for decoder. Except map surface, almost all surface operation for libva driver need the VaDisplayID, including sync/map/query and so on. There is litter effort to init m_display but more flexible for future usage. 
On the other hand, I don't think we can make sure won't map surface forever.

lizhong1008:
I'm fine for this series patches overall.

xuguangxin:
update this patch. so we can pass all regresssion.

jpeg regression failed. so I have to add display to surface....

lizhong1008:
+1. Let's remove XXX_EXTRA_SURFACE_NUMBER of each codec next time.

xuguangxin:
merged as 89f9456df935f920c838bcd7e79a63ffa4180aec

PR #412, opened 2016-01-07
debug: add thread id to log

halleyzhao:
- add tid in log
- add thread name for self created threads
- include common.mk for vpp

### DIFF
--- a/common/log.h
+++ b/common/log.h
@@ -38,6 +38,10 @@
 #include <stdint.h>
 #include <assert.h>
 #include <string.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#define GETTID()    syscall(__NR_gettid)
+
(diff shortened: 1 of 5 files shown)


common/log.h
```diff
@@ -38,6 +38,10 @@
 #include <stdint.h>
 #include <assert.h>
 #include <string.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#define GETTID()    syscall(__NR_gettid)
```
xuguangxin: will this cause problem for pure linux?

halleyzhao: it has been tested on ubuntu.

xuguangxin:
ok, +1, thanks

xuguangxin:
merged as 1ab59bcd12a3db458fe8b25852e9d2258bfbd4bf

PR #417, opened 2016-01-08
enc: add multi-reference-frame support.

lizhong1008:
1. Add multi-reference frames interface and test cases
2. Add multi-reference frames support for h264 and hevc.
   Because of current reoder mechanism, multi-reference is only used
   for forward reference. The max value of back referece frames is still
   one.


xuguangxin:
+1

xuguangxin:
merged as 72470d27ac93df1cf7df049bd34ee80450bc8ab4

PR #420, opened 2016-01-13
tests: remove needless statements

andyLiuChina:
decodehelp.h: return false instead -1 when parse failed.
v4l2decode: remove needless statements which has been done in the function process_cmdline()


xuguangxin:
+1

xuguangxin:
merged as c3dbe6337c6d3cebe3f0a90c02ada917853e3ddd

PR #421, opened 2016-01-13
v4l2: add back texture render mode

halleyzhao:
- add texture render mode of x11 init
- though gcc treat &arrary as array, it's better use array
- fix for IOCTL_COMMAND_STRING_MAP, force to use int for ioctl command

however, video frame doesn't show because drm_intel_bo_flink() fails in driver.

### DIFF
--- a/egl/egl_vaapi_image.cpp
+++ b/egl/egl_vaapi_image.cpp
@@ -78,6 +78,7 @@ EGLImageKHR EglVaapiImage::createEglImage(EGLDisplay eglDisplay, EGLContext eglC
         return m_eglImage;
     if (!acquireBufferHandle(memoryType))
         return EGL_NO_IMAGE_KHR;
+    //FIXME, it doesn't support planar video frame yet
     m_eglImage = createEglImageFromHandle(eglDisplay, eglContext, memoryType, m_bufferInfo.handle,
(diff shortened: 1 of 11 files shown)


xuguangxin:
compile failed if i use this command.revert this patch is ok.
 ./autogen.sh --prefix=$LIBYAMI_PREFIX  --enable-tests   --enable-tests-gles --enable-avformat  --enable-dmabuf --enable-v4l2-glx

halleyzhao:
update the first patch to pass build

halleyzhao:
does chromeos still uses glx? if not, we'd better drop glx support, it isn't useful but much complex.

xuguangxin:
-m 3 -m 4 is ok on my HSW.

xuguangxin:
it's not support i think, sure we can remove it.

halleyzhao:
wait a min to merge, some issues met on atoi, may be atoll should be used

xuguangxin:
+1. is it ok to merge this?

halleyzhao:
手工rebase了一下，merge吧。

xuguangxin:
+1

xuguangxin:
merged as 7df2e93d98ab3d0b9fe2bf1725eb491edcb6d37d

xuguangxin:
thanks halley

PR #423, opened 2016-01-14
make: modify comparison between int and uint

wudping:
Delete flag '-Wno-sign-compare' from both CFLAGS and CXXFLAGS,
and work out the compiling error in comparison between int and uint.
In case of some type int variables involved in too many functions,
it is difficult for me to change its type to uint in declaration.
There are still some cast expressions,
such as casting int to uint.
So I appreciate advices to reduce the use of cast expression honestly.

This vession has passed through the regression testing by the testing script.

### DIFF
--- a/encoder/vaapiencoder_hevc.cpp
+++ b/encoder/vaapiencoder_hevc.cpp
@@ -128,7 +128,8 @@ static uint32_t log2 (uint32_t num)
     uint32_t ret = 0;
     assert(num);
 
-    while (num > (1 << ret))
+    uint32_t one = 1;
+    while (num > (one << ret))
         ++ret;
 
     return ret;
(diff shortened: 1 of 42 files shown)


encoder/vaapiencoder_hevc.cpp
```diff
@@ -1268,7 +1269,11 @@ void VaapiEncoderHEVC::shortRfsUpdate(const PicturePtr& picture)
             m_shortRFS.num_positive_pics      = 1;
             m_shortRFS.delta_poc_s1_minus1[0]  = m_refList1[0]->m_poc - picture->m_poc - 1;
             m_shortRFS.used_by_curr_pic_s1_flag[0]            = 1;
+<<<<<<< HEAD
```
xuguangxin: it's not build able

PR #424, opened 2016-01-15
make: modify comparison between int and uint

wudping:
Delete flag '-Wno-sign-compare' from both CFLAGS and CXXFLAGS,
and work out the compiling error in comparison between int and uint.
In case of some type int variables involved in too many functions,
it is difficult for me to change its type to uint in declaration.
There are still some cast expressions,
such as casting int to uint.
So I appreciate advices to reduce the use of cast expression honestly.

This vession has passed through the regression testing by the testing script.

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -46,7 +46,7 @@ uint32_t guessFourcc(const char* fileName)
     const char* extension = strrchr(fileName, '.');
     if (extension) {
         extension++;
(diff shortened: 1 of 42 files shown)


xuguangxin:
+1

xuguangxin:
merged as 4b78764e10165ca6d48c58be5550baf67fe80d5a

PR #427, opened 2016-01-20
android: enable std::bind

andyLiuChina:
define BIND as std::bind which declare in c++11,
so we should add compiling option -std=c++11.

### DIFF
--- a/interface/VideoCommonDefs.h
+++ b/interface/VideoCommonDefs.h
@@ -28,17 +28,21 @@
 #ifdef __cplusplus
 #ifdef ANDROID
 #include <memory>
+#include <functional>
 #else
 #include <tr1/memory>
+#include <tr1/functional>
 #endif
 namespace YamiMediaCodec{
 #ifdef ANDROID
+    #define Bind std::bind
     #define SharedPtr std::shared_ptr
     #define WeakPtr std::weak_ptr
     #define DynamicPointerCast std::dynamic_pointer_cast
     #define StaticPointerCast std::static_pointer_cast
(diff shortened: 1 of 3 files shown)


xuguangxin:
+1

interface/VideoCommonDefs.h
```diff
@@ -33,12 +33,14 @@
...
 namespace YamiMediaCodec{
 #ifdef ANDROID
+    #define BIND std::bind
     #define SharedPtr std::shared_ptr
     #define WeakPtr std::weak_ptr
     #define DynamicPointerCast std::dynamic_pointer_cast
     #define StaticPointerCast std::static_pointer_cast
     #define EnableSharedFromThis std::enable_shared_from_this
 #else
+    #define BIND std::tr1::bind
```
uartie: Why all uppercase?  It's inconsistent with the rest of the naming.  How about "Bind" instead?

interface/VideoCommonDefs.h
```diff
@@ -33,12 +33,14 @@
 #endif
 namespace YamiMediaCodec{
 #ifdef ANDROID
+    #define BIND std::bind
```
uartie: std::bind is defined in <functional> and should be included.

interface/VideoCommonDefs.h
uartie: std::tr1::bind is defined in <tr1/functional> and should be included

xuguangxin:
+1

xuguangxin:
merged as 643017287b8258fe4566f6828977efc5a9cd2950

PR #429, opened 2016-01-21
remove glx mode since it no use anymore

xuguangxin:
glx mode original for chrome os, but it evolving from glx to ozone. So we'd better remove it.


uartie:
`./autogen.sh --enable-v4l2 --enable-tests`
...
`../v4l2/.libs/libyami_v4l2.so: undefined reference to 'eglGetProcAddress'`

uartie:
There's still a bunch of places in configure.ac that evaluate "$enable_v4l2_glx" that need removed... that might fix the compile error.

xuguangxin:
done, it's so hard...

uartie:
+1 lgtm

tests/v4l2decode.cpp
```diff
@@ -207,10 +201,6 @@ static Display * x11Display = NULL;
 static Window x11Window = 0;
 #endif
 #ifdef ANDROID
```
chivakker: This is an empty #ifdef. consider removing it

xuguangxin:
update according to Daniel's comments.

halleyzhao:
+1
Thanks

xuguangxin:
merged, thanks all.

PR #431, opened 2016-01-28
enc: remove surface validity check

leileishang:
Since in some cases(like android libva) surface id 0 is valid,
and may support windows in future, remove surface validity check
to make yami general. Driver will decide it.

Signed-off-by: Leilei Shang <leilei.shang@intel.com>


xuguangxin:
+1

xuguangxin:
merged as 4ebd37df9f337ea63a3bdb08b3dfca5fb0618bef

PR #435, opened 2016-02-01
create YamiVersion.h for api version track.

### LINKED ISSUE #318: Need a way to get libyami version
libva has a good method for getting the version both at compile time and run time.
va_version.h for compile time and vaInitialize for run time.
Can we add similar version info for libyami?

xuguangxin:
Create YamiVersion.h for api version, we should update it
when the interface between yami and user changed.
this fix #318


xuguangxin:
merged as c7607b12b30bfefdd99efbada6dfe98d96b60baa

PR #440, opened 2016-02-05
tests: Add "ipbmode" and "keyperiod" options to yamitranscoder.

ZhongXiaoxia:
tests: Add "ipbmode" and "keyperiod" options to yamitranscoder.

### DIFF
--- a/tests/yamitranscode.cpp
+++ b/tests/yamitranscode.cpp
@@ -52,6 +52,8 @@ static void print_help(const char* app)
     printf("   -N <number of frames to encode(camera default 50), useful for camera>\n");
     printf("   --qp <initial qp> optional\n");
     printf("   --rcmode <CBR|CQP> optional\n");
+    printf("   --ipbmode <0(I frame only ) | 1 (I and P frames) | 2 (I,P,B frames)> optional\n");
+    printf("   --keyperiod <key frame period(default 30)> optional\n");
     printf("   --refnum <number of referece frames(default 1)> optional\n");
 }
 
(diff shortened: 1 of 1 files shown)


encoder/vaapiencoder_h264.cpp
```diff
@@ -939,7 +939,10 @@ Encode_Status VaapiEncoderH264::reorder(const SurfacePtr& surface, uint64_t time
         m_reorderFrameList.push_front(picture);
         m_curFrameNum++;
         m_reorderState = VAAPI_ENC_REORD_DUMP_FRAMES;
-    }else if (m_frameIndex % (m_numBFrames + 1) != 0) {
+    } else if ((m_frameIndex % (m_numBFrames + 1) != 0)
+            && (m_frameIndex + 1 != intraPeriod())
+            && (m_frameIndex + 1 != keyFramePeriod())) {
+        // If next frame will be I or Idr, set this frame to P instead of B.
```
lizhong1008: Only when next frame is IDR, need to set this frame to be P frame, because only IDR( not I frame ) need to refresh DPB buffer.

lizhong1008: if KeyFramePeriod is 30, but the fourth frame is set to be forceKeyFrame, encoding sequence will be: IDR BB IDR, right? You patch can't handle this case.

ZhongXiaoxia: As I tested, this error also happens if next frame is I frame (non-IDR). So I have to add two conditions to avoid of this issue.

Best Regards,
Xiaoxia Zhong

ZhongXiaoxia:
Here I split the yamitranscoder new options to a separate commit. The H264 IPB mode issue will be submitted later after I fix the forceKeyFrame case.

PR #445, opened 2016-02-18
h264enc: Fix the encoding failure issue for IBBP mode.

ZhongXiaoxia:


encoder/vaapiencoder_h264.cpp
```diff
@@ -930,6 +930,15 @@ Encode_Status VaapiEncoderH264::reorder(const SurfacePtr& surface, uint64_t time
     bool isIdr = (m_frameIndex == 0 ||m_frameIndex >= keyFramePeriod() || forceKeyFrame);
 
     if (isIdr) {
+        // If the last frame before IDR is B frame, set it to P frame.
```
xuguangxin: What if last gop do not have enough frame? Suppose you have IBBBBP pattern, but you only have two frame ...
Similar logical should apply to last gop too.

ZhongXiaoxia: As I verified, it will discard the last few B frames if last GOP has no enough frames. So the core dumped error won’t happen in such case. Maybe driver handles it already.

Best Regards,
Xiaoxia Zhong

lizhong1008: It is discarded by yami(because VAAPI_ENC_REORD_DUMP_FRAMES is not set), not driver.
Maybe I need add flush logic for this case to avoid B frame discarded. But this patch LGTM since it can fix driver crash issue.

xuguangxin:
ok, maybe we can send NULL to VaapiEncoderBase::encode to indicate flush....

xuguangxin:
+1

lizhong1008:
+1

lizhong1008:
merged as ca1a2b75248430d77a9d5bff7c0e92918aa4469b

PR #447, opened 2016-02-26
Add IDR interval setting to H.264 encoder

ZhongXiaoxia:
h264enc: Modify IDR interval and key period parameters setting.
tests: Add "idrinterval" setting for encoder to indicate how many I frames between two IDRs.
tests: Change "keyperiod" option name to "intraperiod" to be consistent with its feature.

### DIFF
--- a/encoder/vaapiencoder_h264.cpp
+++ b/encoder/vaapiencoder_h264.cpp
@@ -44,7 +44,7 @@ using std::list;
 using std::vector;
 
 /* Define the maximum IDR period */
-#define MAX_IDR_PERIOD 512
+#define MAX_IDR_PERIOD 2000
 
 #define LEVEL51_MAX_MBPS 983040
 #define H264_FRAME_FR 172
@@ -700,7 +700,8 @@ VaapiEncoderH264::VaapiEncoderH264():
     m_useDct8x8(false),
     m_reorderState(VAAPI_ENC_REORD_WAIT_FRAMES),
     m_streamFormat(AVC_STREAM_FORMAT_ANNEXB),
-    m_frameIndex(0)
+    m_frameIndex(0),
+    m_keyPeriod(30)
 {
     m_videoParamCommon.profile = VAProfileH264Main;
     m_videoParamCommon.level = 40;
@@ -708,7 +709,7 @@ VaapiEncoderH264::VaapiEncoderH264():
     m_videoParamCommon.rcParams.minQP = 1;
 
     memset(&m_videoParamAVC, 0, sizeof(m_videoParamAVC));
-    m_videoParamAVC.idrInterval = 30;
+    m_videoParamAVC.idrInterval = 0;
 }
 
 VaapiEncoderH264::~VaapiEncoderH264()
@@ -763,11 +764,11 @@ void VaapiEncoderH264::resetParams ()
         m_numBFrames = ipPeriod() - 1;
 
     assert(intraPeriod() > ipPeriod());
+	
+    m_keyPeriod = intraPeriod() * (m_videoParamAVC.idrInterval + 1);
 
-    if (keyFramePeriod() < intraPeriod())
-        keyFramePeriod() = intraPeriod();
-    if (keyFramePeriod() > MAX_IDR_PERIOD)
-        keyFramePeriod() = MAX_IDR_PERIOD;
+    if (m_keyPeriod > MAX_IDR_PERIOD)
+        m_keyPeriod = MAX_IDR_PERIOD;
 
     if (minQP() > initQP() ||
             (rateControlMode()== RATE_CONTROL_CQP && minQP() < initQP()))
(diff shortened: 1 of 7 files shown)


lizhong1008:
@xuguangxin ,any comment for this patch?

encoder/vaapiencoder_h264.cpp
```diff
@@ -927,9 +919,18 @@ Encode_Status VaapiEncoderH264::reorder(const SurfacePtr& surface, uint64_t time
 
     PicturePtr picture(new VaapiEncPictureH264(m_context, surface, timeStamp));
 
-    bool isIdr = (m_frameIndex == 0 ||m_frameIndex >= keyFramePeriod() || forceKeyFrame);
+    bool isIdr = (m_frameIndex == 0 ||m_frameIndex >= m_keyPeriod || forceKeyFrame);
 
     if (isIdr) {
+        // If the last frame before IDR is B frame, set it to P frame.
```
xuguangxin: why still have this. seems it have been merged to master. https://github.com/01org/libyami/blob/master/encoder/vaapiencoder_h264.cpp#L934

encoder/vaapiencoder_hevc.cpp
```diff
@@ -1022,16 +1023,7 @@ Encode_Status VaapiEncoderHEVC::setParameters(VideoParamConfigType type, Yami_PT
     case VideoParamsTypeAVC: {
```
xuguangxin: could we change this to VideoParamsTypeHEVC?

tests/encodehelp.h
```diff
@@ -152,13 +155,16 @@ static bool process_cmdline(int argc, char *argv[])
                     rcMode = string_to_rc_mode(optarg);
                     break;
                 case 3:
-                    ipbMode= atoi(optarg);
+                    ipbMode = atoi(optarg);
```
xuguangxin: This changed by git-format-clang,right? Instead of format your file you can format your patch only. git-format-clang --sytle=webrtc HEAD^. Use this approach we can find who is really change the code by "git blame"

interface/VideoEncoderDefs.h
```diff
@@ -407,12 +406,6 @@ typedef struct VideoConfigBitRate {
     VideoRateControlParams rcParams;
 }VideoConfigBitRate;
 
-typedef struct VideoConfigAVCIntraPeriod {
```
xuguangxin: This is a interface change. We should increase micro api version number https://github.com/01org/libyami/blob/master/configure.ac#L9

encoder/vaapiencoder_h264.cpp
ZhongXiaoxia: Yes, this fix has been merged to master already. But I don’t know why it shows here or how to remove it. Maybe it’s generated automatically by github? Could you advise how to deal with it?

Best Regards,
Xiaoxia Zhong

xuguangxin: maybe you should

git fetch 01org

git reset --hard 01org/master

git cherry-pick your-last-commit

git push to your github branch with force

encoder/vaapiencoder_hevc.cpp
ZhongXiaoxia: Besides this one, there are also other “AVC” structs and variables used in HEVC source code. Shall we change them all? Maybe we need choose a name which can covers both AVC and HEVC. I think we’d better think over it and change it later, instead of changing  in this version in a hurry.

Best Regards,
Xiaoxia Zhong

tests/encodehelp.h
ZhongXiaoxia: Thanks for your comments. Actually I didn’t use git-format-clang. I just noticed this line in irregular format and then changed it incidentally.

Best Regards,
Xiaoxia Zhong

xuguangxin: It's good but may introduce some side effect in this situation. Someone may relay on git blame find real author for this line. So modify it just for style may make them confused. So we'd better keep it untouched. 
The format is not right because it's hard to check very details about code style. We find https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format is very helpful on this. Please use it to format all your patches. 

command line is 
git-format-clang --sytle=webkit HEAD^.
git commit -a --amend

thanks

ZhongXiaoxia: OK, I see. I will recover it and keep it untouched.

Best Regards,
Xiaoxia Zhong

encoder/vaapiencoder_hevc.cpp
lizhong1008: I think it should be ok for hevc use avc structs, because hevc is evolution of avc. But Avc should not reuse hevc structs.

xuguangxin: Suppose you are a user. Who did not read the code. Just want know which configure can apply to HEVC. this name may make him confuse.

lizhong1008: How about providing a VideoParamsHEVC struct, like this:
typedef struct VideoParamsHEVC {
VideoParamsAVC avcParams;
...other hevc parameters.
}VideoParamsHEVC;

xuguangxin: typedef VideoParamsAVC VideoParamsHEVC maybe better. first fields always a size...

xuguangxin:
close this since xiangxia send an new pull request #449

PR #454, opened 2016-03-07
ocl: fix rgba access stride issue in blend kernel

dspmeng:
each work item reads 2x2 rgba pixels, so both x and y access stride of read_image should be 2.


xuguangxin:
+1

xuguangxin:
merged as 684272026f5533300efc02e77ec125b945965c22

PR #456, opened 2016-03-15
ocl: cache cl kernel objects within ocl vpp handlers

dspmeng:
each ocl vpp handler will cache cl kernel objects after creating them, to avoid find from kernel map every frame.


xuguangxin:
great! +1

PR #462, opened 2016-03-24
testscripts: support all encoder unit test

clearlylin:


xuguangxin:
is it possible to update unit_test.sh instead of create new one?

clearlylin:
It can replace the old one.

xuguangxin:
I mean,could we just send a pull request to replace old unit test using this one?

seanvk:
Testscripts have moved to libyami-utils project.

PR #465, opened 2016-04-05
decoder/vaapidecoder_h265: split annexb and is_nalff in h265

popkartyeah:
add the is_nalff mode in vaapidecoder_h265.cpp to support decode the mp4 media file.

Signed-off-by: Yun Zhou <yunx.z.zhou@intel.com>


xuguangxin:
hi  : 
Since we have changed license form LGPL to Apache v2. Could you resubmit patch to apache branch ? thanks.

PR #466, opened 2016-04-06
encoder/vp8: fix small resolution video encoding bug

clearlylin:
When the resolution of video is very small, the size of vp8 header
maybe be larger than the body so that m_maxCodedbufSize is less than
the size of actual output of buffer.

### DIFF
--- a/encoder/vaapiencoder_vp8.cpp
+++ b/encoder/vaapiencoder_vp8.cpp
@@ -78,6 +78,9 @@ void VaapiEncoderVP8::resetParams()
 {
     //5 times compress ratio
     m_maxCodedbufSize = width() * height() * 3 / 2;
+    //when the resolution of video is very small, maybe the size of header is bigger than the body.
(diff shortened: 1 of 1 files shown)


encoder/vaapiencoder_vp8.cpp
```diff
@@ -77,7 +77,7 @@ Encode_Status VaapiEncoderVP8::getMaxOutSize(uint32_t *maxSize)
 void VaapiEncoderVP8::resetParams()
 {
     //5 times compress ratio
-    m_maxCodedbufSize = width() * height() * 3 / 2;
+    m_maxCodedbufSize = width() * height() * 3 / 2 + 0x1000;
```
xuguangxin: could we add some comments for this?

xuguangxin:
hi  : 
Since we have changed license form LGPL to Apache v2. could you resubmit patch to apache branch ? thanks.

PR #467, opened 2016-04-19
enc: fix h264 core dump issue when intra_period larger than 1024

lizhong1008:
Fix h264/hevc core dump issues when intra_period larger than 1024, or intra_period equal to 1.


xuguangxin:
+1

xuguangxin:
hi  : 
Since we have changed license form LGPL to Apache v2. could you resubmit patch to apache branch ? thanks.

lizhong1008:
Need to push to apache branch, and duplicated with https://github.com/01org/libyami/pull/492. So close it.

Issue #382, opened 2015-12-11
CXXFLAGS and CFLAGS should not be specified in configure.ac or Makefile.am


uartie:
Commit 7e2556efc8223b0f0289894364012e47cb6de548 moved these *user* flags out of the autogen.sh script, which is good.  However, although they don't belong in autogen.sh, they don't belong in configure.ac or any of the Makefile.am files, either.

As far as I can tell, it looks like the project is trying to globally enforce various compiler/preprocessor flags (e.g. -Wall -Werror), which is great.  However, specifying that policy within the *user* flags is not the correct way to do it.  The problem is that if I, as the user, want to specify some additional CXXFLAGS to configure, then the ones in configure.ac get overridden by my local CXXFLAGS setting.  I don't think this is the intention that the project is going for.  Instead, we should be enforcing these flags via the AM_ variants.

See https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html for more details on the correct way to do this.

seanvk:
We need to remove this and use AM_flags in Makefile.am.  
+1

uartie:
As a note, preprocessor/compiler flags specified in AM_C[XX]FLAGS can be negated by their user variable counterparts.  For example, AM_C[XX]FLAGS="-Wall -Werror" can be negated by the user with C[XX]FLAGS="-Wno-all -Wno-error".  I think it's better since it's more explicit and it won't silently override other project-configured compiler/preprocessor flags.  Also consider that if I want to debug the library, I can just set C[XX]FLAGS="-O0" in my environment (or command line) and would still get the project-wide C[XX]FLAGS if they are correctly configured within the AM_C[XX]FLAGS.

C[XX]FLAGS meaning either CFLAGS or CXXFLAGS, of course.

PR #472, opened 2016-06-08
Apache branch: fixes to build system

chivakker:
These patches remove the warnings when running ./autogen.sh && make && make install && make uninstall && make distclean

### DIFF
--- a/capi/Makefile.am
+++ b/capi/Makefile.am
@@ -1,7 +1,3 @@
-INCLUDES = -I$(top_srcdir) \
-	-I$(top_srcdir)/interface \
-	$(NULL)
-
 libyami_capi_source_c = \
 	VideoDecoderCapi.cpp \
 	VideoEncoderCapi.cpp \
@@ -32,31 +28,7 @@ libyami_capi_la_LDFLAGS     = $(libyami_capi_ldflags)
 libyami_capi_la_CPPFLAGS    = $(libyami_capi_cppflags)
 
 if ENABLE_UNITTESTS
-noinst_PROGRAMS = unittest
-
-unittest_SOURCES = \
-	$(top_builddir)/common/unittest_main.cpp \
-	$(NULL)
-
-unittest_LDFLAGS = \
-	$(GTEST_LDFLAGS) \
-	$(NULL)
-
-unittest_LDADD = \
-	libyami_capi.la \
-	$(GTEST_LIBS) \
-	$(NULL)
-
-unittest_CPPFLAGS = \
-	$(GTEST_CPPFLAGS) \
-	$(NULL)
-
-unittest_CXXFLAGS = \
-	$(GTEST_CXXFLAGS) \
-	$(NULL)
-
-check-local: unittest
-	$(builddir)/unittest
+-include Makefile.unittest
 endif
 
 DISTCLEANFILES = \
--- a/codecparsers/Makefile.am
+++ b/codecparsers/Makefile.am
@@ -1,5 +1,3 @@
-INCLUDES = -I$(top_srcdir)
-
 libyami_codecparser_source_c = \
 	bitReader.cpp \
 	bitWriter.cpp \
@@ -82,7 +80,7 @@ libyami_codecparser_source_h_priv += \
 	$(NULL)
 endif
 
-libyami_codecparser_source_h_priv = \
+libyami_codecparser_source_h_priv += \
 	dboolhuff.h \
 	$(NULL)
 EXTRA_DIST = dboolhuff.LICENSE dboolhuff.PATENTS dboolhuff.AUTHORS
@@ -125,70 +123,7 @@ libyami_codecparser_la_LDFLAGS     = $(libyami_codecparser_ldflags)
 libyami_codecparser_la_CPPFLAGS    = $(libyami_codecparser_cppflags)
 
 if ENABLE_UNITTESTS
-noinst_PROGRAMS = unittest
-
-unittest_SOURCES = \
-	$(top_builddir)/common/unittest_main.cpp \
-	bitReader_unittest.cpp \
-	bitWriter_unittest.cpp \
-	$(NULL)
-
-if BUILD_VP8_DECODER
-unittest_SOURCES += \
-        vp8_bool_decoder_unittest.cpp \
-        $(NULL)
-endif
-
-if BUILD_JPEG_PARSER
-unittest_SOURCES += \
-        jpegParser_unittest.cpp \
-        $(NULL)
-endif
-
-if BUILD_H264_DECODER
-unittest_SOURCES += \
-        h264Parser_unittest.cpp \
-        $(NULL)
-endif
-
-if BUILD_H265_DECODER
-unittest_SOURCES += \
-        h265Parser_unittest.cpp \
-        $(NULL)
-endif
-
-if BUILD_MPEG2_DECODER
-unittest_SOURCES += \
-        mpeg2_parser_unittest.cpp \
-        $(NULL)
-endif
-
-if BUILD_VC1_DECODER
-unittest_SOURCES += \
-        vc1Parser_unittest.cpp \
-        $(NULL)
-endif
-
-unittest_LDFLAGS = \
-	$(GTEST_LDFLAGS) \
-	$(NULL)
-
-unittest_LDADD = \
-	libyami_codecparser.la \
-	$(top_builddir)/common/libyami_common.la \
-	$(GTEST_LIBS) \
-	$(NULL)
-
-unittest_CPPFLAGS = \
-	$(GTEST_CPPFLAGS) \
-	$(NULL)
-
-unittest_CXXFLAGS = \
-	$(GTEST_CXXFLAGS) \
-	$(NULL)
-
-check-local: unittest
-	$(builddir)/unittest
+include Makefile.unittest
 endif
 
 DISTCLEANFILES = \
--- a/codecparsers/bitReader_unittest.cpp
+++ b/codecparsers/bitReader_unittest.cpp
@@ -48,12 +48,12 @@ BITREADER_TEST(NoOverflow)
     std::vector<uint8_t>& data = largeData();
     const uint32_t nBytes = data.size();
 
-    ASSERT_EQ(0, nBytes % (1024 * 1024 * 512));
+    ASSERT_EQ(0u, nBytes % (1024 * 1024 * 512));
 
     BitReader reader(&data[0], nBytes);
 
     EXPECT_FALSE(reader.end());
-    EXPECT_EQ(0, reader.getPos());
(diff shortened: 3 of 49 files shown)


uartie:
FYI, I was able to reproduce the signed/unsigned errors on GCC 4.9.2, too.

Today, I realized my build environment had user defined CFLAGS/CXXFLAGS which caused the compiler flags in `configure.ac` to get overshadowed on my builds (see bug #382).  In turn, I've been compiling without `-Werror -Wall` for who knows how long :-/.  I apologize for the false information about it likely being GCC 5.3.x that increased strictness (although it does seem to catch more of these).  You may want to remove the GCC version blurb from the commit message.  And I guess I should go fix #382 sooner than later ;-)  Cheers.

seanvk:
So let's make it a point to focus on compatibility at this time only
with Ubuntu LTS.  Let's be consistent on the compilers we use and test
with.

While it is fine to experiement across a wide range, I feel that it is
getting out of hand.

Thanks,

Sean

chivakker:
It is not my intention to change focus on compatibility or modify a regular building environment with autotools.  With this clear and Ubuntu being the target OS, this patch is fixing a problem Ubuntu 16.04 faces today.  As mentioned in the commit message, the version is expressed to make people aware of which version was used to test and that no other older version will regress due to this patch.

seanvk:
Sounds good.  Let's have PRC verify when back in.
lgtm +1

uartie:
I'm getting the following compiler error:

```
vaapiencoder_hevc.cpp:21:23: fatal error: bitWriter.h: No such file or directory
#include "bitWriter.h"
^
compilation terminated.
vaapiencoder_jpeg.cpp:23:26: fatal error: scopedlogger.h: No such file or directory
#include "scopedlogger.h"
^
compilation terminated.
```

capi/Makefile.am
```diff
@@ -1,4 +1,4 @@
-INCLUDES = -I$(top_srcdir) \
+AM_CPPFLAGS = -I$(top_srcdir) \
```
uartie: You can remove AM_CPPFLAGS definition since  `-I$(top_srcdir)` should be automatic.  And `-I($top_srcdir)/interface` is unnecessary since you fixed the include paths in the source.

codecparsers/Makefile.am
```diff
@@ -1,4 +1,4 @@
-INCLUDES = -I$(top_srcdir)
+AM_CPPFLAGS = -I$(top_srcdir)
```
uartie: You can remove AM_CPPFLAGS since `-I$(top_srcdir)` should be automatic.

common/Makefile.am
```diff
@@ -1,4 +1,4 @@
-INCLUDES = -I$(top_srcdir)
+AM_CPPFLAGS = -I$(top_srcdir)
```
uartie: You can remove AM_CPPFLAGS since `-I$(top_srcdir)` should be automatic.

vaapi/Makefile.am
```diff
@@ -1,4 +1,4 @@
-INCLUDES = -I$(top_srcdir)
+AM_CPPFLAGS = -I$(top_srcdir)
```
uartie: You can remove AM_CPPFLAGS since `-I$(top_srcdir)` should be automatic.

capi/Makefile.am
uartie: https://www.gnu.org/software/automake/manual/html_node/Program-Variables.html

i.e. "Automake already provides some -I options automatically, in a separate variable that is also passed to every compilation that invokes the C preprocessor. In particular it generates ‘-I.’, ‘-I$(srcdir)’, and a -I pointing to the directory holding config.h (if you’ve used AC_CONFIG_HEADERS). "

We use AC_CONFIG_HEADERS, and config.h is in $(top_srcdir), thus we satisfy the last rule.

chivakker: Yup I'll remove it.

codecparsers/Makefile.am
chivakker: yes, to be removed

common/Makefile.am
chivakker: yes, to be removed.

vaapi/Makefile.am
chivakker: to be removed

chivakker:
I'll also fix the compilation issues, thanks for helping to properly test this.

chivakker:
So, I fixed the compilation issues, the comments from @uartie  and squashed the fixes into the original patch.  It needs more testing from different ends.  I'd wait for others to comment in as per @seanvk suggestion.

uartie:
lgtm

jayesht:
lgtm, validated this pull request on SKL ubuntu.

codecparsers/bitReader_unittest.cpp
```diff
@@ -48,12 +48,12 @@ BITREADER_TEST(NoOverflow)
     std::vector<uint8_t>& data = largeData();
     const uint32_t nBytes = data.size();
 
-    ASSERT_EQ(0, nBytes % (1024 * 1024 * 512));
+    ASSERT_EQ(0u, nBytes % (1024 * 1024 * 512));
```
xuguangxin, line 51: It's very incontinence to add 'u' prefix for all  unsigned value. And it's hard to maintain.It's very likely we have someone do not use gcc 5.3.1 and did not pay attention to the singed unsigned comparison. He will introduce some unit test which break build for gcc 5.3.1.
Is it ok to ignore signed/unsigned compare warning in unit test?

uartie, line 51: I don't think integer-suffixes are inconvenient.  Plus, they advertise the intentions of the test comparison better.  FYI, these warnings are happening on older GCC versions too.  IMO we shouldn't be ignoring any warnings.

jayesht:
Ubuntu 15.10 with 4.2.0-36-generic #41-Ubuntu SMP Mon Apr 18 15:49:10 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux kernel also faces issue with unsigned comparison with GCC.

seanvk:
Jayesh your testing is out of date.  You should be using Ubuntu 16.04.  Please upgrade to latest LTS and re-verify.

seanvk:
-2 until tested

jayesht:
upgrading my system. Will post once verified with latest LTS.

jayesht:
Skylake machine:

$uname -a
4.4.0-24-generic #43-Ubuntu SMP Wed Jun 8 19:27:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux 

Stock GCC version 5.3.1 - verified that the patch fixes the unsigned comparison issue. 

LGTM.

codecparsers/bitReader_unittest.cpp
xuguangxin, line 51: Hi uarite:
I did not see this warning in my machine. Do I need turn on some compile flag? Could we add some compile flag to treat this as error in older gcc? I do not want some one add  a unittest in old gcc without error, then break other one who use new gcc.

uartie, line 51: What version of GCC?  When you compile, do you see `-Wall -Werror` compiler flags in the compiler command line output?  If not, are you're setting C/CXXFLAGS in your environment? Environment settings will override the project settings (see #382 bug for more details on that).  So either add `-Wall -Werror` to your env C/CXXFLAGS or unset them.  You can also try adding `-Wsign-compare`.

Otherwise, older GCC (< 4.8.2 ?) might not be strict or smart enough to catch these with `-Wall -Werror` and/or `-Wsign-compare`.  If we are diligently validating and testing patches on all Ubuntu LTS releases (precise, trusty, and xenial - i.e. the "compatible" platforms) before accepting them (as we ought to), then breaking others shouldn't be too problematic.  If a patch gets in that breaks a newer GCC compile, then they are easy to fix.  These things just come with the territory of tools and software evolution.  I don't think it's critical to hold up this patchset for this concern.

xuguangxin, line 51: here is my gcc version `gcc (Ubuntu 4.8.4-2ubuntu1~14.04) 4.8.4`
and I do see`-Wall` and `-Werror` in Makefile.

I will try add some signed and unsigned compare in my h264 decoder

```
CC = gcc
CCDEPMODE = depmode=gcc3
CFLAGS = -g -O2 -Wno-unused-function  -Wno-cpp -Wall -Werror
CPP = gcc -E
CPPFLAGS =
CXX = g++
CXXCPP = g++ -E
CXXDEPMODE = depmode=gcc3
CXXFLAGS = -g -O2 -Wno-unused-function -Wno-cpp -Wall -Werror
CYGPATH_W = echo
DEFS = -DHAVE_CONFIG_H
DEPDIR = .deps
DLLTOOL = false

```

xuguangxin, line 51: kind of wired, if i apply this patch

```
diff --git a/decoder/vaapidecoder_h264.cpp b/decoder/vaapidecoder_h264.cpp
index 646f500..66c8b70 100644
--- a/decoder/vaapidecoder_h264.cpp
+++ b/decoder/vaapidecoder_h264.cpp
@@ -1501,7 +1501,14 @@ YamiStatus VaapiDecoderH264::decodeCurrent()
     YamiStatus status = YAMI_SUCCESS;
     if (!m_currPic)
         return status;
-
+    int t = -1;
+    unsigned int s = 0;
+    if (t == 0u) {
+        printf("aaaaaa");
+    }
+    if (t == s) {
+        printf("failed");
+    }
     if (!m_currPic->decode()) {
         ERROR("decode %d failed", m_currPic->m_poc);
         // ignore it to let application continue to decode the next frame

```

gcc only report error for 
`if (t == s) {`
not for
`if (t == 0u) {`

```
make[2]: Entering directory `/home/thomas/projects/yami/libyami/decoder'
  CXX      libyami_decoder_la-vaapidecoder_h264.lo
vaapidecoder_h264.cpp: In member function 'YamiStatus YamiMediaCodec::VaapiDecoderH264::decodeCurrent()':
vaapidecoder_h264.cpp:1509:14: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     if (t == s) {
              ^
cc1plus: all warnings being treated as errors
make[2]: *** [libyami_decoder_la-vaapidecoder_h264.lo] Error 1
```

uartie, line 51: `s` is not a constant, but 0u is.  If you change `s` type to `const unsigned int` then the warning goes away.

xuguangxin, line 51: for 

> EXPECT_EQ(0u, reader.getPos());

reader.getPos is not const, right?

xuguangxin:
merged as 5680ce21f5018d545ff82b36ba092a53ccbd36be. Since it will not break old gcc

PR #475, opened 2016-06-08
include config.h in all .c/.cpp units

### LINKED ISSUE #473: decoder/unittest Factory tests crash with memory corruption in debug build
After commit 89ef3dea365930d430d00df457446fcd347d43d9 decoder/unittest Factory tests crash with memory corruption error when built with --enable-debug.

```
[==========] Running 10 tests from 6 test cases.
[----------] Global test environment set-up.
[----------] 1 test from VaapiDecoderVP8Test
[ RUN      ] VaapiDecoderVP8Test.Factory
[       OK ] VaapiDecoderVP8Test.Factory (0 ms)
[----------] 1 test from VaapiDecoderVP8Test (0 ms total)

[----------] 4 tests from VaapiDecoderJPEGTest
[ RUN      ] VaapiDecoderJPEGTest.Factory
*** Error in `./build/libyami/decoder/unittest': double free or corruption (!prev): 0x0000000001da4450 ***
*** Error in `./build/libyami/decoder/unittest': malloc(): memory corruption: 0x0000000001da4580 ***
```

The crash occurs during virtual destruction of the decoders.  It seems `VaapiDecoderBase::renderPictureCount` is part of one compilation step but not another and therefore causes data/memory layout inconsistencies.  This happens when `config.h` (which defines `__ENABLE_DEBUG__`) is included in one place but not included in another before including `vaapidecoder_base.h`.  Therefore, the aforementioned commit did not cause this bug, but rather only exposed it when the class data layout changed.  It's odd this didn't surface before then, but I guess that requires a deeper understanding of the compiler and data layout.

The right thing to do is include `config.h` in _every_ `.cpp` unit before including other headers, especially when those headers depend on a define from `config.h`.  This is also highly recommended in the gnu documentation (https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Configuration-Headers.html)

I am working on a patch now.

uartie:
Always include config.h first in every .c/.cpp file
before including other headers.  This will ensure
consistent declarations when headers depend on
config.h defs.

https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Configuration-Headers.html

Also Fixes #473

Signed-off-by: U. Artie Eoff ullysses.a.eoff@intel.com

### DIFF
--- a/codecparsers/bitReader.cpp
+++ b/codecparsers/bitReader.cpp
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
 #include <assert.h>
 #include "bitReader.h"
 
--- a/codecparsers/bitReader_unittest.cpp
+++ b/codecparsers/bitReader_unittest.cpp
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
 // primary header
 #include "bitReader.h"
 
--- a/codecparsers/bitWriter.cpp
+++ b/codecparsers/bitWriter.cpp
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
 #include "bitWriter.h"
 #include "common/log.h"
 #include <assert.h>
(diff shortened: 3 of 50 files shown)


chivakker:
lgtm +1, verified +1, ./decoder/unittest segmentation fault after jpeg unittest is gone with this patch.

xuguangxin:
Hi @uartie 
Thanks, great discovery. Seems missed one file h265Parser_unittest.cpp. Is it intend to?

uartie:
@xuguangxin, yep I missed that one... update coming.

seanvk:
lgtm.  Needs verified.

xuguangxin:
verfiied and merged as d6cdba84a3e2f95c2d0d4956497e71640c415f27

PR #482, opened 2016-06-14
Use AM_*FLAGS instead of user flags

### LINKED ISSUE #382: CXXFLAGS and CFLAGS should not be specified in configure.ac or Makefile.am

uartie:
This fixes #382

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -50,6 +50,9 @@ libyami_la_LDFLAGS += $(LIBVA_X11_LIBS) $(X11_LIBS)
 libyami_la_CPPFLAGS += $(LIBVA_X11_CFLAGS)
 endif
 
+libyami_la_LDFLAGS += $(AM_LDFLAGS)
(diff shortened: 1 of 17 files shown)


chivakker:
+1 lgtm

PR #483, opened 2016-06-16
readme: add reference for yami-utils

xuguangxin:
Better point out yami-utils in readme.

### DIFF
--- a/README
+++ b/README
@@ -48,6 +48,14 @@ Git repository for work-in-progress changes is available at:
 <https://github.com/01org/libyami>
 
 
+Demos, Examples and Test Applications
+---------------------------------------------------
+The libyami-utils project provides various example, test and demo
+applications that use libyami.  For more details, please refer to
+
+    https://github.com/01org/libyami-utils
+
+
 Simple api demo application
 ---------------------------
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -48,6 +48,13 @@ Git repository for work-in-progress changes is available at:
 <https://github.com/01org/libyami>
 
 
+Test Applications
+-----------------
+To avoid add too many dependency to your project, we split all test applications
+to yami-utils. It include yamidecode/yamiencode and conformance test script.
+You can get it from https://github.com/01org/libyami-utils
+
+
```
uartie, line 58: I don't think we need to explain ourselves for why they are moved to libyami-utils.

How about something like...

```
Demos, Examples and Test Applications
---------------------------------------------------
The libyami-utils project provides various example, test and demo
applications that use libyami.  For more details, please refer to

    https://github.com/01org/libyami-utils
```

...then we can remove the next section too.

xuguangxin, line 58: Good suggestion, your version is better. But next section is important too, it give user preview of decode api. Because yamidecode maybe to complex to first time user.

chivakker:
lgtm :+1:

uartie:
:+1:

PR #486, opened 2016-06-23
encoder/vp8: fix small resolution video encoding bug

clearlylin:
When the resolution of video is very small such as 96x96,
the size of output buffer including the header of vp8 maybe
be larger than m_maxCodedbufSize.

### DIFF
--- a/encoder/vaapiencoder_vp8.cpp
+++ b/encoder/vaapiencoder_vp8.cpp
@@ -75,6 +75,10 @@ void VaapiEncoderVP8::resetParams()
 {
     //5 times compress ratio
     m_maxCodedbufSize = width() * height() * 3 / 2;
+    //when the resolution of video is very small such as 96x96 and buffer plus the header of vp8,
+    //the size of actual output buffer maybe be bigger the m_maxCodedbufSize.
+    //So plusing 4096 avoids the situation.
+    m_maxCodedbufSize += 0x1000;
 }
 
 YamiStatus VaapiEncoderVP8::start()


encoder/vaapiencoder_vp8.cpp
```diff
@@ -75,6 +75,10 @@ void VaapiEncoderVP8::resetParams()
 {
     //5 times compress ratio
     m_maxCodedbufSize = width() * height() * 3 / 2;
+    //when the resolution of video is very small such as 96x96 and buffer plus the header of vp8,
+    //the size of actual output buffer maybe be bigger the m_maxCodedbufSize.
+    //So plusing 4096 avoids the situation.
+    m_maxCodedbufSize += 0x1000;
```
chivakker, line 81: do you know which is the maximum resolution that needs the extra padding?  If such it would be good to add the padding bytes only on those resolutions.  It will maintain the functionality for high resolutions.

clearlylin, line 81: Currently, I use the minimum resolution, 96x96 , to encode vp8, which needs the extra padding.
But 176x144 do not need exptra padding.
 the maximum resolution is 96x96.

xuguangxin, line 81: Hi Daniel:
The m_maxCodedbufSize is just a guessed value. plus 4k should be a problem for large resolution,  Since it's already very big.
We do not know the compression ratio in driver, So we use entire frame size previously, But Linhai found for small resolution the header size maybe very large. So he plus 4k to make sure it's not overflow

mypopy, line 81: I think we need find a way to calc the padding size, not a fixed value.

2016-06-30 9:45 GMT+08:00 Xu Guangxin notifications@github.com:

> In encoder/vaapiencoder_vp8.cpp
> https://github.com/01org/libyami/pull/486#discussion_r69057949:
> 
> > @@ -75,6 +75,10 @@ void VaapiEncoderVP8::resetParams()
> >  {
> >      //5 times compress ratio
> >      m_maxCodedbufSize = width() \* height() \* 3 / 2;
> > -    //when the resolution of video is very small such as 96x96 and buffer plus the header of vp8,
> > -    //the size of actual output buffer maybe be bigger the m_maxCodedbufSize.
> > -    //So plusing 4096 avoids the situation.
> > -    m_maxCodedbufSize += 0x1000;
> 
> Hi Daniel:
> The m_maxCodedbufSize is just a guessed value. plus 4k should be a problem
> for large resolution, Since it's already very big.
> We do not know the compression ratio in driver, So we use entire frame
> size previously, But Linhai found for small resolution the header size
> maybe very large. So he plus 4k to make sure it's not overflow
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> https://github.com/01org/libyami/pull/486/files/d8ace19d20da9eb807147ec9cb73668de1de7868#r69057949,
> or mute the thread
> https://github.com/notifications/unsubscribe/AA0HYUSgnqTEoI0SjrfHObsTp1XekFoZks5qQx-kgaJpZM4I8YpG
> .

## 

# 

Pixelworks
Room 301-303 No. 88,Lane 887 Zuchongzhi Road, Zhangjiang Hi-tech Park,
Shanghai 201203, China
Best Regards,
Jun zhao/赵军
+++++++++++++++++++++++++++++++++++++++

xuguangxin, line 81: seems no way, We do not know driver's compress ratio.Even driver do not know it i think;)

chivakker, line 81: I think this can be then

if (width <=96 || height <=96)
   m_maxCodedbufSize+=4096;

xuguangxin, line 81: why 96 is so special.It is only an experience value. May related to driver changes. Other resolution may have this issue too. like 320x240.

chivakker, line 81: @clearlylin already explained that the maximum resolution needing the padding is 96x96.  Based on that information the if can be included.

xuguangxin, line 81: Yes, But it is only an experience value for him, 48x188 should have same problem, and it may related to the content. If the content is complex, it should have problem for high resolution.

chivakker, line 81: "Yes, But it is only an experience value for him, 48x188 should have same problem, and it may related to the content. If the content is complex, it should have problem for high resolution."

Can @clearlylin experiment with other values or try to find the root cause of the problem?  Is this a driver requirement?  If so, it would be good to fix it at the driver level instead.

xuguangxin, line 81: No it's not driver's issue. The root cause of this problem is, when you do encode, it will encode frame to slice data,then append a header to the slice data. When resolution is small, the header + slice data may excess the size of width_height_3/2. And 4k is enough for any headers.So linhai provide this solution.

lizhong1008, line 81: Hi Daniel:
     Yeah, if the context is very complex and the quantization value (q_index for vp8 codec) is very small, the coded slice data will be very close to the limitation value width() \* height() \* 3 / 2. And the coded bitstream (slice_data + frame headers) will more than width() \* height() \* 3 / 2. 
  Any resolution may encounter this issue. But it more likely happens on small resolution. Because frame header with fixed size independent on resolution, but smaller resolution with smaller slice_data. As Guangxin mentioned, here 0x1000 is presented frame header size, it is large enough.

lizhong1008, line 81: Encoding driver can't estimate the compress ratio, it depends on context complexity，quantization value and entropy coding result.

chivakker:
Ok, then the header is a fixed sized, isn't it?  So, it would be good to use #define to define header size and extra size which I understand can be bigger than w_h_3/2.  I just want to make it more understandable than a magic number.

xuguangxin:
Hi Daniel:
Magic number is:

> Unique values with unexplained meaning or multiple occurrences which could (preferably) be replaced with named constants

But this 0x1000 is only in one place and fully commented. So it's not a magic number. :)

chivakker:
I am not sure that the comment in the code is reflecting the explanation on the github review comments. That's were "unexplained meaning" is happening.  Naming the constants can certainly help.  I leave it up to you.

xuguangxin:
Ok, I will change it. If this make you confuse, It may confuse other person too. But i do not know how to find a good name for it. Do you think it's ok for

```
//when the resolution of video is very small such as 96x96 and buffer plus the header of vp8,
//the size of actual output buffer maybe be bigger the m_maxCodedbufSize.
//So plusing 4096 avoids the situation.
#define  VP8_EXTRA_SIZE_FOR_HEADER 0x1000
```

xuguangxin:
close this since magic number issue addressed by #512

PR #489, opened 2016-06-28
build: make libyami works on android

YuJiankang:
make the libyami works on android. verified with v4l2decode and v4l2encode.
Signed-off-by: jkyu jiankang.yu@intel.com

### DIFF
--- a/encoder/vaapiencoder_h264.cpp
+++ b/encoder/vaapiencoder_h264.cpp
@@ -147,8 +147,15 @@ static uint8_t h264_get_profile_idc(VideoProfile profile)
 {
     uint8_t idc;
     switch (profile) {
+    case PROFILE_H264_BASELINE:
+    case PROFILE_H264_CONSTRAINED_BASELINE:
+        idc = 66;
+        break;
     case PROFILE_H264_MAIN:
         idc =  77;
+        break;
+    case PROFILE_H264_HIGH:
+        idc = 100;
         break;
     default:
         assert(0);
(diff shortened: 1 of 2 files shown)


decoder/vaapidecoder_h264.h
```diff
@@ -47,9 +51,15 @@ class VaapiDecoderH264 : public VaapiDecoderBase {
 
     class DPB {
         typedef VaapiDecoderH264::RefSet RefSet;
+#ifdef ANDROID
```
xuguangxin: could we define this in  common/Functional.h

decoder/vaapidecoder_mpeg2.cpp
```diff
@@ -206,8 +210,13 @@ VaapiDecoderMPEG2::DPB::getReferencePictures(const PicturePtr& current_picture,
 }
 
 VaapiDecoderMPEG2::VaapiDecoderMPEG2()
+#ifdef ANDROID
```
xuguangxin: use common/Functional.h

codecparsers/mpeg2_parser.cpp
```diff
@@ -83,14 +83,22 @@ namespace MPEG2 {
 
     // default matrix for intra blocks
     // section 6.3.7
+#ifdef ANDROID
+    const static std::array<const uint8_t, 64> kDefaultIntraBlockMatrix
```
xuguangxin: better define this to Arrary and move all things to common/Array.h

decoder/Android.mk
```diff
@@ -22,7 +21,10 @@ LOCAL_SRC_FILES += \
         vaapidecoder_vp9.cpp
 
 LOCAL_SRC_FILES += \
-        vaapidecoder_jpeg.cpp
```
xuguangxin, line 25: So original code can support jpeg?

encoder/Android.mk
```diff
@@ -12,9 +12,6 @@ LOCAL_SRC_FILES += \
         vaapiencoder_h264.cpp
 
 LOCAL_SRC_FILES += \
-        vaapiencoder_jpeg.cpp
```
xuguangxin, line 15: why remove jpeg encoder?

seanvk:
@YuJiankang Can you respond to the question of JPEG encoder?  Also this patch is stale.

YuJiankang:
@seanvk , It's only for android, As there is no requirement for android support from customer currently, I  closed this as this patch need some more improvement. For the JPEG encoder, we should support it on android. I will improve the patch and submit it again if we get the final decision of Android related things. thanks a lot.

PR #491, opened 2016-07-01
fix compile error on fedora 17.

xuguangxin:
This fix issue in gcc (GCC) 4.7.0 20120507 (Red Hat 4.7.0-5)
1. First two commits are libyami's problem
2.  Latest one caused by limitation of gcc 4.7, but we had better fix it since tr1::array do not bring too much benefit in this context.
3. Jpeg still breaking because it have too many places need to change.

Hi @uartie :
Do you think it is ok to remove all array usage in jpeg decoder and encoder? I hope we can support more platforms if it’s no major function missed in those platforms.


lizhong1008:
+1

uartie:
@xuguangxin,
The problem you're seeing with tr1::array usage and GCC 4.7 is caused by a bug in GCC (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=25137).  Newer versions of GCC no longer enables the `-Wmissing-braces` with `-Wall`.  Thus, it's probably better to just add `-Wno-missing-braces` to the `AM_CXXFLAGS` in our `configure.ac`.

uartie:
Also see the latest gnu docs (https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) for `-Wall` enabled warnings.  In particular it says `-Wmissing-braces (only for C/ObjC)`.  Thus, explicitly disabling the warning in our configure.ac will only make gcc 4.7 flags behave more like newer gcc versions.

Issue #439, opened 2016-02-05
Don't abort in encoder if fourcc not supported/handled


uartie:
If the encoder doesn't support a particular fourcc in createNewSurface() then we should not be asserting and forcing the client program to terminate.  Instead we should log an error and return an uninitialized SurfacePtr so the encoder can gracefully propagate the error state back to the client.  This will allow the client program to decide how to recover if needed.

https://github.com/01org/libyami/blob/0412d64f1c8468a87eff5c88c45d7fc061ee5e3a/encoder/vaapiencoder_base.cpp#L266

In general, ASSERT should only be used in rare cases... especially when it's a client library.

yizhouwei:
@xuguangxin it seems that the problem has already solved.
I tested with .yv12 files and the program printed unsupportted fourcc.
although it still abortted in the end.

xuguangxin:
seems this fixed

PR #492, opened 2016-07-04
fix some intra_period issues of encoder.

lizhong1008:


encoder/vaapiencoder_h264.cpp
```diff
@@ -743,13 +743,21 @@ void VaapiEncoderH264::resetParams ()
     DEBUG("resetParams, ensureCodedBufferSize");
     ensureCodedBufferSize();
 
+    if (intraPeriod() == 0) {
+        ERROR("intra period must larger than 0");
+        assert(0);
```
xuguangxin: Better return error here. assert should use for bug only. It's not user friendly if we assert on input value on core library

xuguangxin:
Please help me better understand the issue. Where is 1024 come from, did not see any indicate here.

> enc: fix h264 core dump issue when large intra_period
> When intra_period larger than 1024, no reference list for P frame.
> It is due to poc calculated error.
> And HEVC encoding should also be fixed though no core dump found,
> because hevc poc is equal to m_frameIdx, but h264 is equal to
> 2*m_frameIdx.
> 
> Signed-off-by: Zhong Li zhong.li@intel.com
> Change-Id: I2b4625a7021fa26e1f4ee54f661fd39c155b7776

encoder/vaapiencoder_h264.cpp
lizhong1008: Here resetParams is a void function, if we change it to bool, upper-layer functions need to check it. And I think it's normal to assert parameters with ERROR info printed. We can see this way everywhere in  yami project and libva project.

lizhong1008:
Hi Guangxin:
    you can check the fuction h264_get_log2_max_frame_num in vaapiencoder_h264.cpp. The max value of m_maxFrameNum is 1024.

encoder/vaapiencoder_h264.cpp
xuguangxin: Could we silent change this to other value? We can assert on exception. But assert on user input value maybe not so good especially when it in core library. U.Arite have this concern too.https://github.com/01org/libyami/issues/439

xuguangxin:
+1

Hi @wudping :
Please help to verify it.
two things should be make sure.
1. we have this issue when you did not apply this patch.
2. this patch fix the issue, and did not bring side effect to normal stream. 
Please use analysiser to check the encoded file

wudping:
ok, Guangxin

wudping:
verify + 1

xuguangxin:
merged as 0e4e54f1369978452b283e4c11d5ae227443d27f

PR #500, opened 2016-07-08
h264enc: reduce coded buffer size to boost performance

xuguangxin:
Old code falsely set the codeded buffer size, for 1080P it will set to 7M.
This patch will consider profile and level, it will reduce codeded buffer size.
It benefit small memory system a lot.

Signed-off-by: Xu Guangxin guangxin.xu@intel.com
Signed-off-by: Meng Jia jia.meng@intel.com


lizhong1008:
+1

mypopydev:
+1， after apply this patch, it's increase about 8%- 10% fps in transcode case in our test bed.(Total 2G RAM shared by GPU/CPU in the test bed.)

PR #501, opened 2016-07-11
fix mkv file decode error issue

popkartyeah:
fix hevc extra data do not parse properly issue.


decoder/vaapidecoder_h265.cpp
```diff
@@ -1102,7 +1102,7 @@ bool VaapiDecoderH265::decodeAvcRecordData(uint8_t* buf, int32_t bufSize)
         ERROR("invalid record data");
         return false;
     }
-    if (buf[0] != 1) {
+    if (!(buf[0] || buf[1] || buf[2] > 1)) {
```
xuguangxin: why change this, could you give some comments on source file? Is h264 have this problem too?

PR #508, opened 2016-07-18
mpeg2dec: do not change the status reported

### LINKED ISSUE #505: mpeg2 yami decode can't get format change event
When call yami decoder API, we need to get the YAMI_FORMAT_CHANGE status after call decode( ) function, then we can used getFormatInfo( ) to get some format info, but now, mpeg2 decoder can't get this status, it's will break getFormatInfo() function in this case.

chivakker:
Client needs to know about the format change on the first
time they'll try to decode a sequence header.

Fixes #505

### DIFF
--- a/decoder/vaapidecoder_mpeg2.cpp
+++ b/decoder/vaapidecoder_mpeg2.cpp
@@ -289,9 +289,6 @@ YamiStatus VaapiDecoderMPEG2::start(VideoConfigBuffer* buffer)
         m_parser->nextStartCode(m_stream.get(), next_code);
         DEBUG("start Next start_code %x", next_code);
         status = processConfigBuffer();
-        if (status == YAMI_DECODE_FORMAT_CHANGE) {
-            status = YAMI_SUCCESS;
-        }
     }
 
     return status;


decoder/vaapidecoder_mpeg2.cpp
```diff
@@ -289,9 +289,6 @@ YamiStatus VaapiDecoderMPEG2::start(VideoConfigBuffer* buffer)
         m_parser->nextStartCode(m_stream.get(), next_code);
         DEBUG("start Next start_code %x", next_code);
         status = processConfigBuffer();
-        if (status == YAMI_DECODE_FORMAT_CHANGE) {
```
mypopydev, line 292: So after apply this patch, the MPEG2 decoder start() maybe return the DECODE_FORMAT_CHANGE ? But h264 decoder start() just get DECODE_FAIL or YAMI_SUCCESS, I think we need to make the start() have a uniform action no matter mpeg2/vc1/vp8/vp9/h264/h265/mjpeg decoder.

@xuguangxin do you have any plan to clear the start() function?

uartie, line 292: I agree, it should be uniform across decoders.  DECODE_FORMAT_CHANGE shouldn't be returned from start().  The interface documention already eludes to this: https://github.com/01org/libyami/blob/apache/interface/VideoDecoderInterface.h#L34

uartie, line 292: "configure" or "init" probably would have been a better name, IMO, because "start" seems kind of ambiguous or counterintuitive.

xuguangxin, line 292: Hi @mypopydev 
Yes, we should use gtest to enforce this.

xuguangxin, line 292: Hi @uartie 
Actually, we use start to decode codec data. In some usecases, codec data will send in other channel and it will send multiple times. Change start() to decodeHeader() maybe more reasonable.

chivakker:
Before we think on changing the API, let's work with the current one.  start(configBuffer) gets a configBuffer that can have nal units in the buffer member.  That information has to be processed.  I think it is okay to return the FORMAT_CHANGE if needed but it can also hold it till the first time Client calls decode(videoBuffer).

I would prefer to keep the current functionality and have that across the board with the codecs.  Then we work on cleaning up the API.

@mypopydev  mpeg2 start will not return FORMAT_CHANGE unless you call it with a valid nal unit from the bitstream.  FORMAT_CHANGE will be returned by decode if you send the first nal unit on it.

xuguangxin:
Hi @chivakker : 
So we do not need change anything about this, right? @mypopydev Just need getFormatInfo after he call start.

mypopydev:
@xuguangxin @chivakker 
I will call getFormatInfo() after the first decode() success, but I think it's a work round. :)
Tks.

xuguangxin:
User should call getFormatInfo in two situation:
1. after after start() decoder witch codec data, and start() return YAMI_SUCCUESS.
2. decode return a format change.
So, you did not get format  info from first situation?

mypopydev:
In current code, the caller (me :) ):
1. call getFormatInfo() after the first decode() success, but I think it's a work round. :)
2. call getFormatInfo() after decode() return a format change.

chivakker:
@mypopydev  you can call getFormatInfo whenever you want in your two cases. If called after decode(), it is because it returned FORMAT_CHANGE

The patch enables you to do it after calling decode.  The first use case was already covered and it is not a workaround

mypopydev:
@chivakker  Now is working, Thanks. And we don't used this patch. :). 

2 reasons:
1. Now after call start() with  coded buffer, we can call getFormatInfo( ) get format info, now mpeg2 decoder without this patch is working in this condition.
2. Break the start( ) action :)

xuguangxin:
Hi @chivakker 
Do we still need merge this patch?

seanvk:
@chivakker What is the status of this patch?

chivakker:
This can be merged as the functionality might be required by another customer.  I know the original reporter decided to use the other logic.  IMO, it can be merged.

xuguangxin:
Hi @chivakker 
But this patch will return FORMAT_CHANGE in start() right?Like we discussed above and #505, we should make sure start() only return SUCCESS, FAIL and other failed like unsupported.

chivakker:
@xuguangxin, no this patch does not modify start.  I'll submit a separate patch to remove FORMAT_CHANGE from start and we discuss there the implications.

chivakker:
Actually you're right based on the discussion about what start should return. Closing this as no-op

PR #510, opened 2016-07-19
h264dec: only fill referenced dpb pictures for vaapi

uartie:
Only fill the VAAPI picture reference list with pictures
that are currently referenced in the dpb.  Loading
unreferenced dpb pictures in VAAPI is unnecessary.

Signed-off-by: U. Artie Eoff ullysses.a.eoff@intel.com

### DIFF
--- a/decoder/vaapidecoder_h264.cpp
+++ b/decoder/vaapidecoder_h264.cpp
@@ -1287,10 +1287,12 @@ void VaapiDecoderH264::fillReference(VAPictureH264* refs, size_t size)
     size_t i = 0;
     DPB::PictureList::iterator it = m_dpb.m_pictures.begin();
 
-    for (; it != m_dpb.m_pictures.end(); i++, it++) {
-        fillVAPictureH264(&refs[i], *it);
-        DEBUG("poc %d, isShortRef %d, isRef %d", (*it)->m_poc,
-              (*it)->m_shortTermRefFlag, (*it)->m_isReference);
+    for (; it != m_dpb.m_pictures.end(); it++) {
+        if (!isReference(*it))
+            continue;
+        fillVAPictureH264(&refs[i++], *it);
+        DEBUG("id %d, poc %d, isShortRef %d, isRef %d", (*it)->getSurfaceID(),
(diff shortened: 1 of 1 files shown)


decoder/vaapidecoder_h264.cpp
```diff
@@ -1287,9 +1287,11 @@ void VaapiDecoderH264::fillReference(VAPictureH264* refs, size_t size)
     size_t i = 0;
     DPB::PictureList::iterator it = m_dpb.m_pictures.begin();
 
-    for (; it != m_dpb.m_pictures.end(); i++, it++) {
-        fillVAPictureH264(&refs[i], *it);
-        DEBUG("poc %d, isShortRef %d, isRef %d", (*it)->m_poc,
+    for (; it != m_dpb.m_pictures.end(); it++) {
+        if (!isReference(*it))
```
mypopydev, line 1291: LGTM

Other question,  is it have some issue in the other decoder (e,g HEVC) ?

xuguangxin, line 1291: Maybe not, for h265, we did not loop entire dpb. Just fill short and long term reference. https://github.com/01org/libyami/blob/apache/decoder/vaapidecoder_h265.cpp#L572

lizhong1008:
One benefit to pass all dpb pictures is it's very convenient to check dbp list in va_trace log. I agree to don't load unreference pictures to ReferenceFrames[16] since it named ReferenceFrames.  But it should have little overhead.

chivakker:
lgtm :+1:

xuguangxin:
Hi @lizhong1008 
So, do you think it's ok to merge this patch or you have other suggestion.

lizhong1008:
LGTM. Would better don't mention “overhead” in the comment message. The others may think this patch will have obvious benefit to performance, but actually there are litter performance benefit.

seanvk:
Agree, remove “overhead”. 

Lgtm

uartie:
"overhead" removed from message.

lizhong1008:
+1

mypopydev:
+1

xuguangxin:
merged as a09e9cba523e40852785d0c3f1b414c0169c208a

PR #514, opened 2016-07-21
serials patch to fix #495

### LINKED ISSUE #495: infinite loop when decode h264 with error bits
#0  0x00007fffdfc732cb in YamiParser::BitReader::read(unsigned int) (this=this@entry=0x7fffb9c7e7b0, nbits=nbits@entry=1) at bitReader.cpp:99
#1  0x00007fffdfc7d8c0 in YamiParser::NalReader::readUe() (this=this@entry=0x7fffb9c7e7b0) at nalReader.cpp:72
#2  0x00007fffdfc75690 in YamiParser::H264::Parser::parseSps(std::tr1::shared_ptrYamiParser::H264::SPS&, YamiParser::H264::NalUnit const*) (this=this@entry=0x14741d8, sps=..., nalu=nalu@entry=0x7fffb9c7e8d0) at h264Parser.cpp:398
#3  0x00007fffdfc8d2af in YamiMediaCodec::VaapiDecoderH264::decodeSps(YamiParser::H264::NalUnit*) (this=this@entry=0x14740e0, nalu=0x7fffb9c7e8d0) at vaapidecoder_h264.cpp:1203
#4  0x00007fffdfc9411c in YamiMediaCodec::VaapiDecoderH264::decodeNalu(YamiParser::H264::NalUnit*) (this=this@entry=0x14740e0, nalu=0x7fffa4011c90, nalu@entry=0x7fffb9c7e8d0) at vaapidecoder_h264.cpp:1688
#5  0x00007fffdfc941d3 in YamiMediaCodec::VaapiDecoderH264::decode(VideoDecodeBuffer*) (this=0x14740e0, buffer=<optimized out>) at vaapidecoder_h264.cpp:1727

xuguangxin:
The root cause of #495 is we have issue in error bit stream handle. When we read ue at end of stream. We will enter infinite loop. Core code is here.

```
for (uint32_t b = 0; !b; leadingZeroBits++) {
  b = read(1);
```

when we read to end of stream read always return 0, so !b never become false. It enter infinite loop.
We need return a value in read to indicate read failed.

### DIFF
--- a/codecparsers/bitReader.cpp
+++ b/codecparsers/bitReader.cpp
@@ -64,6 +64,14 @@ void BitReader::loadDataToCache(uint32_t nbytes)
     m_bitsInCache = nbytes << 3;
 }
 
+inline void BitReader::reload()
+{
+    assert(m_size >= m_loadBytes);
+    uint32_t remainingBytes = m_size - m_loadBytes;
+    if (remainingBytes > 0)
+        loadDataToCache(std::min(remainingBytes, CACHEBYTES));
+}
+
 inline uint32_t BitReader::extractBitsFromCache(uint32_t nbits)
 {
     if (!nbits)
@@ -74,35 +82,31 @@ inline uint32_t BitReader::extractBitsFromCache(uint32_t nbits)
     return tmp;
 }
 
-uint32_t BitReader::read(uint32_t nbits)
+bool BitReader::read(uint32_t& v, uint32_t nbits)
 {
     assert(nbits <= (CACHEBYTES << 3));
 
-    /* Firstly loading data to m_cache, only need to read aligned bytes.
-       So we can load 8 bytes which aligned with machine address in the next per time*/
-    if (!m_loadBytes) {
-        uint32_t alignedBytes = CACHEBYTES - ((uintptr_t)m_stream % CACHEBYTES);
-        uint32_t toBeLoadSize = std::min(alignedBytes, m_size);
-        loadDataToCache(toBeLoadSize);
-    }
-
-    uint32_t res = 0;
-
     if (nbits <= m_bitsInCache) {
-        res = extractBitsFromCache(nbits);
-    } else { /*not enough bits, need to save remaining bits
-               in current cache and then reload more bits*/
-        uint32_t toBeReadBits = nbits - m_bitsInCache;
-        uint32_t tmp = extractBitsFromCache(m_bitsInCache);
-        uint32_t remainingBytes = m_size - m_loadBytes;
-        uint32_t toBeLoadSize = std::min(remainingBytes, CACHEBYTES);
-        toBeReadBits = std::min(toBeReadBits, toBeLoadSize << 3);
-        tmp <<= toBeReadBits;
-        loadDataToCache(toBeLoadSize);
-        res = tmp | extractBitsFromCache(toBeReadBits);
+        v = extractBitsFromCache(nbits);
+        return true;
     }
+    /*not enough bits, need to save remaining bits
+      in current cache and then reload more bits*/
+    uint32_t toBeReadBits = nbits - m_bitsInCache;
+    uint32_t tmp = extractBitsFromCache(m_bitsInCache);
+    reload();
+    if (toBeReadBits > m_bitsInCache)
+        return false;
+    v = tmp << toBeReadBits | extractBitsFromCache(toBeReadBits);
+    return true;
+}
 
-    return res;
+uint32_t BitReader::read(uint32_t nbits)
+{
+    uint32_t res;
+    if (read(res, nbits))
+        return res;
+    return 0;
 }
 
 void BitReader::skip(uint32_t nbits)
--- a/codecparsers/nalReader.h
+++ b/codecparsers/nalReader.h
@@ -28,7 +28,9 @@ public:
     NalReader(const uint8_t *data, uint32_t size);
 
     /*parse Exp-Golomb coding*/
+    bool readUe(uint32_t& v);
     uint32_t readUe();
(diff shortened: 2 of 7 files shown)


xuguangxin:
also i refacted  original bitreader, make code more simpler.

codecparsers/bitReader_unittest.cpp
```diff
@@ -73,4 +74,26 @@ BITREADER_TEST(NoOverflow)
     EXPECT_EQ(0u, reader.getRemainingBitsCount());
 }
 
+BITREADER_TEST(ReadBeyondBoundary)
```
uartie: This test should go in "nalReader_unittest.cpp"

xuguangxin: ok, i will fix it

xuguangxin:
@uartie 
fixed

xuguangxin:
Hi @lizhong1008 
Pleas help to review.

Hi @wudping 
After lizhong +1ed for this, please help to run daily regression on it.

thanks

wudping:
okay, I get it. Guangxin.

seanvk:
lgtm +1

codecparsers/bitReader.cpp
```diff
@@ -74,35 +81,31 @@ inline uint32_t BitReader::extractBitsFromCache(uint32_t nbits)
...
+        v = extractBitsFromCache(nbits);
+        return true;
     }
+    /*not enough bits, need to save remaining bits
+      in current cache and then reload more bits*/
+    uint32_t toBeReadBits = nbits - m_bitsInCache;
+    uint32_t tmp = extractBitsFromCache(m_bitsInCache);
+    reload();
+    if (toBeReadBits > m_bitsInCache)
+        return false;
```
lizhong1008, line 99: Do we need add ERROR info here?

xuguangxin, line 99: No, we need provide error information in higher level. Only higher level(codec parser in most of case) can print useful information. I will demo this in h264 parser. You can check it after send h264 fixes.

xuguangxin:
This patch should work with https://github.com/01org/libyami-utils/pull/8. Please help review it too.

mypopydev:
+1

codecparsers/bitReader.cpp
```diff
@@ -64,6 +64,13 @@ void BitReader::loadDataToCache(uint32_t nbytes)
     m_bitsInCache = nbytes << 3;
 }
 
+inline void BitReader::reload()
+{
+    uint32_t remainingBytes = m_size - m_loadBytes;
```
lizhong1008, line 70: would better change uint32_t to int32_t to improve this function's robustnes. Else if m_size < m_loadBytes, remainingBytes will still > 0.

codecparsers/bitReader.cpp
lizhong1008, line 99: OK

codecparsers/nalReader.h
```diff
@@ -28,7 +28,9 @@ class NalReader : public BitReader
     NalReader(const uint8_t *data, uint32_t size);
 
     /*parse Exp-Golomb coding*/
+    bool readUe(uint32_t& v);
```
lizhong1008, line 31: Do we need to move it as private? Seems we just use uint32_t readUe() as public interface. Same as  bool readSe(int32_t& v) and bool read(uint32_t& v, uint32_t nbits) in bitreader.

codecparsers/bitReader.cpp
xuguangxin, line 70: no, m_loadBytes is current position, it never larger than m_size.

codecparsers/nalReader.h
xuguangxin, line 31: We should use this version for every codec parser.  Maybe we need remove uint32_t readUe() after all codec parser called this function.

codecparsers/bitReader.cpp
lizhong1008, line 70: I know it. What I mean is that we should make sure each independent function's robustness. m_size and m_loadBytes are changed by  the other functions, and this reload function doesn't make sure  m_size > m_loadBytes.

xuguangxin, line 70: how about add an assert here. We hope it expose as early as possible.

codecparsers/nalReader.h
lizhong1008, line 31: Yeah, but changing it in all codec parser may not be completed during a short time. Could we move it as private right now to avoid similar interfaces? After all codec parser updated, we can move it as public and remove uint32_t readUe().

codecparsers/bitReader.cpp
lizhong1008, line 70: Yeah，it is ok. If we use uint32_t, we need assert it.

codecparsers/nalReader.h
xuguangxin, line 31: No, I will submit a patch for h264 parser very soon. It will use this function.

codecparsers/bitReader.cpp
xuguangxin, line 70: @lizhong1008 fixed, I use assert because if we use`int32_t`, `std::min` will return `int32_t`. It's not match `void BitReader::loadDataToCache(uint32_t nbytes)` 's input type,

lizhong1008:
+1

lizhong1008:
@wudping , could you help to verify this patch?

wudping:
okay, LiZhong.

wudping:
verify + 1

xuguangxin:
merged as 509d0cda8071073e302a11b9f17cdfbc71144721

Issue #494, opened 2016-07-04
[Apache] h264 decoder DPB implement not follow the h264 spec


mypopydev:
Now h264 decoder DPB management is not follow the h264 spec, and always used 16 surfaces, it's not a good action in the low memory system.

lizhong1008:
We already followed the h264 DPB management spec, else it can't be decoded. 
16 decoded surfaces are the max value as spec. Yeah, we will optimize it to reduce DPB surfaces as h264 level value.

lizhong1008:
Hi Donging:
     Let's discuss how to do it before you start to do it. I can give you some guideline.

wudping:
Thanks LiZhong, ok.

xuguangxin:
fixed by https://github.com/01org/libyami/pull/519

PR #519, opened 2016-07-26
h264dec: reduce decoder buffering frame number.

lizhong1008:
1. remove H264_EXTRA_SURFACE_NUMBER, since it handled by
   vaapisurfaceallocator.
2. limit m_maxDecFrameBuffering as level_idc, it is benefit to reduce
   memory consume and decoding latency.
3. chech bump status to avoid infinite loop

### DIFF
--- a/decoder/vaapidecoder_h264.cpp
+++ b/decoder/vaapidecoder_h264.cpp
@@ -1085,8 +1085,10 @@ bool VaapiDecoderH264::DPB::add(const PicturePtr& picture)
         return output(picture);
     }
 
-    while (isFull())
-        bump();
+    while (isFull()) {
+        if (!bump())
+            return false;
+    }
 
     if (!isSecondField(picture))
         m_pictures.insert(picture);
--- a/decoder/vaapidecoder_h264.cpp
+++ b/decoder/vaapidecoder_h264.cpp
@@ -336,8 +336,7 @@ VaapiDecoderH264::DPB::DPB(OutputCallback output)
     , m_noOutputOfPriorPicsFlag(false)
     , m_maxFrameNum(0)
     , m_maxNumRefFrames(0)
-    , m_maxDecFrameBuffering(H264_MAX_REFRENCE_SURFACE_NUMBER) // should calc by
-// level_idc, fix it later
+    , m_maxDecFrameBuffering(H264_MAX_REFRENCE_SURFACE_NUMBER)
 {
 }
 
@@ -974,7 +973,8 @@ bool VaapiDecoderH264::DPB::init(const PicturePtr& picture,
                                  const PicturePtr& prevPicture,
                                  const SliceHeader* const slice,
                                  const NalUnit* const nalu, bool newStream,
-                                 bool contextChanged)
+                                 bool contextChanged,
+                                 uint32_t maxDecFrameBuffering)
 {
     const SharedPtr<PPS> pps = slice->m_pps;
     const SharedPtr<SPS> sps = pps->m_sps;
@@ -983,6 +983,7 @@ bool VaapiDecoderH264::DPB::init(const PicturePtr& picture,
     m_maxFrameNum = 1 << (sps->log2_max_frame_num_minus4 + 4);
     m_decRefPicMarking = slice->dec_ref_pic_marking;
     m_maxNumRefFrames = MAX(sps->num_ref_frames, 1);
+    m_maxDecFrameBuffering = maxDecFrameBuffering;
     if (isField(picture))
         m_maxNumRefFrames *= 2;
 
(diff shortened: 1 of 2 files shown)


lizhong1008:
it is to fix https://github.com/01org/libyami/issues/494

lizhong1008:
Regression test is on going.  @wudping will update the test result when completed.

decoder/vaapidecoder_h264.cpp
```diff
@@ -1520,15 +1521,89 @@ YamiStatus VaapiDecoderH264::decodeCurrent()
     return status;
 }
 
+uint32_t calcMaxDecFrameBufferingNum(const SharedPtr<SPS> sps)
```
xuguangxin: should use 
`const SharedPtr<SPS>& sps`
to avoid copy.

decoder/vaapidecoder_h264.cpp
```diff
@@ -1520,15 +1521,89 @@ YamiStatus VaapiDecoderH264::decodeCurrent()
...
+
 bool VaapiDecoderH264::isDecodeContextChanged(const SharedPtr<SPS> sps)
 {
-    uint8_t surfaceNumber = H264_MAX_REFRENCE_SURFACE_NUMBER
-                            + H264_EXTRA_SURFACE_NUMBER;
+    uint32_t maxDecFrameBuffering;
+
+    maxDecFrameBuffering = calcMaxDecFrameBufferingNum(sps);
+
+    if (maxDecFrameBuffering > H264_MAX_REFRENCE_SURFACE_NUMBER)
```
xuguangxin, line 1600: could we hide this in calcMaxDecFrameBufferingNum?

decoder/vaapidecoder_h264.cpp
```diff
@@ -1520,15 +1521,89 @@ YamiStatus VaapiDecoderH264::decodeCurrent()
...
+    case 50:
+        maxDpbMbs = 110400;
+        break;
+    case 51:
+    case 52:
+        maxDpbMbs = 184320;
+        break;
+    default:
+        ERROR("undefined level_idc");
+        assert(0);
```
xuguangxin: This will crash when we handle error bitstream, Could we assign a reasonable default value here?

decoder/vaapidecoder_h264.cpp
```diff
@@ -1084,8 +1085,10 @@ bool VaapiDecoderH264::DPB::add(const PicturePtr& picture)
         return output(picture);
     }
 
-    while (isFull())
-        bump();
+    while (isFull()) {
+        if (!bump())
```
xuguangxin, line 1089: Why we need check this? It's mean we are full, but we can't bump any buffer? This not reasonable since we should bump buffer at full, else we should lose buffer or something.

decoder/vaapidecoder_h264.cpp
lizhong1008: Seems not necessary, since here sps is a sharedptr.

decoder/vaapidecoder_h264.cpp
lizhong1008, line 1600: Put it to calcMaxDecFrameBufferingNum need duplicate this code for both sps->m_vui.max_dec_frame_buffering and maxDpbFrames.  Another way to to wrap this code to another funciton named checkMaxDecFrameBufferingNum, but I don't think it's necessary.

decoder/vaapidecoder_h264.cpp
lizhong1008: level_idc is existed in bitstream, it's not input parameter, user can't change it, which is differnent from the case of https://github.com/01org/libyami/issues/439.

decoder/vaapidecoder_h264.cpp
lizhong1008, line 1089: Maybe you misunderstand it. if it is full , we need to bump it. But if we get error from the bump function, we need jump out the loop.

decoder/vaapidecoder_h264.cpp
xuguangxin: This code will call SharedPtr's copy constructor. const SharedPtr<SPS>& sps will not.

decoder/vaapidecoder_h264.cpp
xuguangxin: Yes, But we are handle real world situation. We will face error bitstream(it usually from network or bit damaged disk)

lizhong1008: H264 spec required sps/pps NAL unit must be transferred more robustly, that is why it separates different NAL types. On the other hand, we need to handle some incorrect sliced_date, but we can't decode bitsteams with incorrect sps.

decoder/vaapidecoder_h264.cpp
lizhong1008: ok, I will update it.

decoder/vaapidecoder_h264.cpp
xuguangxin, line 1600: ok.

decoder/vaapidecoder_h264.cpp
lizhong1008: For undefined level_idc, I'm going to set maxDpbMbs  to 184320 (equal to level 52).

xuguangxin:
Seems this patch will introduce regression in daily regression test. @lizhong1008 will do more investigate more details about this

lizhong1008:
Yes, this patch requires h264 clips  strictly following h264 spec (Table A-1). Most of clips follow spec, but a little clips don't follow. Let's me try to find a way to let these clips work.

xuguangxin:
@lizhong1008 @mypopydev 
Even this patch have conformance drop. However, it do fix long latency issue. 
If we test this patch, and find it is no big conformance drop compare to master. Do you think it is ok to merge this?

mypopydev:
+1

decoder/vaapidecoder_h264.cpp
xuguangxin: Hi @lizhong1008 
any update?

decoder/vaapidecoder_h264.cpp
xuguangxin: @lizhong1008 any update?

lizhong1008:
updated

xuguangxin:
@wudping Please help to do test all h264 clips in daily regression. Make sure it is no significant drop for conformance test.
thanks

wudping:
okay, Guangxin.

wudping:
verify + 1, a little performance drop.

xuguangxin:
you mean conformance drop for h264 right?

xuguangxin:
merged as e0a5f44da3c08637e0833eb0e32e1c284c90dc11

PR #522, opened 2016-07-29
Apache

halleyzhao:


v4l2/v4l2_codecbase.cpp
```diff
@@ -391,7 +391,7 @@ int32_t V4l2CodecBase::ioctl(int command, void* arg)
                 ERROR("unknown request buffer type: %d", reqbufs->type);
                 break;
             }
-            ASSERT(port == INPUT || reqbufs->memory == m_memoryMode[port]);
+            // ASSERT(port == INPUT || reqbufs->memory == m_memoryMode[port]);
```
xuguangxin, line 394: why remove this assert, do we need some comment?

xuguangxin:
Most ok to me.Wait for comment about assert

v4l2/v4l2_codecbase.cpp
halleyzhao, line 394: VIDIOC_REQBUFS init memory type of input/output buffer.
1.  chromeos assumed input/output buffer memory type is MMAP, so we init m_memoryModep[port] to MMAP and ASSERT() here
2. next, we add USRPTR memory type for input buffer, it means 'port == INPUT' needn't assert memory type (but accept memmory type set from client).
3. now we add ANDROID_BUFFER_HANDLE memory type for output buffer, then 'port == OUTPUT' needn't ASSERT memory type either.

xuguangxin:
ok, +1

interface/VideoCommonDefs.h
```diff
@@ -94,7 +94,7 @@ typedef enum {
     VIDEO_DATA_MEMORY_TYPE_DMA_BUF,      // share buffer with camera device etc
     VIDEO_DATA_MEMORY_TYPE_SURFACE_ID,  // it can be used for surface sharing of transcoding, benefits suppressed rendering as well.
                                         //it is discouraged to use it for video rendering.
-    VIDEO_DATA_MEMORY_TYPE_ANDROID_NATIVE_BUFFER, // ANativeWindowBuffer for android
```
xuguangxin, line 97: We need increase API version for this  in configure.ac

xuguangxin:
Hi @halleyzhao 
Please change 
`m4_define([yami_api_micro_version], 2)`
to
`m4_define([yami_api_micro_version], 3)`
in configure.ac

xuguangxin:
merged as dd36dee96a218fa04ccc1c94ea74565821e99019. thanks @halleyzhao

PR #524, opened 2016-08-01
h264dec: check read beyond boundary for every bit

xuguangxin:
We need check read beyond boundary. To make sure we can detect error bitstream as early as possible.
If we do not check this, maybe we will encounter a serious error.
#495 is good example

### DIFF
--- a/codecparsers/bitReader.h
+++ b/codecparsers/bitReader.h
@@ -35,6 +35,15 @@ public:
     /* will return 0 if not enough data*/
     uint32_t read(uint32_t nbits);
 
+    /*TODO: change to this to read, and change read to readBits */
+    template <class T>
+    inline bool readT(T& v, uint32_t nbits);
+
+    template <class T>
+    inline bool readT(T& v);
+
+    inline bool readT(bool& v);
+
     /*read the next nbits bits from the bitstream but not advance the bitstream pointer*/
     uint32_t peek(uint32_t nbits) const;
 
@@ -72,6 +81,27 @@ private:
     inline uint32_t extractBitsFromCache(uint32_t nbits);
     inline void reload();
 };
+
+template <class T>
+bool BitReader::readT(T& v, uint32_t nbits)
+{
+    uint32_t tmp;
+    if (!read(tmp, nbits))
+        return false;
+    v = tmp;
+    return true;
+}
+
+template <class T>
+bool BitReader::readT(T& v)
+{
+    return readT(v, sizeof(v) << 3);
+}
+
+bool BitReader::readT(bool& v)
+{
+    return readT(v, 1);
+}
 
 } /*namespace YamiParser*/
 
--- a/codecparsers/h264Parser.cpp
+++ b/codecparsers/h264Parser.cpp
@@ -22,6 +22,38 @@
 
 #include <math.h>
 
+#define READ(f)                             \
+    do {                                    \
+        if (!br.readT(f)) {                 \
+            ERROR("failed to read %s", #f); \
+            return false;                   \
+        }                                   \
+    } while (0)
+
+#define READ_BITS(f, bits)                              \
+    do {                                                \
+        if (!br.readT(f, bits)) {                       \
+            ERROR("failed to read %d to %s", bits, #f); \
+            return false;                               \
+        }                                               \
+    } while (0)
+
+#define READ_UE(f)                            \
+    do {                                      \
+        if (!br.readUe(f)) {                  \
+            ERROR("failed to readUe %s", #f); \
+            return false;                     \
+        }                                     \
+    } while (0)
+
+#define READ_SE(f)                            \
+    do {                                      \
+        if (!br.readSe(f)) {                  \
+            ERROR("failed to readSe %s", #f); \
+            return false;                     \
+        }                                     \
+    } while (0)
+
 namespace YamiParser {
 namespace H264 {
 
@@ -82,14 +114,14 @@ static int32_t searchStartPos(const uint8_t* src, uint32_t size)
 }
 
 //7.3.2.1.1.1 Scaling list syntax
-static void scalingList(NalReader& nr, uint8_t* sl, uint32_t size, uint32_t index)
+static bool scalingList(NalReader& br, uint8_t* sl, uint32_t size, uint32_t index)
 {
     uint8_t lastScale = 8;
     uint8_t nextScale = 8;
     int32_t delta_scale;
     for (uint32_t j = 0; j < size; j++) {
         if (nextScale) {
-            delta_scale = nr.readSe();
+            READ_SE(delta_scale);
             nextScale = (lastScale + delta_scale + 256) % 256;
         }
         if (!j && !nextScale) {
@@ -99,6 +131,7 @@ static void scalingList(NalReader& nr, uint8_t* sl, uint32_t size, uint32_t inde
         sl[j] = (!nextScale) ? lastScale : nextScale;
         lastScale = sl[j];
     }
+    return true;
 }
 
 PPS::~PPS()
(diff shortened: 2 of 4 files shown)


xuguangxin:
Hi @chivakker @YuJiankang @wudping 
Please help to review this.

codecparsers/bitReader.h
```diff
@@ -35,6 +35,15 @@ class BitReader {
     /* will return 0 if not enough data*/
     uint32_t read(uint32_t nbits);
 
+    /*TODO: change to this to read, and change read to readBits */
```
chivakker, line 38: Actually change this, avoid adding TODO statements.

codecparsers/h264Parser.cpp
```diff
@@ -22,6 +22,38 @@
...
+    } while (0)
+
+#define READ_SE(f)                            \
+    do {                                      \
+        if (!br.readSe(f)) {                  \
+            ERROR("failed to readSe %s", #f); \
+            return false;                     \
+        }                                     \
+    } while (0)
+
```
chivakker, line 56: Can all the above be inline functions?

codecparsers/h264Parser.cpp
```diff
@@ -82,14 +114,14 @@ static int32_t searchStartPos(const uint8_t* src, uint32_t size)
...
-static void scalingList(NalReader& nr, uint8_t* sl, uint32_t size, uint32_t index)
+static bool scalingList(NalReader& br, uint8_t* sl, uint32_t size, uint32_t index)
 {
     uint8_t lastScale = 8;
     uint8_t nextScale = 8;
     int32_t delta_scale;
     for (uint32_t j = 0; j < size; j++) {
         if (nextScale) {
-            delta_scale = nr.readSe();
+            READ_SE(delta_scale);
```
chivakker, line 124: Return gets hidden in the macro; making it a inline function will have a return value and hence better readable on this function.

codecparsers/h264Parser.cpp
```diff
@@ -107,29 +140,31 @@ PPS::~PPS()
...
+    READ(m_svc.idr_flag);
+    READ_BITS(m_svc.priority_id, 6);
+    READ(m_svc.no_inter_layer_pred_flag);
+    READ_BITS(m_svc.dependency_id, 3);
+    READ_BITS(m_svc.quality_id, 4);
+    READ_BITS(m_svc.temporal_id, 3);
+    READ(m_svc.use_ref_base_pic_flag);
+    READ(m_svc.discardable_flag);
+    READ(m_svc.output_flag);
+    READ_BITS(m_svc.reserved_three_2bits, 2);
```
chivakker, line 154: with a inline function this can be changed to return read() || read() || read() || read_bits(), etc.

codecparsers/h264Parser.cpp
```diff
@@ -107,29 +140,31 @@ PPS::~PPS()
...
-    m_mvc.temporal_id = br.read(3);
-    m_mvc.anchor_pic_flag = br.read(1);
-    m_mvc.inter_view_flag = br.read(1);
+    READ(m_mvc.non_idr_flag);
+    READ_BITS(m_mvc.priority_id, 6);
+    READ_BITS(m_mvc.view_id, 10);
+    READ_BITS(m_mvc.temporal_id, 3);
+    READ(m_mvc.anchor_pic_flag);
+    READ(m_mvc.inter_view_flag);
+    return true;
```
chivakker, line 167: ditto

codecparsers/h264Parser.cpp
```diff
@@ -349,12 +386,16 @@ bool Parser::parseSps(SharedPtr<SPS>& sps, const NalUnit* nalu)
...
             uint32_t loop = (sps->chroma_format_idc != 3) ? 8 : 12;
             for (uint32_t i = 0; i < loop; i++) {
-                sps->seq_scaling_list_present_flag[i] = nr.read(1);
+                READ(sps->seq_scaling_list_present_flag[i]);
                 if (sps->seq_scaling_list_present_flag[i]) {
-                    if (i < 6)
-                        scalingList(nr, sps->scaling_lists_4x4[i], 16, i);
-                    else
-                        scalingList(nr, sps->scaling_lists_8x8[i - 6], 64, i);
+                    if (i < 6) {
```
chivakker, line 391: change this 6 to a macro, while you're modifying this piece of code.

codecparsers/nalReader.h
```diff
@@ -43,6 +45,24 @@ class NalReader : public BitReader
...
+
+bool NalReader::readSe(int8_t& v)
+{
+    int32_t tmp;
+    if (!readSe(tmp))
+        return false;
+    v = tmp;
+    return true;
+}
+
```
chivakker, line 65: and these are defined as inline functions

codecparsers/bitReader.h
xuguangxin, line 38: We should do this after vc1, mpeg2, hevc start use new readT.

codecparsers/h264Parser.cpp
xuguangxin, line 56: No, if we use a function we still need judge return value. Use macro makes code cleaner

codecparsers/h264Parser.cpp
xuguangxin, line 124: If we use a function, we need write a "if else" here. Consider they are too many place need read bitstream. It makes code unreadble

codecparsers/h264Parser.cpp
xuguangxin, line 154: If we change to inline function we need write 10 "if else" for this function. It not need to be

codecparsers/h264Parser.cpp
xuguangxin, line 391: We can use other patch to fix this, this patch only focus on check read beyond boundary things.

codecparsers/nalReader.h
xuguangxin, line 65: Yes. It's defined in L32, it's no need repeat inline here.

xuguangxin:
Hi @chivakker 
This patch will check every read operation to make sure it not read beyond boundary.
Return a reasonable value is not enough for some case. 
https://github.com/01org/libyami/pull/514 is good example.

Here is original code from NalReader::readUe

```
for (uint32_t b = 0; !b; leadingZeroBits++) {
  b = read(1);
```

We will return 0 if read(1) read beyond boundary.But this will make this loop trap to a infinite loop. This why we need change this code to this
https://github.com/01org/libyami/pull/514/files#diff-7409028df9bce563cd348eb36025945fR67
Same check need apply to codec parser too.
This why I submit this patch.

However, for codec parser, we’d better not add “if else” branch for every read. It is will be too many places. Therefore, I use a macro to do all return check.

codecparsers/bitReader.h
chivakker, line 38: Yes, in that case please go ahead and make the changes across the board.

chivakker:
I don't want to start a discussion between inline and macros in C++.  Let's keep the macros then.  The only comment I have then is that it would be better to make the change across the board to avoid the future renaming.

xuguangxin:
inline function is not a replacement for macro. The vp8 parser use similar macro too.You can refer to https://github.com/01org/libyami/blob/apache/codecparsers/vp8_parser.cpp#L54

chivakker:
@xuguangxin the vp8 parser example is also a good data point that it is fine to do it that way.  Pardon the confusion.

xuguangxin:
@chivakker @YuJiankang 
Please +1 for me. if you think it's ok

YuJiankang:
+1

chivakker:
Can we hold on merging this patch until we have the other codecs fixed?

xuguangxin:
Hi @chivakker 
Why? This will not have side effects to other codec.

codecparsers/bitReader.h
```diff
@@ -73,6 +82,27 @@ class BitReader {
...
+bool BitReader::readT(T& v)
+{
+    return readT(v, sizeof(v) << 3);
+}
+
+bool BitReader::readT(bool& v)
+{
+    return readT(v, 1);
+}
+
```
chivakker, line 105: Is peek,skip supposed to be replaced by readT as well?  mpeg2 uses peek.  If those changes are made and read is not renamed to readT then mpeg2 doesn't need any change.

xuguangxin, line 105: The main benefit of readT is it will guess how many bits should read base on the type of `v`. if `v`is `uint32_t`it will read 32 bits, if it’s `bool`it will read 1 bits.
I see peek code in mpeg2.

> bitsToParse = kVLCTable[i][0];
> bitReaderPeekBits(bitsToParse, &mbVLC);

Seems it will read bitsToParse bits. Therefore, you cannot write a template based peek for this. You may need change peek to `bool peek(uint32_t&v, uint32_t bit)`, so you can check peek status

wudping:
verify + 1

xuguangxin:
merged as 2844dcfb5c1c80f6c78d50ad8aa298fc849cb3ee

PR #529, opened 2016-08-03
hevcenc: enable deblocking as default

lizhong1008:
1. Enable deblocking as default.
2. deblocking_filter_override_flag isn't set, so even
   slice_deblocking_filter_disabled_flag is set to 1, it won't be written
   into bitstream, in this case, slice_deblocking_filter_disabled_flag
   should be same as pps_deblocking_filter_disabled_flag.


mypopydev:
+1

xuguangxin:
+1

PR #531, opened 2016-08-09
vpp : change the NONE define for the caller.

mypopydev:
before [MIN, MAX] means enable with the scope, used MAX+1 = NONE means
disable this function, it's strange for the caller. Change to used
MIN-1 = NONE to disable this function, it's more intuitive for the caller.
And bump the version.

Signed-off-by: Jun Zhao jun.zhao@intel.com

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,7 +8,7 @@ m4_define([yami_api_major_version], 0)
 # update this for every release when micro version large than zero
 m4_define([yami_api_minor_version], 2)
 # change this for any api change
-m4_define([yami_api_micro_version], 3)
+m4_define([yami_api_micro_version], 4)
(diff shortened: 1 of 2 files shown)


xuguangxin:
+1

chivakker:
why are you bumping the micro version?

mypopydev:
@chivakker 
As I know, when change the interface header file, need to bump the version

xuguangxin:
@chivakker , yes we need bump the version when we change anything in header. Do you have concern about this?

xuguangxin:
@chivakker , I will merge this first, since other commit dependent version number of this

xuguangxin:
thanks @mypopydev

PR #534, opened 2016-08-16
Vp9encoder Initial implementation

chivakker:
Initial Implementation for vp9 encoder on libyami.

It supports:

CQP CBR, key frame and inter frame.
Some parameters will be available to command line on yamiencode with future patches.
Current reference scheme uses golden, alternate reference to the previous key frame, last reference frame is the previous encoded frame.

For this patch to work properly it required some fixes on the driver level.  So, to actually validate it the recommended sha's for libva and libva-intel-driver are:

libva:
commit  b27feb9b8e62725fed959c090feb2f1c853f74b3
tree    65a88a229d71fbc5f50c7f6e9a890f911a7bd994
parent  695f99ef0405cf4255e7767b44effb0da2fe706e

libva-intel-driver: 
commit  98555d27c8291fe2511355d53924d17a5263dc22
tree    c34ff240a83e1ab20f9862ab952318ac147560ae
parent  3e115315e9ef2df1b3d8f64bced69d0f685c6651

### DIFF
--- a/encoder/vaapiencoder_vp9.cpp
+++ b/encoder/vaapiencoder_vp9.cpp
@@ -1,0 +1,322 @@
+/*
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "vaapiencoder_vp9.h"
+#include "common/scopedlogger.h"
+#include "common/common_def.h"
+#include "vaapi/vaapicontext.h"
+#include "vaapi/vaapidisplay.h"
+#include "vaapicodedbuffer.h"
+#include "vaapiencpicture.h"
+#include "vaapiencoder_factory.h"
+#include <algorithm>
+
+namespace YamiMediaCodec {
+
+// each frame can have at most 3 reference frames, golden, alt and last
+enum maxSizeValues {
+    kMaxReferenceFrames = 8,
+    kMaxWidth = 4096,
+    kMaxHeight = kMaxWidth,
+    kMaxHeaderSize=kMaxWidth
+};
+
+enum VP9QPValues { kMinQPValue = 9, kDefaultQPValue = 60, kMaxQPValue = 127 };
+
+enum VP9FrameType { kKeyFrame, kInterFrame };
+
+enum VP9LevelValues { kDefaultSharpnessLevel, kDefaultFilterLevel = 10 };
+
+class VaapiEncPictureVP9 : public VaapiEncPicture {
+public:
+    VaapiEncPictureVP9(const ContextPtr& context, const SurfacePtr& surface,
+                       int64_t timeStamp)
+        : VaapiEncPicture(context, surface, timeStamp)
+    {
+        return;
+    }
+
+    VAGenericID getCodedBufferID() { return m_codedBuffer->getID(); }
+};
+
+VaapiEncoderVP9::VaapiEncoderVP9()
+    : m_frameCount(0)
+{
+    m_videoParamCommon.profile = VAProfileVP9Profile0;
+    m_videoParamCommon.rcParams.minQP = kMinQPValue;
+    m_videoParamCommon.rcParams.maxQP = kMaxQPValue;
+    m_videoParamCommon.rcParams.initQP = kDefaultQPValue;
+
+    // add extra surfaces to operate due to kMaxReferenceFrames
+    // vaapi_encoder class will create 5 extra surfaces already
+    m_maxOutputBuffer = kMaxReferenceFrames;
+}
+
+VaapiEncoderVP9::~VaapiEncoderVP9() {}
+
+YamiStatus VaapiEncoderVP9::getMaxOutSize(uint32_t* maxSize)
+{
+    FUNC_ENTER();
+    *maxSize = m_maxCodedbufSize;
+    return YAMI_SUCCESS;
+}
+
+void VaapiEncoderVP9::resetParams()
+{
+    m_maxCodedbufSize = width() * height() * 3 / 2;
+
+    // adding extra padding.  In particular small resolutions require more
+    // space depending on other quantization parameters during execution. The
+    // value below is a good compromise
+    m_maxCodedbufSize += kMaxHeaderSize;
+}
+
+YamiStatus VaapiEncoderVP9::start()
+{
+    FUNC_ENTER();
+    resetParams();
+    return VaapiEncoderBase::start();
+}
+
+void VaapiEncoderVP9::flush()
+{
+    FUNC_ENTER();
+    m_frameCount = 0;
+    m_reference.clear();
+    VaapiEncoderBase::flush();
+}
+
+YamiStatus VaapiEncoderVP9::stop()
+{
+    flush();
+    return VaapiEncoderBase::stop();
+}
+
+YamiStatus VaapiEncoderVP9::setParameters(VideoParamConfigType type,
+                                          Yami_PTR videoEncParams)
+{
+    YamiStatus status = YAMI_SUCCESS;
+    FUNC_ENTER();
+    if (!videoEncParams)
+        return YAMI_INVALID_PARAM;
+
+    switch (type) {
+    default:
+        status = VaapiEncoderBase::setParameters(type, videoEncParams);
+        break;
+    }
+    return status;
+}
+
+YamiStatus VaapiEncoderVP9::getParameters(VideoParamConfigType type,
+                                          Yami_PTR videoEncParams)
+{
+    FUNC_ENTER();
+    if (!videoEncParams)
+        return YAMI_INVALID_PARAM;
+
+    return VaapiEncoderBase::getParameters(type, videoEncParams);
+}
+
+YamiStatus VaapiEncoderVP9::doEncode(const SurfacePtr& surface,
+                                     uint64_t timeStamp, bool forceKeyFrame)
+{
+    YamiStatus ret;
+    if (!surface)
+        return YAMI_INVALID_PARAM;
+
+    PicturePtr picture(new VaapiEncPictureVP9(m_context, surface, timeStamp));
+
+    m_frameCount %= keyFramePeriod();
+    picture->m_type = (m_frameCount ? VAAPI_PICTURE_P : VAAPI_PICTURE_I);
+    m_frameCount++;
+
+    CodedBufferPtr codedBuffer
+        = VaapiCodedBuffer::create(m_context, m_maxCodedbufSize);
+    if (!codedBuffer)
+        return YAMI_OUT_MEMORY;
+    picture->m_codedBuffer = codedBuffer;
+    codedBuffer->setFlag(ENCODE_BUFFERFLAG_ENDOFFRAME);
+    if (picture->m_type == VAAPI_PICTURE_I) {
+        codedBuffer->setFlag(ENCODE_BUFFERFLAG_SYNCFRAME);
+    }
+    ret = encodePicture(picture);
+    if (ret != YAMI_SUCCESS) {
+        return ret;
+    }
+    output(picture);
+    return YAMI_SUCCESS;
+}
+
+bool VaapiEncoderVP9::fill(VAEncSequenceParameterBufferVP9* seqParam) const
+{
+    seqParam->max_frame_width = kMaxWidth;
+    seqParam->max_frame_height = kMaxHeight;
+
+    seqParam->bits_per_second = bitRate();
+    seqParam->intra_period = intraPeriod();
+    seqParam->kf_min_dist = 1;
+    seqParam->kf_max_dist = intraPeriod();
+
+    return true;
+}
+
+// Fills in VA picture parameter buffer
+bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
+                           const PicturePtr& picture,
+                           const SurfacePtr& surface) const
+{
+    picParam->reconstructed_frame = surface->getID();
+    picParam->coded_buf = picture->getCodedBufferID();
+
+    if (picture->m_type == VAAPI_PICTURE_I) {
+        for (uint32_t i = 0; i < kMaxReferenceFrames ; i++)
+            picParam->reference_frames[i] = VA_INVALID_SURFACE;
+    }
+    else {
+        picParam->refresh_frame_flags = 0x01; // refresh last frame
+        picParam->pic_flags.bits.frame_type = kInterFrame;
+
+        ReferenceQueue::const_iterator it = m_reference.begin();
+
+        for (uint32_t i = 0; it != m_reference.end(); ++it, i++)
+            picParam->reference_frames[i] = (*it)->getID();
+
+        // last/golden/alt is used as reference frame. L0 forward
+        picParam->ref_flags.bits.ref_frame_ctrl_l0 = 0x7;
+
+        // golden and alt are last KeyFrame
+        // last is last decoded frame
+        picParam->ref_flags.bits.ref_last_idx = 0;
+        picParam->ref_flags.bits.ref_gf_idx = 1;
+        picParam->ref_flags.bits.ref_arf_idx = 2;
+        picParam->pic_flags.bits.frame_context_idx = 0;
+    }
+
+    picParam->frame_width_src = width();
+    picParam->frame_height_src = height();
+    picParam->frame_width_dst = width();
+    picParam->frame_height_dst = height();
+
+    picParam->pic_flags.bits.show_frame = 1;
+
+    picParam->luma_ac_qindex = kDefaultQPValue;
+    picParam->luma_dc_qindex_delta = 1;
+    picParam->chroma_ac_qindex_delta = 1;
+    picParam->chroma_dc_qindex_delta = 1;
+
+    picParam->filter_level = kDefaultFilterLevel;
+    picParam->sharpness_level = kDefaultSharpnessLevel;
+
+    return true;
+}
+
+bool VaapiEncoderVP9::fill(
+    VAEncMiscParameterTypeVP9PerSegmantParam* segParam) const
+{
+    return true;
+}
+
+bool VaapiEncoderVP9::ensureSequence(const PicturePtr& picture)
+{
+    if (picture->m_type != VAAPI_PICTURE_I)
+        return true;
+
+    VAEncSequenceParameterBufferVP9* seqParam;
+    if (!picture->editSequence(seqParam) || !fill(seqParam)) {
+        ERROR("failed to create sequence parameter buffer (SPS)");
+        return false;
+    }
+    return true;
+}
+
+bool VaapiEncoderVP9::ensurePicture(const PicturePtr& picture,
+                                    const SurfacePtr& surface)
+{
+    VAEncPictureParameterBufferVP9* picParam;
+
+    if (!picture->editPicture(picParam) || !fill(picParam, picture, surface)) {
+        ERROR("failed to create picture parameter buffer (PPS)");
+        return false;
+    }
+    return true;
+}
+
+bool VaapiEncoderVP9::ensureQMatrix(const PicturePtr& picture)
+{
+    VAEncMiscParameterTypeVP9PerSegmantParam* segmentParam;
+
+    if (picture->m_type != VAAPI_PICTURE_I)
+        return true;
+
+    if (!picture->editQMatrix(segmentParam) || !fill(segmentParam)) {
+        ERROR("failed to create qMatrix");
+        return false;
+    }
+    return true;
+}
+
+bool VaapiEncoderVP9::referenceListUpdate(const PicturePtr& pic,
+                                          const SurfacePtr& recon)
+{
+
+    if (pic->m_type == VAAPI_PICTURE_I) {
+        m_reference.clear();
+        m_reference.insert(m_reference.end(), kMaxReferenceFrames, recon);
+    }
+    else {
+        m_reference.pop_front();
+        m_reference.push_front(recon);
+    }
+
+    return true;
+}
+
+YamiStatus VaapiEncoderVP9::encodePicture(const PicturePtr& picture)
+{
+    YamiStatus ret = YAMI_FAIL;
+    SurfacePtr reconstruct = createSurface();
+    if (!reconstruct)
+        return ret;
+
+    if (!ensureSequence(picture))
+        return ret;
+
+    if (!ensureQMatrix(picture))
+        return ret;
+
+    if (!ensureMiscParams(picture.get()))
+        return ret;
+
+    if (!ensurePicture(picture, reconstruct))
+        return ret;
+
+    if (!picture->encode())
+        return ret;
+
+    if (!referenceListUpdate(picture, reconstruct))
+        return ret;
+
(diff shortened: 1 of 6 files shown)


encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+// Fills in VA picture parameter buffer
+bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
+                           const PicturePtr& picture,
+                           const SurfacePtr& surface) const
+{
+    picParam->reconstructed_frame = surface->getID();
+    picParam->coded_buf = picture->getCodedBufferID();
+
+    if (picture->m_type == VAAPI_PICTURE_I) {
+        memset(picParam->reference_frames, 0xFF,
```
xuguangxin: Better set reference_frame[i] to VA_INVALID_SURFACE instead of assign it to 0xff. We cannot do any assumption on the VA_INVALID_SURFACE value.

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+    else {
+        picParam->refresh_frame_flags = 0x01; //refresh last frame
+        picParam->pic_flags.bits.frame_type = kInterFrame;
+
+        ReferenceQueue::const_iterator it = m_reference.begin();
+
+        for (uint32_t i = 0; it != m_reference.end(); ++it, i++)
+            picParam->reference_frames[i] = (*it)->getID();
+
+        it = m_reference.begin();
```
xuguangxin: What this for? It never used after this line

encoder/vaapiencoder_vp9.h
```diff
@@ -0,0 +1,79 @@
...
+    virtual void flush();
+    virtual YamiStatus stop();
+
+    virtual YamiStatus getParameters(VideoParamConfigType type, Yami_PTR);
+    virtual YamiStatus setParameters(VideoParamConfigType type, Yami_PTR);
+    virtual YamiStatus getMaxOutSize(uint32_t* maxSize);
+
+protected:
+    virtual YamiStatus doEncode(const SurfacePtr&, uint64_t timeStamp,
+                                bool forceKeyFrame = false);
```
xuguangxin, line 45: seems not format by clang-format

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+    if (ret != YAMI_SUCCESS) {
+        return ret;
+    }
+    output(picture);
+    return YAMI_SUCCESS;
+}
+
+bool VaapiEncoderVP9::fill(VAEncSequenceParameterBufferVP9* seqParam) const
+{
+    seqParam->max_frame_width = kMaxWidth;
```
xuguangxin, line 170: For our case, this should be width(). Not 8096. Actually we can't do any assumption on max width height. It's decide by driver

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+        ReferenceQueue::const_iterator it = m_reference.begin();
+
+        for (uint32_t i = 0; it != m_reference.end(); ++it, i++)
+            picParam->reference_frames[i] = (*it)->getID();
+
+        it = m_reference.begin();
+
+        // last/golden/alt is used as reference frame. L0 forward
+        picParam->ref_flags.bits.ref_frame_ctrl_l0 = 0x7;
```
xuguangxin, line 203: l0 means vp9 have B frame?

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+{
+    FUNC_ENTER();
+    if (!videoEncParams)
+        return YAMI_INVALID_PARAM;
+
+    return VaapiEncoderBase::getParameters(type, videoEncParams);
+}
+
+YamiStatus VaapiEncoderVP9::doEncode(const SurfacePtr& surface,
+                                     uint64_t timeStamp, bool forceKeyFrame)
```
xuguangxin, line 139: Seems we ignored forceKeyFrame ?

xuguangxin:
@mypopydev , Please help to review this

encoder/vaapiencoder_vp9.cpp
chivakker, line 203: l0 means forward reference with 3 frames (golden, alt and last), key frame and inter frame can be references.

encoder/vaapiencoder_vp9.cpp
chivakker: This can go away.

encoder/vaapiencoder_vp9.cpp
chivakker: It has to be 0xFF, check va_enc_vp9.h reference_frames[8]

encoder/vaapiencoder_vp9.cpp
chivakker, line 139: Yes, I think this has to be implemented.  I think also vp8 encoder is not using it.

encoder/vaapiencoder_vp9.cpp
chivakker, line 170: I need to consult this with the driver folks.

encoder/vaapiencoder_vp9.cpp
xuguangxin: ;), I see the comments on the fields. However, the prototype is VASurfaceID. VASurfaceID is unsigned int. `memset(picParam->reference_frames, 0xFF,sizeof(picParam->reference_frames));` will set 0xffffffff to picParam->reference_frames. not 127. 
Invalid value for VASurfaceID is VA_INVALID_SURFACE. Maybe it is a wrong comment. We should send a patch to libva.

encoder/vaapiencoder_vp9.cpp
xuguangxin, line 203: I see ref_frame_ctrl_l1 in define too. When to use l1?  Is it improve image quality if we use l0 and l1 at same time.

encoder/vaapiencoder_vp9.cpp
xuguangxin, line 139: Thanks for point out, seems we need other patch for vp8

seanvk, line 139: Good point we need to be able to forceKeyFrame for videoconferencing.  I'm surprised we don't have it for VP8.

encoder/vaapiencoder_vp9.cpp
seanvk, line 170: That is actually the max width per platform, we should not be using that here...

encoder/vaapiencoder_vp9.cpp
chivakker, line 139: So far on both VP8/VP9 this is controlled by the intraPeriod parameter given in the command line.  It needs some work to translate the intraPeriod value to the doEncode call.

encoder/vaapiencoder_vp9.cpp
chivakker, line 170: This is to be used with the spatial conversion when implemented.  Of course, if the spatial conversion is done to downscale it won't be a problem.  This can be changed to what all platforms support, e.g. 4K

encoder/vaapiencoder_vp9.cpp
xuguangxin, line 139: It not cover by our test application. Maybe we can write a unit test to enforce this.

encoder/vaapiencoder_vp9.cpp
xuguangxin, line 170: From the comments in va_enc_vp9.h.  This field is for max resolution in video sequence. It has should not related to platform. For our case, we never change resolution in our encoder (at least in yami). So it should be width().

encoder/vaapiencoder_vp9.cpp
chivakker, line 203: I prefer to use l1 on a future patch.

encoder/vaapiencoder_vp9.cpp
xuguangxin: Hi @chivakker 
Any update on this

encoder/vaapiencoder_vp9.cpp
xuguangxin: Any update on this?

encoder/vaapiencoder_vp9.h
xuguangxin, line 45: Any update on this?

encoder/vaapiencoder_vp9.cpp
xuguangxin, line 203: It's Ok for me,

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
+/*
+ * Copyright (C) 2014 Intel Corporation. All rights reserved.
```
lizhong1008: Should be 2016?

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+#include "vaapicodedbuffer.h"
+#include "vaapiencpicture.h"
+#include "vaapiencoder_factory.h"
+#include <algorithm>
+
+namespace YamiMediaCodec {
+
+// each frame can have at most 3 reference frames, golden, alt and last
+enum maxSizeValues {
+    kMaxReferenceFrames = 8,
```
lizhong1008, line 35: "each frame can have at most 3 reference frames, golden, alt and last", so why kMaxReferenceFrames is 8, not 3?

lizhong1008, line 35: Another question is what does the prefix "k" of these elements(such as kMaxReferenceFrames) mean? How about using vp9 instead?

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+    kMaxWidth = 8192,
+    kMaxHeight = kMaxWidth
+};
+
+enum VP9QPValues { kMinQPValue = 9, kDefaultQPValue = 60, kMaxQPValue = 127 };
+
+enum frameType { kKeyFrame, kInterFrame };
```
lizhong1008: Would better use Vp9FrameType since all codec using YamiMediaCodec name space.

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+enum levelValues { kDefaultSharpnessLevel, kDefaultFilterLevel = 10 };
```
lizhong1008: same as frameType

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+    : m_frameCount(0)
+{
+    m_videoParamCommon.profile = VAProfileVP9Profile0;
+    m_videoParamCommon.rcParams.minQP = kMinQPValue;
+    m_videoParamCommon.rcParams.maxQP = kMaxQPValue;
+    m_videoParamCommon.rcParams.initQP = kDefaultQPValue;
+
+    //add extra surfaces to operate due to kMaxReferenceFrames
+    //vaapi_encoder class will create 5 extra surfaces already
+    m_maxOutputBuffer = kMaxReferenceFrames;
```
lizhong1008, line 69: Now I know why you set kMaxReferenceFrames to 8, not 3. but we can  let m_maxOutputBuffer = kMaxReferenceFrames + MaxOutputBuffer, it should be more clear.

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+}
+
+void VaapiEncoderVP9::resetParams()
+{
+    m_maxCodedbufSize = width() * height() * 3 / 2;
+
+    // adding extra padding.  In particular small resolutions require more
+    // space depending on other quantization parameters during execution. The
+    // value below is a good compromise
+    m_maxCodedbufSize += 0x1000;
```
lizhong1008: Just like you commented in vp8 encoder, we can define a MACRO, just like 07bba09ade99fb4cc4a01677d3ed35a768cfb8b1.

lizhong1008: @xuguangxin , I noted you changed the VP8_HEADER_MAX_SIZE from 0x1000 to 0x4000 in the commit https://github.com/01org/libyami/commit/07bba09ade99fb4cc4a01677d3ed35a768cfb8b1. Coudld you tell me why?

encoder/vaapiencoder_vp9.cpp
lizhong1008: Yes, it should be the max resolution in a GOP (key frame period).

lizhong1008: @chivakker , as comment in libva, "The source and recon surfaces allocated should be 64x64(SB) aligned",  have you tested resolution is not 64x64 aligned yuv input?

encoder/vaapiencoder_vp9.cpp
lizhong1008: Yes, would better use VA_INVALID_SURFACE, we are using it in vp8/h264/h265 encoders and decoders.

encoder/vaapiencoder_vp9.cpp
lizhong1008: if we use ref_frame_ctrl_l1 , do we need to reorder the input frames， just like h264 encoder?

encoder/vaapiencoder_vp9.h
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (C) 2014 Intel Corporation. All rights reserved.
```
lizhong1008: should be 2016

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,323 @@
...
+                                          const SurfacePtr& recon)
+{
+
+    if (pic->m_type == VAAPI_PICTURE_I) {
+        m_reference.clear();
+        m_reference.insert(m_reference.end(), kMaxReferenceFrames, recon);
+    }
+    else {
+        m_reference.pop_front();
+        m_reference.push_front(recon);
```
lizhong1008, line 286: Seems for non-key frame, there are always only one reference frame (last frame) in the reference list. But we set picParam->ref_flags.bits.ref_frame_ctrl_l0 = 0x7, it means we always use 3 reference frame. I am not sure it is correct, @chivakker , could you help to clarify?

encoder/vaapiencoder_vp9.h
```diff
@@ -0,0 +1,79 @@
...
+    bool ensureQMatrix(const PicturePtr&);
+    bool referenceListUpdate(const PicturePtr&, const SurfacePtr&);
+
+    void resetParams();
+
+    int keyFramePeriod()
+    {
+        return m_videoParamCommon.intraPeriod ? m_videoParamCommon.intraPeriod
+                                              : 1;
+    }
```
lizhong1008, line 67: We don't need to calculate keyFramePeriod() every time, we just need to set m_videoParamCommon.intraPeriod = 1 if it is zero once. Just like resetParams() in h264 encoder.

encoder/vaapiencoder_vp9.cpp
chivakker: I am gonna keep it to 0xFF for the length of reference frame type.

encoder/vaapiencoder_vp9.cpp
chivakker: It is a naming convention for enum values.

chivakker: on the reference frames, we will initially use only 3 of them.  Future patches should expand the use of all reference frames.

encoder/vaapiencoder_vp9.cpp
chivakker: agree, I'm changing it

encoder/vaapiencoder_vp9.cpp
chivakker: agree

encoder/vaapiencoder_vp9.cpp
chivakker: This can be addressed on a separate because the base class is also adding extra surfaces, I think it will be better that each component tells the real maxOutputBuffer.

encoder/vaapiencoder_vp9.h
chivakker: yes

encoder/vaapiencoder_vp9.cpp
chivakker: The encoder should expose different schemes to assign reference frames.  The initial implementation will use last frame previous frame and both key and alt ref frames will be the last key frame.  

Other schemes will be added later.

encoder/vaapiencoder_vp9.h
chivakker: This is called once every frame is encoded and yes it can be known ahead of time.  This is the same on vp8.  I'll send a separate fix for this.

encoder/vaapiencoder_vp9.h
chivakker: I applied clang format on all the new files, it did find places needed to modify, this wasn't one of those.

encoder/vaapiencoder_vp9.cpp
lizhong1008: “both key and alt ref frames will be the last key frame”, but seems no last key frame in the reference list, since you pop_front every time. There are always only one reference frame (last frame) in the reference list.

encoder/vaapiencoder_vp9.h
lizhong1008: That's fine, we can fix it with vp8.

encoder/vaapiencoder_vp9.cpp
lizhong1008: @chivakker , any update with non-64x64-aligned case?

chivakker: yes, resolution tested is 1080p

encoder/vaapiencoder_vp9.cpp
chivakker: yes and when the next key frame is encoded the recon will be added to all the reference frames, so replacing the first slot on the reference array will not modify the rest in the array and those are used for gold and alt

xuguangxin: No, we push 8 frame to reference list when we have I frame, so it always have 8 frame in list. 
Hi @chivakker ,
This means for first 5 P frames, you always use the I frame as key, alt, and last as reference frame. It is not right behaviors.

encoder/vaapiencoder_vp9.cpp
```diff
@@ -0,0 +1,322 @@
...
+// Fills in VA picture parameter buffer
+bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
+                           const PicturePtr& picture,
+                           const SurfacePtr& surface) const
+{
+    picParam->reconstructed_frame = surface->getID();
+    picParam->coded_buf = picture->getCodedBufferID();
+
+    if (picture->m_type == VAAPI_PICTURE_I) {
+        for (uint32_t i = 0; i < kMaxReferenceFrames ; i++)
```
xuguangxin, line 190: We need use N_ELEMENT(picParam->reference_frames)  instead of kMaxReferenceFrames, this totally decide by libva, we can't assume it's 8.

encoder/vaapiencoder_vp9.cpp
chivakker: Why is not the right behavior?  As explained multiple times the current scheme that is supported will use last key frame as gold and alt ref while previous frame (inter frame) will be the last reference.  There's no use of 5 inter frames as reference or anything like that.

encoder/vaapiencoder_vp9.cpp
chivakker: Spec and then libva API use 8.  I think it is a stretch to think that they will change it.

encoder/vaapiencoder_vp9.cpp
xuguangxin: Hi @chivakker 
The code is not match you descriptions.
For first P frame, the reference list is
KKKKKKKK
You use first 3 reference, it will be KKK
For second P frame, the reference list is 
KKKKKKKP,
You still use first 3 reference, it will be KKK too.

encoder/vaapiencoder_vp9.cpp
lizhong1008: @chivakker , 1080p will be created as 1920x1088 surface, it is 64x64 aligned. Have you tried a 352x288 video? it is 16 aligned, but not 64 aligned.

encoder/vaapiencoder_vp9.cpp
chivakker: It is like this:

First key frame gets copied to reference list KKKKKKKK

For first Inter Frame (P frame in your comment) it will use the above and then update reference frames as

PKKKKKKK

Second inter frame (a.k.a. P frame)

PKKKKKKK

So it is updated at the front of the reference frames.

A run of this will look like:
...
[27117.381181][ctx 0x02000001]  reconstructed_frame = 0x04000005
[27117.381182][ctx 0x02000001]  reference_frames[0] = 0xffffffff
[27117.381183][ctx 0x02000001]  reference_frames[1] = 0xffffffff
[27117.381184][ctx 0x02000001]  reference_frames[2] = 0xffffffff
[27117.381185][ctx 0x02000001]  reference_frames[3] = 0xffffffff
[27117.381185][ctx 0x02000001]  reference_frames[4] = 0xffffffff
[27117.381186][ctx 0x02000001]  reference_frames[5] = 0xffffffff
[27117.381187][ctx 0x02000001]  reference_frames[6] = 0xffffffff
[27117.381187][ctx 0x02000001]  reference_frames[7] = 0xffffffff
...
[27117.396715][ctx 0x02000001]  reconstructed_frame = 0x04000006
[27117.396716][ctx 0x02000001]  reference_frames[0] = 0x04000005
[27117.396717][ctx 0x02000001]  reference_frames[1] = 0x04000005
[27117.396718][ctx 0x02000001]  reference_frames[2] = 0x04000005
[27117.396718][ctx 0x02000001]  reference_frames[3] = 0x04000005
[27117.396719][ctx 0x02000001]  reference_frames[4] = 0x04000005
[27117.396720][ctx 0x02000001]  reference_frames[5] = 0x04000005
[27117.396721][ctx 0x02000001]  reference_frames[6] = 0x04000005
[27117.396721][ctx 0x02000001]  reference_frames[7] = 0x04000005
...
[27117.411372][ctx 0x02000001]  reconstructed_frame = 0x04000007
[27117.411372][ctx 0x02000001]  reference_frames[0] = 0x04000006
[27117.411373][ctx 0x02000001]  reference_frames[1] = 0x04000005
[27117.411374][ctx 0x02000001]  reference_frames[2] = 0x04000005
[27117.411375][ctx 0x02000001]  reference_frames[3] = 0x04000005
[27117.411375][ctx 0x02000001]  reference_frames[4] = 0x04000005
[27117.411376][ctx 0x02000001]  reference_frames[5] = 0x04000005
[27117.411377][ctx 0x02000001]  reference_frames[6] = 0x04000005
[27117.411378][ctx 0x02000001]  reference_frames[7] = 0x04000005
...
[27117.424727][ctx 0x02000001]  reconstructed_frame = 0x04000008
[27117.424728][ctx 0x02000001]  reference_frames[0] = 0x04000007
[27117.424728][ctx 0x02000001]  reference_frames[1] = 0x04000005
[27117.424729][ctx 0x02000001]  reference_frames[2] = 0x04000005
[27117.424730][ctx 0x02000001]  reference_frames[3] = 0x04000005
[27117.424731][ctx 0x02000001]  reference_frames[4] = 0x04000005
[27117.424731][ctx 0x02000001]  reference_frames[5] = 0x04000005
[27117.424732][ctx 0x02000001]  reference_frames[6] = 0x04000005
[27117.424733][ctx 0x02000001]  reference_frames[7] = 0x04000005
....

xuguangxin:
+1

lizhong1008:
+1

wudping:
verify + 1

lizhong1008:
merged as 7423a9752ff0f140b956618c84e26bd48d33b853

PR #536, opened 2016-08-18
v4l2decode: add wayland display to v4l2 decoder

YuJiankang:
support wayland display for v4l2 decoder
Signed-off-by: jkyu jiankang.yu@intel.com

### DIFF
--- a/v4l2/v4l2_codecbase.cpp
+++ b/v4l2/v4l2_codecbase.cpp
@@ -37,6 +37,8 @@
 #include <va/va_android.h>
 #include <ufo/gralloc.h>
 #include <ufo/graphics.h>
+#elif __ENABLE_WAYLAND__
+#include <va/va_wayland.h>
 #endif
 
 #if ANDROID
@@ -83,8 +85,8 @@ V4l2CodecBase::V4l2CodecBase()
     , m_started(false)
 #if ANDROID
     , m_reqBuffCnt(0)
-#elif __ENABLE_X11__
-    , m_x11Display(NULL)
+#elif (__ENABLE_WAYLAND__ || __ENABLE_X11__)
+    , m_Display(NULL)
 #else
     , m_drmfd(0)
 #endif
@@ -335,7 +337,7 @@ int32_t V4l2CodecBase::ioctl(int command, void* arg)
             ASSERT(ret != -1);
             if (port == INPUT) {
                 DEBUG("start decoding/encoding");
-            #ifdef ANDROID
+            #if (defined(ANDROID) || defined(__ENABLE_WAYLAND__))
                 // FIXME, I remember cros flush uses STREAMON/STREAMOFF as well
                 if (!setVaDisplay()) {
                     ERROR("fail to set up VADisplay");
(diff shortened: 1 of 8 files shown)


v4l2/v4l2_codecbase.cpp
```diff
@@ -731,4 +738,60 @@ bool V4l2CodecBase::mapVideoFrames(int32_t width, int32_t height)
...
+    return true;
+}
+
+inline bool V4l2CodecBase::createVpp()
+{
+    NativeDisplay nativeDisplay;
+    nativeDisplay.type = NATIVE_DISPLAY_VA;
+    nativeDisplay.handle = (intptr_t)m_vaDisplay;
+    m_vpp.reset(createVideoPostProcess(YAMI_VPP_SCALER), releaseVideoPostProcess);
+    return m_vpp->setNativeDisplay(nativeDisplay) == YAMI_SUCCESS;
```
halleyzhao, line 756: does the m_vpp take effect?

v4l2/v4l2_codecbase.cpp
```diff
@@ -83,6 +85,9 @@ V4l2CodecBase::V4l2CodecBase()
     , m_started(false)
 #if ANDROID
     , m_reqBuffCnt(0)
+
+#elif __ENABLE_WAYLAND__
+    , m_wlDisplay(NULL)
```
halleyzhao: usually there is a native display except android, so we can use a void\* to avoid too many #if

seanvk: Hi, can we use English in the patch comments please?  Thanks!

v4l2/v4l2_codecbase.cpp
xuguangxin, line 756: Yeah, base class already.

v4l2/v4l2_wrapper.cpp
```diff
@@ -149,6 +149,18 @@ int32_t YamiV4L2_Munmap(void* addr, size_t length)
...
 
 #if ANDROID
+#elif __ENABLE_WAYLAND__
+int32_t YamiV4L2_SetWaylandDisplay(int32_t fd, struct wl_display *wlDisplay)
+{
+    V4l2CodecPtr v4l2Codec = _findCodecFromFd(fd);
+     bool ret = true;
+
+     ASSERT(v4l2Codec);
+     DEBUG("x11display: %p", wlDisplay);
```
halleyzhao: typo for x11display

v4l2/v4l2_codecbase.cpp
halleyzhao, line 756: is it possible to create wl_buffer from the decoder vaSurface directly?
the effort should be less than it on android.

xuguangxin, line 756: Discussed this with jiankang before, It may need more time for this. How about look this after this patch merged.

v4l2/v4l2_wrapper.cpp
YuJiankang: yes, I will fix this

halleyzhao:
+1

xuguangxin:
merged as 2d4a5362ddefd2a71929c3fce28b4af35c2be356

PR #537, opened 2016-08-19
Fix jpeg conformance.

xuguangxin:
Previous code will convert all output to i420, using software, it assume all decoder output is NV12. However, this not true for jpeg, jpeg will output many formats like IMC3, 422H, 422V, 444P etc. This patch and related patch in libyami-utils will skip the conversion if user do not assign one. So we can compare the yuv with software decoder.

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -35,11 +35,13 @@ namespace YamiMediaCodec{
 
 uint32_t guessFourcc(const char* fileName)
 {
-    static const char *possibleFourcc[] = {
-            "I420", "NV12", "YV12",
-            "YUY2", "UYVY",
-            "RGBX", "BGRX", "XRGB", "XBGR"
-        };
+    static const char* possibleFourcc[] = {
+        "I420", "NV12", "YV12",
+        "YUY2", "UYVY",
+        "RGBX", "BGRX", "XRGB", "XBGR",
+        //for jpeg
+        "444P", "422V", "422H", "IMC3"
+    };
     const char* extension = strrchr(fileName, '.');
     if (extension) {
         extension++;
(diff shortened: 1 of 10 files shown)


common/utils_unittest.cpp
```diff
@@ -111,33 +112,48 @@ UTILS_TEST(guessResolutionBasic) {
...
+UTILS_TEST(getPlaneResolution)
+{
+
+    uint32_t width = 1920;
+    uint32_t height = 1080;
+    uint32_t planes;
+    uint32_t w[3];
+    uint32_t h[3];
+    EXPECT_FALSE(getPlaneResolution(0, 1920, 1080, w, h, planes));
+    EXPECT_EQ(planes, 0);
```
uartie: This breaks compilation for sign comparison warning.  Also, the expected value should be the first parameter (e.g. `EXPECT_EQ(0u, planes)`)

common/utils_unittest.cpp
```diff
@@ -111,33 +112,48 @@ UTILS_TEST(guessResolutionBasic) {
...
+    for (size_t i = 0; i < N_ELEMENTS(bppEntrys); i++) {
+        const BppEntry& e = bppEntrys[i];
+        EXPECT_TRUE(getPlaneResolution(e.fourcc, width, height, w, h, planes));
+        EXPECT_EQ(planes, e.planes);
```
uartie: The expected value should be the first parameter.

common/utils_unittest.cpp
```diff
@@ -111,33 +112,48 @@ UTILS_TEST(guessResolutionBasic) {
...
+
+        float bpp;
+        uint32_t bytes = 0;
+        for (uint32_t p = 0; p < planes; p++) {
+            bytes += w[p] * h[p];
+        }
+        bpp = (float)bytes / (width * height);
+        EXPECT_FLOAT_EQ(bpp, e.bpp);
```
uartie: The expected value should be the first parameter.

common/utils.cpp
```diff
@@ -121,52 +123,65 @@ bool guessResolution(const char* filename, int& w, int& h)
     return w && h;
 }
 
+struct RoslutionEntry {
```
uartie: Spelling: `RoslutionEntry` to `ResolutionEntry`

common/utils.cpp
```diff
@@ -121,52 +123,65 @@ bool guessResolution(const char* filename, int& w, int& h)
...
+    uint32_t fourcc;
+    uint32_t planes;
+    //multiple to half width
+    //if it equals 1, you need divide width with 2
+    //if it equals 4, you need wultiple width with 2
+    uint32_t widthMultiple[3];
+    uint32_t heightMultiple[3];
+};
+
+const static RoslutionEntry resolutionEntrys[] = {
```
uartie: Spelling: `resolutionEntrys` to `resolutionEntries`

common/utils.cpp
```diff
@@ -121,52 +123,65 @@ bool guessResolution(const char* filename, int& w, int& h)
     return w && h;
 }
 
+struct RoslutionEntry {
+    uint32_t fourcc;
+    uint32_t planes;
+    //multiple to half width
+    //if it equals 1, you need divide width with 2
+    //if it equals 4, you need wultiple width with 2
```
uartie: Spelling: `wultiple` to `multiply`

common/utils.cpp
```diff
@@ -121,52 +123,65 @@ bool guessResolution(const char* filename, int& w, int& h)
...
+    { VA_FOURCC_444P, 3, { 2, 2, 2 }, { 2, 2, 2 } },
+    { VA_FOURCC_YUY2, 1, { 4 }, { 2 } },
+    { VA_FOURCC_UYVY, 1, { 4 }, { 2 } },
+    { VA_FOURCC_RGBX, 1, { 8 }, { 2 } },
+    { VA_FOURCC_RGBA, 1, { 8 }, { 2 } },
+    { VA_FOURCC_BGRX, 1, { 8 }, { 2 } },
+    { VA_FOURCC_BGRA, 1, { 8 }, { 2 } },
+};
+
+static void getPlaneWidth(uint32_t w, uint32_t plane, const uint32_t multiple[3], uint32_t width[3])
```
uartie: can this be renamed to something more generic since you use it for both width and height below?

common/utils_unittest.cpp
```diff
@@ -111,33 +112,48 @@ UTILS_TEST(guessResolutionBasic) {
...
+
+UTILS_TEST(getPlaneResolution)
+{
+
+    uint32_t width = 1920;
+    uint32_t height = 1080;
+    uint32_t planes;
+    uint32_t w[3];
+    uint32_t h[3];
+    EXPECT_FALSE(getPlaneResolution(0, 1920, 1080, w, h, planes));
```
uartie: What about testing `VA_FOURCC_NV12` format?

Also, we should `EXPECT_FALSE` on some other valid `VA_FOURCC_*` values that are currently unsupported.  (e.g. `VA_FOURCC_Y800`, `VA_FOURCC_ARGB`, `VA_FOURCC_ABGR`, etc.).

decoder/vaapiDecoderJPEG.cpp
```diff
@@ -390,6 +425,10 @@ YamiStatus VaapiDecoderJPEG::start(VideoConfigBuffer* buffer)
     m_configBuffer.height = frame->imageHeight;
     m_configBuffer.surfaceWidth = frame->imageWidth;
     m_configBuffer.surfaceHeight = frame->imageHeight;
+    m_configBuffer.fourcc = getFourcc(frame);
+    if (!m_configBuffer.fourcc) {
+        return YAMI_UNSUPPORTED;
+    }
```
uartie, line 438: Can you add a unittest to test various fourcc jpeg images?

decoder/vaapiDecoderJPEG.cpp
```diff
@@ -358,6 +360,39 @@ YamiStatus VaapiDecoderJPEG::decode(VideoDecodeBuffer* buffer)
...
+        if (v1 == v2)
+            return YAMI_FOURCC_444P;
+        if (v1 == 2 * v2)
+            return YAMI_FOURCC_422V;
+    }
+    else if (h1 == 2 * h2) {
+        if (v1 == v2)
+            return YAMI_FOURCC_422H;
+        if (v1 == 2 * v2)
+            return YAMI_FOURCC_IMC3;
```
uartie: A unique `DEBUG` message before each return value to indicate which fourcc was detected might be nice here ;)

interface/VideoDecoderDefs.h
```diff
@@ -83,6 +84,7 @@ typedef struct {
     int32_t bitrate;
     int32_t framerateNom;
     int32_t framerateDenom;
+    uint32_t fourcc;
```
uartie, line 87: Do we need to increment the libyami version with this interface data structure change?

common/utils_unittest.cpp
```diff
@@ -112,25 +112,6 @@ UTILS_TEST(guessResolutionBasic) {
...
-
-    sstream << std::numeric_limits<long>::max()
-            << "x"
-            << std::numeric_limits<long>::max();
-
-    EXPECT_FALSE(guessResolution(sstream.str().c_str(), w, h));
-    EXPECT_EQ(w, 0);
-    EXPECT_EQ(h, 0);
-}
-
```
uartie, line 132: I disagree, we should keep this test and fix `guessResolution`.  The `guessResolution` function can take any arbitrary `char *` string that is parsed unconditionally.  If the user specifies a filename that includes dimensions larger than what an `int` can hold, then an overflow occurs.  I agree that in the real world the video resolution won't be that large, but we have no control over how the user names their files (i.e. it could be a misnamed filename).  The `guessResolution` function should validate and handle this case more gracefully.  That is, it should return `false` if it overflows.  Otherwise, as it is now, statements like `if (guessResolution(...))` will pass erroneously for overflows.

common/utils_unittest.cpp
xuguangxin: Ok,  I will fix it.

common/utils_unittest.cpp
xuguangxin: I will fix it.

common/utils_unittest.cpp
xuguangxin: I will fix it.

common/utils.cpp
xuguangxin: I will fix it.

common/utils.cpp
xuguangxin: Sorry for this, I should use a dictionary for this.

common/utils.cpp
xuguangxin: I will fix

common/utils.cpp
xuguangxin: Think about this before, but I do not find a better name, Do you have suggestion?

common/utils_unittest.cpp
xuguangxin: ok

decoder/vaapiDecoderJPEG.cpp
xuguangxin: ok

decoder/vaapiDecoderJPEG.cpp
xuguangxin: ok

interface/VideoDecoderDefs.h
xuguangxin: Bump version is boring and maybe conflict with other commits. Since we do not know which commit will merge first.
Let us focus on code first. I will bump up the version when this merged.

xuguangxin:
Hi @uartie 
I have related code in https://github.com/01org/libyami-utils/pull/18/files. Please help to review it if you have time.
thanks

common/utils_unittest.cpp
xuguangxin: Ok, Thanks for explain.  I will remove this patch.

common/utils.cpp
uartie: Perhaps `getPlaneSize`?

uartie: ...or `getPlaneDimension` ?

xuguangxin: I will use getPlaneLength, since Dimension may means how many dimension.

xuguangxin:
Hi @uartie ,
All things fixed, except unit test. I will sent a separate pull request after this merged.

uartie:
lgtm +1

YuJiankang:
+1

xuguangxin:
@wudping , please help do a full round regression test base on this and https://github.com/01org/libyami-utils/pull/18
thanks.

wudping:
@xuguangxin , okay, you are welcome.

wudping:
verify + 1

lizhong1008:
merged and closed.

PR #541, opened 2016-08-24
h264enc: add simulcast and svc-t suppport

lizhong1008:
1. Support CQP mode of h264 temporal scalability (svc-t) encoding，
   with maximal 4 temporal layers.
2. Only Hierarchical P-pictures structure supported right now,
   which requres GOP should not less than 8, be 2^n, and ipperiod must be
   1.
3. Support simulcast encoding, different priority_id can be set.

### DIFF
--- a/common/common_def.h
+++ b/common/common_def.h
@@ -42,6 +42,10 @@
 #define MAX(a, b)  (((a) > (b)) ? (a) : (b))
 #endif
 
+#define CLIP(a, min, max) do { \
+        a = (MIN(max, MAX(a, min)));      \
+} while (0)
+
 #ifndef N_ELEMENTS
 #define N_ELEMENTS(array) (sizeof(array)/sizeof(array[0]))
 #endif
--- a/encoder/vaapiencoder_h264.cpp
+++ b/encoder/vaapiencoder_h264.cpp
@@ -30,6 +30,7 @@
 #include "vaapiencpicture.h"
 #include "vaapiencoder_factory.h"
 #include <algorithm>
+#include <math.h>
 
 namespace YamiMediaCodec{
 //shortcuts
@@ -47,20 +48,27 @@ using std::vector;
 #define LEVEL51_MAX_MBPS 983040
 #define H264_FRAME_FR 172
 #define H264_MIN_CR 2
+#define H264_NAL_START_CODE 0x000001
 
 #define VAAPI_ENCODER_H264_NAL_REF_IDC_NONE        0
 #define VAAPI_ENCODER_H264_NAL_REF_IDC_LOW         1
 #define VAAPI_ENCODER_H264_NAL_REF_IDC_MEDIUM      2
 #define VAAPI_ENCODER_H264_NAL_REF_IDC_HIGH        3
 
+#define H264_SLICE_TYPE_P 0
+#define H264_SLICE_TYPE_B 1
+#define H264_SLICE_TYPE_I 2
+
 typedef enum {
-  VAAPI_ENCODER_H264_NAL_UNKNOWN     = 0,
-  VAAPI_ENCODER_H264_NAL_NON_IDR     = 1,
-  VAAPI_ENCODER_H264_NAL_IDR         = 5,    /* ref_idc != 0 */
-  VAAPI_ENCODER_H264_NAL_SEI         = 6,    /* ref_idc == 0 */
-  VAAPI_ENCODER_H264_NAL_SPS         = 7,
-  VAAPI_ENCODER_H264_NAL_PPS         = 8
-} GstVaapiEncoderH264NalType;
+    VAAPI_ENCODER_H264_NAL_UNKNOWN = 0,
+    VAAPI_ENCODER_H264_NAL_NON_IDR = 1,
+    VAAPI_ENCODER_H264_NAL_IDR = 5, /* ref_idc != 0 */
+    VAAPI_ENCODER_H264_NAL_SEI = 6, /* ref_idc == 0 */
+    VAAPI_ENCODER_H264_NAL_SPS = 7,
+    VAAPI_ENCODER_H264_NAL_PPS = 8,
+    VAAPI_ENCODER_H264_NAL_PREFIX = 14,
+    VAAPI_ENCODER_H264_NAL_SUBSET_SPS = 15
+} VaapiEncoderH264NalType;
 
 /* Refer to H.264 spec Table A-1 l Level limits */
 struct H264LevelLimits {
@@ -69,6 +77,8 @@ struct H264LevelLimits {
     uint32_t minCR;
 };
 
+#define SCALABILITY_INFO_PAYLOAD_TYPE 24
+
 static const H264LevelLimits LevelLimits[] = {
     {40, 245760, 4},
     {41, 245760, 2},
@@ -76,6 +86,21 @@ static const H264LevelLimits LevelLimits[] = {
     {50, 589824, 2},
     {51, 983040, 2},
 };
+
+#define H264_MIN_TEMPORAL_GOP 8
+#define H264_MAX_TEMPORAL_LAYER_NUM 4
+
+static uint32_t H264TempIds[H264_MAX_TEMPORAL_LAYER_NUM][H264_MIN_TEMPORAL_GOP]
+    = { { 0, 0, 0, 0, 0, 0, 0, 0 },
+        { 0, 1, 0, 1, 0, 1, 0, 1 },
+        { 0, 2, 1, 2, 0, 2, 1, 2 },
+        { 0, 3, 2, 3, 1, 3, 2, 3 } };
+
+static const float H264LayerFps[H264_MAX_TEMPORAL_LAYER_NUM] = { 7.5, 15, 30, 60 };
+
+static inline uint32_t getTemporalId(uint32_t temporalLayerNum, uint32_t frameIndex){
+     return H264TempIds[temporalLayerNum - 1][frameIndex % H264_MIN_TEMPORAL_GOP];
+}
 
 static inline bool
 _poc_greater_than (uint32_t poc1, uint32_t poc2, uint32_t max_poc)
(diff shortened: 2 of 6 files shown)


encoder/vaapiencoder_h264.cpp
```diff
@@ -495,6 +620,19 @@ class VaapiEncStreamHeaderH264
...
+        {
+            ASSERT(m_sps.empty());
+            BitWriter bs;
+            bit_writer_write_sps (&bs, sequence, profile);
+            bit_writer_write_svc_extension (&bs, sequence, profile);
+            bit_writer_write_vui_extension (&bs, sequence, profile);
+            bit_writer_write_addtional_extension (&bs, sequence, profile);
+            bsToHeader(m_sps, bs);
+        }
+    */
```
chivakker, line 627: Are you enabling this function as well?

encoder/vaapiencoder_h264.cpp
```diff
@@ -77,6 +87,29 @@ static const H264LevelLimits LevelLimits[] = {
...
+#define H264_MIN_TEMPORAL_GOP 8
+#define H264_MAX_TEMPORAL_LAYER_NUM 4
+
+static uint32_t TempIds[H264_MAX_TEMPORAL_LAYER_NUM][H264_MIN_TEMPORAL_GOP]
+    = { { 0, 0, 0, 0, 0, 0, 0, 0 },
+        { 0, 1, 0, 1, 0, 1, 0, 1 },
+        { 0, 2, 1, 2, 0, 2, 1, 2 },
+        { 0, 3, 2, 3, 1, 3, 2, 3 } };
+
+static const float LayerFps[H264_MAX_TEMPORAL_LAYER_NUM] = { 7.5, 15, 30, 60 };
```
xuguangxin: We can't assume it base layer is 60 fps.

encoder/vaapiencoder_h264.cpp
```diff
@@ -77,6 +87,29 @@ static const H264LevelLimits LevelLimits[] = {
...
+
+static uint32_t log2(uint32_t num)
```
xuguangxin: why not use built in log2 in math.h

encoder/vaapiencoder_h264.cpp
```diff
@@ -220,10 +253,93 @@ bit_writer_write_trailing_bits(BitWriter *bitwriter)
...
+                     const VAEncSequenceParameterBufferH264* const seq,
+                     uint32_t temporalLayerNum)
+{
+    BitWriter scalabilityInfoWriter;
+    uint32_t i;
+
+    /* Write scalability_info */
+    scalabilityInfoWriter.writeBits(0, 1); // temporal_id_nesting_flag: false
+    scalabilityInfoWriter.writeBits(
+        0, 1); // priority_layer_info_present_flag: false
```
xuguangxin, line 258: little wired, why we need start a new line for this

xuguangxin, line 258: Could we make scalabilityInfoWriter smaller, so we can get rid of this wired line breaking?

encoder/vaapiencoder_h264.cpp
```diff
@@ -220,10 +253,93 @@ bit_writer_write_trailing_bits(BitWriter *bitwriter)
...
+
+    /* rbsp_trailing_bits */
+    bit_writer_write_trailing_bits(&scalabilityInfoWriter);
+
+    uint32_t scalabilityInfoBytes = scalabilityInfoWriter.getCodedBitsCount()
+                                    / 8;
+    uint8_t* scalabilityInfoData = scalabilityInfoWriter.getBitWriterData();
+    ASSERT(scalabilityInfoBytes && scalabilityInfoData);
+
+    bit_writer_write_nal_header(bitwriter, VAAPI_ENCODER_H264_NAL_REF_IDC_NONE,
```
xuguangxin, line 314: Why we need write to scalabilityInfoWriter first then write to bitwriter again? Could we write to bitWriter directly?

encoder/vaapiencoder_h264.cpp
```diff
@@ -854,10 +1033,11 @@ void VaapiEncoderH264::resetParams ()
     if (m_maxRefList0Count >= m_maxOutputBuffer -1)
         m_maxRefList0Count = m_maxOutputBuffer -1;
 
-    m_maxRefFrames =
-        m_maxRefList0Count + m_maxRefList1Count;
+    m_maxRefFrames = m_maxRefList0Count + m_maxRefList1Count;
+    ;
```
xuguangxin: extra ;

common/common_def.h
```diff
@@ -42,6 +42,11 @@
 #define MAX(a, b)  (((a) > (b)) ? (a) : (b))
 #endif
 
+#define CLIP(a, min, max)                                                      \
```
xuguangxin: change to `a = std::min(std::max(a, min), max)`. maybe more readable

encoder/vaapiencoder_h264.cpp
```diff
@@ -1014,6 +1194,10 @@ YamiStatus VaapiEncoderH264::reorder(const SurfacePtr& surface, uint64_t timeSta
     }
 
     picture->m_poc = m_frameIndex * 2;
+    picture->m_priorityId = m_videoParamAVC.priorityId;
+    picture->m_temporalId
+        = TempIds[m_temporalLayerNum - 1][m_frameIndex % H264_MIN_TEMPORAL_GOP];
```
xuguangxin: Code will more cleanly if we hide this detail in an inline function. Like `inline uint32_t getTemporaId(int frameIndex)`.

encoder/vaapiencoder_h264.cpp
```diff
@@ -1142,10 +1326,17 @@ bool  VaapiEncoderH264::pictureReferenceListSet (
 
     for (i = 0; i < m_refList.size(); i++) {
         assert(picture->m_poc != m_refList[i]->m_poc);
-        if (picture->m_poc > m_refList[i]->m_poc)
+        if (picture->m_poc > m_refList[i]->m_poc
```
xuguangxin: Wired logical, could we put `if (picture->m_temporalId >= m_refList[i]->m_temporalId)` to a separate place?

encoder/vaapiencoder_h264.cpp
```diff
@@ -1268,7 +1462,8 @@ bool VaapiEncoderH264::fill(VAEncPictureParameterBufferH264* picParam, const Pic
 
     /* set picture fields */
     picParam->pic_fields.bits.idr_pic_flag = picture->isIdr();
-    picParam->pic_fields.bits.reference_pic_flag = (picture->m_type != VAAPI_PICTURE_B);
+    picParam->pic_fields.bits.reference_pic_flag = picture->m_isReference
+        = (picture->m_type != VAAPI_PICTURE_B);
```
xuguangxin, line 1454: Why not make isRefernce a function, just like isIdr did.

encoder/vaapiencoder_h264.cpp
```diff
@@ -1314,6 +1511,214 @@ bool VaapiEncoderH264::fillReferenceList(VAEncSliceParameterBufferH264* slice) c
...
+                                  bs.getCodedBitsCount())) {
+        ret = false;
+    }
+
+    return ret;
+}
+
+bool VaapiEncoderH264::addPackedSliceHeader(
+    const PicturePtr& picture,
+    const VAEncSliceParameterBufferH264* const sliceParam) const
```
xuguangxin, line 1545: Why we need this? To control reference list?

common/common_def.h
lizhong1008: This header may be used by C file, would better not use C++ functions.

encoder/vaapiencoder_h264.cpp
lizhong1008: When CBR is supported, we will support fps and bitrate set by user.

encoder/vaapiencoder_h264.cpp
lizhong1008: it will return a double type.

common/common_def.h
xuguangxin: oh, sorry, could we use MIN MAX

encoder/vaapiencoder_h264.cpp
lizhong1008: No, we don't know scalabilityInfoBytes at the first.

encoder/vaapiencoder_h264.cpp
xuguangxin: It's no compile problem if you assign a double to int, if you want add 1 for double you can trail ceil which in math.h too.

encoder/vaapiencoder_h264.cpp
lizhong1008: Seem it is not necessary for CQP, I am not sure it is necessary for CBR case, so keep it here right now.

encoder/vaapiencoder_h264.cpp
lizhong1008: it will be removed.

encoder/vaapiencoder_h264.cpp
xuguangxin: Hand writing things are error prone and low performance, if possbile, we had better use standard library if it have.

encoder/vaapiencoder_h264.cpp
xuguangxin: We can keep the position of scalabilityInfoBytes, rewrite it after bit_writer_write_trailing_bits.

encoder/vaapiencoder_h264.cpp
lizhong1008: It is a common logic for bot svt-t and non-svct case. One principle of this patch is make it common for svc-t and non-svct as possible as I can, because svc-t is still a AVC mode.

encoder/vaapiencoder_h264.cpp
lizhong1008: yes, we need to let decoder know reference list need to be ordered. It means ref_pic_list_reordering_flag_l0 and modification_of_pic_nums_idc needs to be set. But libva driver can't handle it, even without such VAAPI interface to pass them to driver.

encoder/vaapiencoder_h264.cpp
xuguangxin: Let me explain my concern in details:
1.  Original code only have two branches if else, this code have three branches if, else if, else. Could we change it to two branches?
2.  If yes, maybe we can change code like this, it maybe more readable

```
if (picture->m_temporalId >= m_refList[i]->m_temporalId) {
    m_refList[i]->m_diffPicNumMinus1 = picture->m_frameNum - m_refList[i]->m_frameNum - 1;
}
if (picture->m_poc > m_refList[i]->m_poc)
    m_refList0.push_back(m_refList[i]);/* set forward reflist: descending order */
else
    m_refList1.push_front(m_refList[i]);/* set backward reflist: ascending order */
```

encoder/vaapiencoder_h264.cpp
xuguangxin: We only have IP frame, is it ok to mark some frame as INVALID_SURFACE_ID, to tell driver not reference it? So we do not need modify the reference list.

encoder/vaapiencoder_h264.cpp
lizhong1008: Maybe you misunderstand this code, my main purpose is we can only use reference picture with m_temporalId <= currecnt_picture_temporalId as reference. m_diffPicNumMinus1 is only used let decoder know how to reorder reference list same as encoder's behavior.

encoder/vaapiencoder_h264.cpp
lizhong1008: Seems hard to do as this with current bitwriter interface, and you know we need bit_writer_write_trailing_bits twice.

encoder/vaapiencoder_h264.cpp
lizhong1008: it is not to let encoder driver know something. We write it to bitstream and let decoder know to how to reordered.

xuguangxin:
Hi @lizhong1008 
any update on this?

lizhong1008:
@xuguangxin , I will update it tomorrow. I am testing this feature with difference input parameters now.

lizhong1008:
@xuguangxin updated, please help to review again, thanks.

xuguangxin:
+1

xuguangxin:
Do not forget bump api version for this when we merge.

lizhong1008:
@wudping ,please help to verify it. Thanks

wudping:
okay, @lizhong1008 , you are welcome.

wudping:
verify + 1

lizhong1008:
merged as ea0b5fd79715d0a154b79319d113cf50e2e9b75e

PR #542, opened 2016-08-24
mpeg2_parser: check for nalunit size before starting bitReader

chivakker:
This is to avoid a potential memory leak when the stream size is not
large enough and function has to return.  bitReader can be created after
such check


xuguangxin:
+1

lizhong1008:
+1

xuguangxin:
merged as 305593ad8eb020e68ff513820493738145573609

Issue #545, opened 2016-08-26
[H264][decoder]decode fail with sps->mui.max_dec_frame_buffering is 0 


popkartyeah:
when sps->mui.max_dec_frame_buffering value is 0 in some streams (such as extreme-plane-pred.h264 in ffmpeg), the yamidecode will break, follow is yamidecode run result:
libva info: VA-API version 0.39.2
libva info: va_getDriverName() returns 0
libva info: Trying to open /usr/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_0_39
libva info: va_openDriver() returns 0
libyami error 16541 (vaapidecsurfacepool.cpp, 51): allocate surface failed (352x288), size = 0
VppInputDecode config failed.
DecodeTest init failed.
I think there was something wrong in vaapidecoder_h264.cpp.

xuguangxin:
This may related to https://github.com/01org/libyami/issues/494.
@lizhong1008 , please help to check it

lizhong1008:
@popkartyeah solved by c3714fa8db5b4b13afc7b963a5c1a7e989556379, please help to verify it.

popkartyeah:
@lizhong1008 yeah, I think it had be fixed.

PR #547, opened 2016-08-31
h264dec: fix surface number calculation error issue

lizhong1008:
Surface number needed by h264 decoder should be maxDecFrameBuffering
(used for reference list) + 1 (used for reconstruct frame).

### DIFF
--- a/decoder/vaapidecoder_h264.cpp
+++ b/decoder/vaapidecoder_h264.cpp
@@ -1605,7 +1605,7 @@ bool VaapiDecoderH264::isDecodeContextChanged(const SharedPtr<SPS>& sps)
     if (m_configBuffer.surfaceWidth < sps->m_width
         || m_configBuffer.surfaceHeight < sps->m_height
         || m_configBuffer.surfaceNumber < (int32_t)maxDecFrameBuffering) {
-        m_configBuffer.surfaceNumber = maxDecFrameBuffering;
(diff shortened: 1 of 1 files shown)


lizhong1008:
it is to fix https://github.com/01org/libyami/issues/545

xuguangxin:
Our default surface allocator will add more extra size to required size. https://github.com/01org/libyami/blob/apache/vaapi/vaapisurfaceallocator.cpp#L48
Why we still need add one for this?

wudping:
agree @xuguangxin

lizhong1008:
@xuguangxin 
1. please check https://github.com/01org/libyami/blob/apache/vaapi/vaapisurfaceallocator.cpp#L41, for video which has only idr frame, maxDecFrameBuffering can be zero because it does not have reference frame, that is why this issue happens.
2. extra_size is used for performance without business of decoder itself, here this one is for decoder reconstruct frame, you can  change extra_size  to 0 and decoder should still can work. 
3. h265 has same behavior: https://github.com/01org/libyami/blob/apache/decoder/vaapidecoder_h265.cpp#L915

xuguangxin:
ok, thanks for explain.

xuguangxin:
+1

lizhong1008:
merged as c3714fa8db5b4b13afc7b963a5c1a7e989556379

PR #553, opened 2016-09-02
bitreader: skip function can skip more bits

yizhouwei:
now skip returns a bool type and can skip more then CACHEBYTES

### DIFF
--- a/codecparsers/bitReader.cpp
+++ b/codecparsers/bitReader.cpp
@@ -109,9 +109,18 @@ uint32_t BitReader::read(uint32_t nbits)
     return 0;
 }
 
-void BitReader::skip(uint32_t nbits)
+bool BitReader::skip(uint32_t nbits)
 {
-    read(nbits);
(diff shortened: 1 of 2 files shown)


codecparsers/bitReader.cpp
```diff
@@ -109,9 +109,18 @@ uint32_t BitReader::read(uint32_t nbits)
     return 0;
 }
 
-void BitReader::skip(uint32_t nbits)
+bool BitReader::skip(uint32_t nbits)
 {
-    read(nbits);
+    uint32_t tmp;
+    while (nbits > 8)
```
xuguangxin, line 115: this maybe better 

```
uint32_t size = sizeof(tmp)<<3;
while (nbits  > size) {
  if (!read(tmp))
     return false;
   nbits -= size;
}
```

xuguangxin:
better submit this with your h265 codec parser, but keep it as standalone commit.

PR #555, opened 2016-09-04
fixes for capi

jsorg71:
These changes make the capi usable.  Since the interface directory is not available outside the project it can not be used.

### DIFF
--- a/capi/Makefile.am
+++ b/capi/Makefile.am
@@ -17,6 +17,7 @@ libyami_capi_ldflags = \
 
 libyami_capi_cppflags = \
 	$(LIBVA_CFLAGS) \
+	-I$(top_srcdir)/interface \
 	$(NULL)
 
 noinst_LTLIBRARIES          = libyami_capi.la
--- a/capi/VideoDecoderCapi.cpp
+++ b/capi/VideoDecoderCapi.cpp
@@ -18,8 +18,8 @@
 #include "config.h"
 #endif
 #include "VideoDecoderCapi.h"
-#include "interface/VideoDecoderInterface.h"
-#include "interface/VideoDecoderHost.h"
+#include "VideoDecoderInterface.h"
+#include "VideoDecoderHost.h"
 
 #include <stdlib.h>
 
--- a/capi/VideoDecoderCapi.h
+++ b/capi/VideoDecoderCapi.h
@@ -20,8 +20,8 @@
 #ifndef __ENABLE_CAPI__
 #define __ENABLE_CAPI__ 1
 #endif
-#include "interface/VideoDecoderDefs.h"
-#include "interface/VideoCommonDefs.h"
+#include "VideoDecoderDefs.h"
+#include "VideoCommonDefs.h"
 
 #ifdef __cplusplus
 extern "C" {
--- a/capi/VideoEncoderCapi.cpp
+++ b/capi/VideoEncoderCapi.cpp
@@ -18,8 +18,8 @@
 #include "config.h"
 #endif
 #include "VideoEncoderCapi.h"
-#include "interface/VideoEncoderInterface.h"
-#include "interface/VideoEncoderHost.h"
+#include "VideoEncoderInterface.h"
+#include "VideoEncoderHost.h"
 
 using namespace YamiMediaCodec;
 
--- a/capi/VideoEncoderCapi.h
+++ b/capi/VideoEncoderCapi.h
@@ -20,7 +20,7 @@
 #ifndef __ENABLE_CAPI__
 #define __ENABLE_CAPI__ 1
 #endif
-#include "interface/VideoEncoderDefs.h"
+#include "VideoEncoderDefs.h"
 
 #ifdef __cplusplus
 extern "C" {
--- a/common/YamiVersion.h.in
+++ b/common/YamiVersion.h.in
@@ -19,6 +19,10 @@
 
 #include <stdint.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /**
  * YAMI_API_MAJOR_VERSION
  * The major version of yami api
@@ -64,5 +68,9 @@
 * return YAMI_API_VERSION;
 */
 void yamiGetApiVersion(uint32_t *apiVersion);
+
+#ifdef __cplusplus
+};
+#endif
 
 #endif /* YAMI_VERSION_H */


chivakker:
Can you describe what is your change on the commit message?

capi/Makefile.am
```diff
@@ -17,6 +17,7 @@ libyami_capi_ldflags = \
 
 libyami_capi_cppflags = \
 	$(LIBVA_CFLAGS) \
+	-I$(top_srcdir)/interface \
```
chivakker, line 20: please check you're doing the same that the rest of the libraries when including the interface headers? Otherwise you'll break standard inclusion way

capi/VideoDecoderCapi.cpp
```diff
@@ -18,8 +18,8 @@
 #include "config.h"
 #endif
 #include "VideoDecoderCapi.h"
-#include "interface/VideoDecoderInterface.h"
-#include "interface/VideoDecoderHost.h"
+#include "VideoDecoderInterface.h"
+#include "VideoDecoderHost.h"
```
chivakker, line 22: same thing here and all includes following this on the other files, keep it standard with encoders, decoders, codecparsers, etc.

capi/Makefile.am
jsorg71: This might be done another way.  I really do not like the directory layout currently.  Things that install in include/libyami come from all over the project directories.  Why not have an include directory in the project so the header use is the same in the project as outside the project.

capi/VideoDecoderCapi.cpp
jsorg71: Since interface not available outside the project, they need to be removed from the headers that get installed.  You can't have installed header file going into /usr/local/include that reference the interface directory.  According the the .pc file, the -I that projects should use is /usr/local/include/libyami or changed if --prefix is used.
If you just create a test program that uses the capi you will see the issues.

capi/VideoDecoderCapi.h
```diff
@@ -20,8 +20,8 @@
 #ifndef __ENABLE_CAPI__
 #define __ENABLE_CAPI__ 1
 #endif
-#include "interface/VideoDecoderDefs.h"
-#include "interface/VideoCommonDefs.h"
+#include "VideoDecoderDefs.h"
+#include "VideoCommonDefs.h"
```
chivakker, line 24: This is the place where needs to be changed only as this is the header on ${PREFIX}/include/libyami

capi/Makefile.am
chivakker: Keep in mind the users of libyami.  Clearly on the Capi header needs change the rest should be included inside libyami in the same way as the other elements of the library as already mentioned.

I know there are different ways to resolve this problem and we choose not to break users first.

capi/VideoDecoderCapi.cpp
chivakker: 👎 in the current form.  @uartie can you please take a look at this?  I remember discussing about this with you in the past.

capi/VideoDecoderCapi.h
jsorg71: Right, this is the fix that we need.  This and the other .h file.
capi/VideoDecoderCapi.h 
capi/VideoEncoderCapi.h 
The other changes are just to get everything to build after this header change.

xuguangxin: Hi @chivakker 
Maybe we should remove capi/xxx.h in yami-utils. yami-utils need only use installed header from libyami.

capi/Makefile.am
xuguangxin: Hi @chivakker , 
@jsorg71 have good point. It's little mess since we install include from diffrent place.
like VideoEncoderCapi.h from capi/  amiVersion.h from common/
Maybe we can group all interface file to libyami project's include/interface  directory and  install to include/libyami. This will not break user.

capi/VideoDecoderCapi.h
xuguangxin: We have installed a copy of VideoDecoderCapi.h to $prefix/include/libyami/ and yami-utils never use it. it use the local one. ;-)

jsorg71: I just looked at libyami-utils, the files VideoDecoderCapi..h and VideoEncoderCapi..h are correct over there :)
Yea, they should be removed.
I see there is another PR for always including capi(which I like), maybe this can be fixed after that is addressed.

xuguangxin: Yeah, other things I should mention. We will create a header file named LibyamiC.h, it will include all capi related file. You only need include this file in your project in future.

capi/VideoDecoderCapi.cpp
uartie: For the public headers, we have to remove `interface` from the include path.  For local library compilation units (.cpp) it doesn't matter so much.  However, it is better to be consistent across the board as it's less confusing.  So I'd prefer we "include" using the same syntax as the headers.  So this change here looks ok to me.  However, I agree with @chivakker that we should fix this across the entire project, too, and not just the capi files.

xuguangxin: Hi @uartie 
How about let this patch merged, so we capi is not breaking. Then we change it in entire project.

chivakker: I would prefer to merge only the external header error with this patch and then create other for the internal includes. Can this be taken care of?

xuguangxin: Ok, but we still need `-I$(top_srcdir)/interface \` in makefile.am. else it not build able, right?

wudping: Hi @xuguangxin , I am okay to figure out the conflict. No problem.

xuguangxin:
Hi @jsorg71 
Do you think it's ok to let dongping handle this patch and keep you in commit's signed-off like 
`Signed-off-by: Zhong Xiaoxia <xiaoxia.zhong@intel.com>`
in this commit 
https://github.com/01org/libyami/commit/f6f1483a62dfd1424483ecb24ec9305ea96c2093

jsorg71:
@xuguangxin np, don't forget about the YamiVersion fix too.  Without that, I can't get the version number from C app.

xuguangxin:
ok, @jsorg71 thanks,

xuguangxin:
close this since #522 already have this commit
thanks @jsorg71

PR #556, opened 2016-09-05
h265parser: add check boundary for every read bits

yizhouwei:
bitreader: add xint16 version for readUe and readSe; modify the skip funtion, now skip can skip more than 32 bits and returns a bool type.

h265parser: dealed with a potential bug for scaling list value

### DIFF
--- a/codecparsers/bitReader.cpp
+++ b/codecparsers/bitReader.cpp
@@ -109,9 +109,18 @@ uint32_t BitReader::read(uint32_t nbits)
     return 0;
 }
 
-void BitReader::skip(uint32_t nbits)
+bool BitReader::skip(uint32_t nbits)
 {
-    read(nbits);
+    uint32_t tmp;
+    uint32_t size = sizeof(unsigned long int);
+    while (nbits > size) {
+        if (!read(tmp, size))
+            return false;
+        nbits -= size;
+    }
+    if (!read(tmp, nbits))
+        return false;
+    return true;
 }
 
 uint32_t BitReader::peek(uint32_t nbits) const
--- a/codecparsers/h265Parser.cpp
+++ b/codecparsers/h265Parser.cpp
@@ -24,6 +24,46 @@
 #include <math.h>
 #include <stddef.h> //offsetof
 
+#define READ(f)                             \
+    do {                                    \
+        if (!br.readT(f)) {                 \
+            ERROR("failed to read %s", #f); \
+            return false;                   \
+        }                                   \
+    } while (0)
+
+#define READ_BITS(f, bits)                              \
+    do {                                                \
+        if (!br.readT(f, bits)) {                       \
+            ERROR("failed to read %d to %s", bits, #f); \
+            return false;                               \
+        }                                               \
+    } while (0)
+
+#define READ_UE(f)                            \
+    do {                                      \
+        if (!br.readUe(f)) {                \
+            ERROR("failed to readUe %s", #f); \
+            return false;                     \
+        }                                     \
+    } while (0)
+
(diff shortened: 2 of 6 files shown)


codecparsers/bitReader.cpp
```diff
@@ -109,9 +109,18 @@ uint32_t BitReader::read(uint32_t nbits)
     return 0;
 }
 
-void BitReader::skip(uint32_t nbits)
+bool BitReader::skip(uint32_t nbits)
 {
-    read(nbits);
+    uint32_t tmp;
+    uint32_t size = sizeof(unsigned long int);
```
xuguangxin, line 115: this shuld be sizeof(tmp) << 3

xuguangxin, line 115: did you do test with jvt clips?

yizhouwei: why should be tmp<<3, sizeof(unsigned long int) is the value of CACHEBYTES it is absolutely right.
Of course, i tested.

codecparsers/h265Parser.cpp
```diff
@@ -609,12 +631,15 @@ bool Parser::scalingListData(ScalingList* dest_scaling_list, NalReader& nr)
                 coefNum = std::min(64, (1 << (4 + (sizeId << 1))));
 
                 if (sizeId > 1) {
-                    dstDcList[matrixId] = nr.readSe();
-                    nextCoef = dstDcList[matrixId] + 8;
+                    int32_t tmp;
```
xuguangxin, line 634: could we rename tmp to scaling_list_dc_coef_minus8  and make assign dstDcList like follow:
dstDcList[matrixId] = scaling_list_dc_coef_minus8 + 8;

xuguangxin, line 634: do we need check range for scaling_list_dc_coef_minus8 ?

codecparsers/h265Parser.cpp
```diff
@@ -883,22 +908,22 @@ bool Parser::parseVps(const NalUnit* nalu)
...
         vps->cprms_present_flag.resize(vps->vps_num_hrd_parameters, 0);
         for (uint32_t i = 0; i < vps->vps_num_hrd_parameters; i++) {
-            vps->hrd_layer_set_idx.push_back(nr.readUe());
+            uint32_t tmp;
+            READ_UE(tmp);
+            vps->hrd_layer_set_idx.push_back(tmp);
             if (i > 0)
-                vps->cprms_present_flag[i] = nr.read(1);
-            hrdParameters(&vps->hrd_parameters, nr, vps->cprms_present_flag[i],
+                READ_BITS(vps->cprms_present_flag[i], 1);
```
xuguangxin, line 915: why this not READ

codecparsers/h265Parser.cpp
yizhouwei: ok, i will change the name. 
I'm not sure whether we need to check the range, because the original code did not check it.

codecparsers/bitReader.cpp
xuguangxin: it's bits not bytes right?

codecparsers/h265Parser.cpp
xuguangxin: yes, we should check the range, maybe original code missed this.

> The value of scaling_list_dc_coef_minus8[ sizeId − 2 ][ matrixId ] shall be in the range of −7 to 247, inclusive.

codecparsers/bitReader.cpp
yizhouwei: yes

PR #557, opened 2016-09-05
h265parser: add check boundary for every read bits.

yizhouwei:
bitreader: skip can now skip more bits and returns a bool type
nalreader: add xint16 version for readSe and readUe
h265parser: solve a potential bug of scaling list value


xuguangxin:
+1
Hi @yizhouwei,
great! really thanks for you help.

xuguangxin:
merged as eeb432c52b0772248f65f5424ab2fabd8776d999

PR #560, opened 2016-09-06
vp9 encoder: enable second reference frame mode

### LINKED ISSUE #554: [vp9 encoder]: implement reference frame scheme with last, gold and alt being last 3 inter frames
Provide the vp9 encoder user with a method to choose which frames are used as reference for encoding.  The scheme should use previous frame as last reference, one before that as gold reference and one before that the alternate reference.  This functionality has to be combined with libyami-utils fix for issue https://github.com/01org/libyami-utils/issues/26

chivakker:
vp9 encoder provides two modes for reference frames

Mode 0: This is using previous frame as last reference and
previous key frame as gold and alternate references.  On this
scheme, the reference_frames array will be cleared by a key frame,
then the key frame will populate all 8 slots and last frame will be
updated on slot 0.

Mode 1: This is using previous frame as last reference, one frame
before that is the gold reference and one more frame before that is the
alternate reference. On this scheme, the reference_frames array will be
cleared by a key frame, then the key frame will populate all 8 slots. Then
the first 3 slots will be used as a circular reference list where last/gold/alt
frames will be kept and will be updated according to the rules given by the
intel driver.

This can be tested with libyami-utils yamiencode patch

Fixes #554

### DIFF
--- a/encoder/vaapiencoder_vp9.cpp
+++ b/encoder/vaapiencoder_vp9.cpp
@@ -112,12 +112,16 @@ YamiStatus VaapiEncoderVP9::stop()
 YamiStatus VaapiEncoderVP9::setParameters(VideoParamConfigType type,
                                           Yami_PTR videoEncParams)
 {
-    YamiStatus status = YAMI_SUCCESS;
-    FUNC_ENTER();
+    YamiStatus status = YAMI_INVALID_PARAM;
+
     if (!videoEncParams)
         return YAMI_INVALID_PARAM;
-
     switch (type) {
+    case VideoParamsTypeVP9: {
+        VideoParamsVP9* vp9Params = (VideoParamsVP9*)videoEncParams;
+        PARAMETER_ASSIGN(m_videoParamsVP9, *vp9Params);
+        status = YAMI_SUCCESS;
+    } break;
     default:
         status = VaapiEncoderBase::setParameters(type, videoEncParams);
         break;
@@ -128,11 +132,22 @@ YamiStatus VaapiEncoderVP9::setParameters(VideoParamConfigType type,
 YamiStatus VaapiEncoderVP9::getParameters(VideoParamConfigType type,
                                           Yami_PTR videoEncParams)
 {
-    FUNC_ENTER();
-    if (!videoEncParams)
-        return YAMI_INVALID_PARAM;
+    YamiStatus status = YAMI_INVALID_PARAM;
 
-    return VaapiEncoderBase::getParameters(type, videoEncParams);
+    if (!videoEncParams)
+        return status;
+    switch (type) {
+    case VideoParamsTypeVP9: {
+        VideoParamsVP9* vp9Params = (VideoParamsVP9*)videoEncParams;
+        PARAMETER_ASSIGN(*vp9Params, m_videoParamsVP9);
+        status = YAMI_SUCCESS;
+    } break;
+    default:
+        status = VaapiEncoderBase::getParameters(type, videoEncParams);
+        break;
+    }
+
+    return status;
 }
 
 YamiStatus VaapiEncoderVP9::doEncode(const SurfacePtr& surface,
@@ -157,6 +172,7 @@ YamiStatus VaapiEncoderVP9::doEncode(const SurfacePtr& surface,
     if (picture->m_type == VAAPI_PICTURE_I) {
         codedBuffer->setFlag(ENCODE_BUFFERFLAG_SYNCFRAME);
     }
+
     ret = encodePicture(picture);
     if (ret != YAMI_SUCCESS) {
         return ret;
@@ -181,7 +197,7 @@ bool VaapiEncoderVP9::fill(VAEncSequenceParameterBufferVP9* seqParam) const
 // Fills in VA picture parameter buffer
 bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
                            const PicturePtr& picture,
-                           const SurfacePtr& surface) const
+                           const SurfacePtr& surface)
 {
     picParam->reconstructed_frame = surface->getID();
     picParam->coded_buf = picture->getCodedBufferID();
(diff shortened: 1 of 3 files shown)


interface/VideoEncoderDefs.h
```diff
@@ -256,6 +256,7 @@ typedef struct VideoParamsCommon {
     uint32_t numRefFrames;
     VideoRateControl rcMode;
     VideoRateControlParams rcParams;
+    uint32_t referenceMode;
```
xuguangxin: Do we need add this to VideoParamsVP9 and make this as enum?

chivakker: It is a good observation, we would need to provide a better way of isolating parameters per codec to expose them.  If it makes sense we can create an issue to fix this for all cases.

xuguangxin: before we fix it all, is it ok to create a VideoParamsVP9 just like VideoParamsAVC? If some one start use  VideoParamsCommon.referenceMode, it's hard to change it to VideoParamsVP9

encoder/vaapiencoder_vp9.cpp
```diff
@@ -191,23 +192,56 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
...
         picParam->ref_flags.bits.ref_frame_ctrl_l0 = 0x7;
 
         // golden and alt are last KeyFrame
         // last is last decoded frame
-        picParam->ref_flags.bits.ref_last_idx = 0;
-        picParam->ref_flags.bits.ref_gf_idx = 1;
-        picParam->ref_flags.bits.ref_arf_idx = 2;
-        picParam->pic_flags.bits.frame_context_idx = 0;
+
+        if (getReferenceMode()) {
```
xuguangxin, line 227: Little complex here, is it ok use a 3 element queue to hold the index?
The init value for the queue Q is 0, 1, 2
For mode 0, you do not need change any thing 
For mode 1, you just need pop back and push it to front, ie 2, 0, 1 for first round.
Benefit for this solution is
The last,gf,arf allways at Q[0], Q[1], Q[2]. And the picParam->refresh_frame_flags always be 1 << Q[0] for any mode.

interface/VideoEncoderDefs.h
chivakker: Yes, let me see what I can do with this given the recent news on the release.

encoder/vaapiencoder_vp9.cpp
chivakker: In my view, this solution is simpler than creating an index array as I am working directly on the reference_frames deque.  m_currentReferenceIndex plays the role of the the index array with only one variable.  I did consider creating the index array and came up with this instead. 

Using the index array can work with the push pop technique as long as the reference frames are all full.  At the beginning the logic won't work and there would be a need to know that the first 3 frames haven't come.  I think this will require extra logic that will make the change look about the same, don't you think?

xuguangxin: You can use a member variable to keep the refresh_frame_flags. the code should like this in referenceListUpdate

```
if (pic->m_type == VAAPI_PICTURE_I) {
          m_reference.clear();
          m_reference.insert(m_reference.end(), kMaxReferenceFrames, recon);
          m_refreshFrameFlags = (1<< kMaxReferenceFrames) – 1;    
          resetIndexArray();
} else {
         updateIndexArray();
         m_reference[m_index[0]] = recon;
         m_refreshFrameFlags = 1 << m_index[0];
}

```

resetIndexArray will set index array to 0, 1, 2
updateIndexArray will do nothing for mode 0 and pop back and push front for mode 1. Only place need care about reference mode  in whole file is updateIndexArray.

xuguangxin: Or you can name else branch as updateReferenceFrame, you can handle all reference update and index update on this function, no matter how many reference mode you added

encoder/vaapiencoder_vp9.cpp
```diff
@@ -191,23 +192,56 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
...
-        picParam->refresh_frame_flags = 0x01; // refresh last frame
+
         picParam->pic_flags.bits.frame_type = kInterFrame;
 
         ReferenceQueue::const_iterator it = m_reference.begin();
 
-        for (uint32_t i = 0; it != m_reference.end(); ++it, i++)
+        for (uint32_t i = 0; it != m_reference.end(); ++it, i++) {
             picParam->reference_frames[i] = (*it)->getID();
+#if __ENABLE_DEBUG__
```
lizhong1008: it is added in log.h, why we need it here?

lizhong1008:
@chivakker , do you we need to keep only one reference mode, for example, we can remove mode 0 and keep mode 1. Seems user don't care about the reference mechanism,they only care about the encoder quality.

encoder/vaapiencoder_vp9.cpp
chivakker: yeah you're right, the #if can be removed.

chivakker:
@lizhong1008, both modes give a different option to do vp9 encoding streams, it is up to the user to pick one or the other.  For our purposes both modes help to quantify quality.

lizhong1008:
@chivakker , seems mode 1 always has no worse encoder quality than mode 0, am I right?  If so, we don't need mode 0.

encoder/vaapiencoder_vp9.cpp
```diff
@@ -280,10 +312,25 @@ bool VaapiEncoderVP9::referenceListUpdate(const PicturePtr& pic,
...
     else {
-        m_reference.pop_front();
-        m_reference.push_front(recon);
+        if (getReferenceMode()) {
+            ReferenceQueue::iterator it
+                = m_reference.begin() + m_currentReferenceIndex;
+
+            (*it) = recon;
+
+#if __ENABLE_DEBUG__
```
lizhong1008, line 340: Sorry I forgot to comment here. It also need to be removed.

chivakker: No, this cannot be removed as we don't want the for cycle to do a no-op. I find printing the reference frame useful in debug mode but if it must be removed then I'll remove the whole thing.

lizhong1008: Got it. That is fine.  Maybe you can use va_trace to check reference list also.

encoder/vaapiencoder_vp9.cpp
```diff
@@ -191,23 +192,54 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
...
+            // set to refresh next slot with current reconstructed surface
+            picParam->refresh_frame_flags = 1 << m_currentReferenceIndex;
+            // assign the references on only 3 slots
+            picParam->ref_flags.bits.ref_arf_idx = m_currentReferenceIndex;
+            picParam->ref_flags.bits.ref_gf_idx = m_currentReferenceIndex + 1;
+            picParam->ref_flags.bits.ref_last_idx = m_currentReferenceIndex - 1;
+
+            if (m_currentReferenceIndex == 0) {
+                picParam->ref_flags.bits.ref_last_idx
+                    = m_currentReferenceIndex + 2;
```
lizhong1008, line 245: Here m_currentReferenceIndex  is 0,  picParam->ref_flags.bits.ref_last_idx is 2, no need to set it to be m_currentReferenceIndex + 2.

encoder/vaapiencoder_vp9.cpp
```diff
@@ -191,23 +192,54 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
...
+            }
+            else if (m_currentReferenceIndex == 2) {
+                picParam->ref_flags.bits.ref_gf_idx
+                    = m_currentReferenceIndex - 2;
```
lizhong1008, line 249: Same as L229, better set picParam->ref_flags.bits.ref_gf_idx = 0.

xuguangxin: Hi @lizhong1008 
It's minor change, Is it ok to merge this first?

lizhong1008:
@chivakker “You might be right but it is not know for how much on this implementation.
By having the two modes we can quantify it for different resolutions,
bitrate and video content.  Once we know it for sure we can make a
decision.”    Maybe we can do some test for mode 0 and mode 1 right now,  then we can get the data about which mode has better quality, and make the decision right now to avoid reverting yami code and api. Do you think it is ok?

encoder/vaapiencoder_vp9.cpp
chivakker: That is where the referenceIndex has to wrap around.  Same below, they are needed for this logic to work.

xuguangxin:
+1, @lizhong1008 , Please help to review

lizhong1008:
+1

xuguangxin:
merged as f4b465eb94d07165cfe4b0ce536461d17980f74c

seanvk:
I'm glad this is finally merged, but in the future @xuguangxin I would ask that we do not nit pick over changes such as this.  It is unproductive.

Thanks,

Sean

xuguangxin:
@seanvk , no problem. I can adjust this in future.
I do not want introduce difficult to code review. I just want code more maintainable and well designed. There is huge gap between workable and well designed. 
Thanks.

PR #562, opened 2016-09-07
hevcenc: fix B frame can't be decoded bug.

lizhong1008:
1. Set sps_max_num_reorder_pics and vps_max_num_reorder_pics.
   They will impact decoder DPB cached surface numbers.
2. Sort forward reflist as descending order to make current P frame
   can refer nearest P frame, just like H264 is doing.

### DIFF
--- a/encoder/vaapiencoder_hevc.cpp
+++ b/encoder/vaapiencoder_hevc.cpp
@@ -393,7 +393,7 @@ private:
         bit_writer_put_ue(bitwriter, 5);
 
         /*vps_max_num_reorder_pics*/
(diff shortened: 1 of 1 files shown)


xuguangxin:
+1

xuguangxin:
@wudping , Please help to verfiy, 
we should check encode pattern for I only, IP only,  IPB  only clips,

wudping:
okay, @xuguangxin

chivakker:
lgtm :+1:

wudping:
verify +１

xuguangxin:
merged as 56a012928f333c3cc22fd93f0f3fbdb18075b750

PR #568, opened 2016-09-08
unittest: jpegdec, add 422H, 422V, 444P unittest

xuguangxin:
The 422H, 422V, 444P unit test data will trigger driver's bug.
We will enable it when https://bugs.freedesktop.org/show_bug.cgi?id=97633c fixed


decoder/vaapiDecoderJPEG_unittest.cpp
```diff
@@ -139,24 +403,41 @@ VAAPIDECODER_JPEG_TEST(Factory)
...
+    decodeJpeg(g_SimpleJPEG, YAMI_FOURCC_444P);
+    decodeJpeg(g_JPEG420, YAMI_FOURCC_IMC3);
+    /* this will trigger bug in dirver,
+     * see https://bugs.freedesktop.org/show_bug.cgi?id=97633c for details
+     */
+    /*
+    decodeJpeg(g_JPEG422H, YAMI_FOURCC_422H);
+    decodeJpeg(g_JPEG422V, YAMI_FOURCC_422V);
+    decodeJpeg(g_JPEG444P, YAMI_FOURCC_444P);
+    */
```
uartie, line 440: Please use a [Value-Parameterized Test](https://github.com/google/googletest/blob/master/googletest/docs/V1_7_AdvancedGuide.md#how-to-write-value-parameterized-tests) here instead.

You will probably need to create a different test class for this.  See [vaapiencoder_jpeg_unittest.cpp](https://github.com/01org/libyami/blob/apache/encoder/vaapiencoder_jpeg_unittest.cpp) for an example of a parameterized test.

xuguangxin: ok, but it seems little complex, maybe it can't catch up Q3 release

xuguangxin: Do you think it's ok to check in this first? It do have functionality right?

decoder/vaapiDecoderJPEG_unittest.cpp
```diff
@@ -139,24 +403,41 @@ VAAPIDECODER_JPEG_TEST(Factory)
...
+
+VAAPIDECODER_JPEG_TEST(Decode_Simple)
+{
+    decodeJpeg(g_SimpleJPEG, YAMI_FOURCC_444P);
+    decodeJpeg(g_JPEG420, YAMI_FOURCC_IMC3);
+    /* this will trigger bug in dirver,
+     * see https://bugs.freedesktop.org/show_bug.cgi?id=97633c for details
+     */
+    /*
+    decodeJpeg(g_JPEG422H, YAMI_FOURCC_422H);
```
uartie, line 437: This one passes for me if I uncomment it.

decoder/vaapiDecoderJPEG_unittest.cpp
```diff
@@ -139,24 +403,41 @@ VAAPIDECODER_JPEG_TEST(Factory)
...
+    decodeJpeg(g_JPEG422V, YAMI_FOURCC_422V);
```
uartie, line 438: Instead of generating the JPEG's with ffmpeg (since driver can't handle the output), try using ImageMagick's `convert` program.  It seems to generate images that our driver can handle.  You can convert to different formats with:

`$ convert simple-444P.jpeg -sampling-factor 2x2,2x1,2x1 simple-422V.jpeg`
`$ convert simple-444P.jpeg -sampling-factor 2x2,1x2,1x2 simple-422H.jpeg`

... also you could start with the g_SimpleJPEG image (444P) to generate the others with `convert`.  Finally, I use `okteta` to export the bytes to a c array.

Here's a simple program to write a byte array to a file (in case you need it):

```
#include <fstream>
#include <tr1/array>

const static std::tr1::array<char, 844> bytes = { /* put bytes here and set array size appropriately */ };

int main(int argc, char** argv)
{
    std::ofstream of("writebytes.out", std::ios::out | std::ios::binary);
    of.write(bytes.data(), bytes.size());
    of.close();
    return 0;
}
```

decoder/vaapiDecoderJPEG_unittest.cpp
```diff
@@ -139,24 +403,41 @@ VAAPIDECODER_JPEG_TEST(Factory)
...
+    decodeJpeg(g_JPEG444P, YAMI_FOURCC_444P);
```
uartie, line 439: g_SimpleJPEG is 444P... not sure we need another one.  What is it about this image and the 422V that causes problems?  Is ffmpeg generating bad sampling factors or is it truly a driver bug?

uartie: I can take a stab at it... hope that's ok.

decoder/vaapiDecoderJPEG_unittest.cpp
uartie: If you're ok with it, I sent pull request #576 that does this with a Value-Parameterized Test instead.  It's not too complicated once you get familiar with the Gtest Value-Parameterized Test API.

xuguangxin: You are so kind

decoder/vaapiDecoderJPEG_unittest.cpp
xuguangxin: From $va/test/decode/tinyjpeg.c, It only support u,v  h_sampling_factor 1. It expect 444p's h_sampling_factor[] equals to 1, 1, 1. But ffmpeg's will set 444P's h_sampling_factor[] to 4,4,4.  From my view,ffmpeg is no problem, and it is wildly used, we should support this in driver.

g_SimpleJPEG  is 444P but it's h_sampling_factor[] is 1,1,1. So it's no problem.

```
#ifdef SANITY_CHECK
  if (   (priv->component_infos[cY].Hfactor < priv->component_infos[cCb].Hfactor)
      || (priv->component_infos[cY].Hfactor < priv->component_infos[cCr].Hfactor))
    error("Horizontal sampling factor for Y should be greater than horitontal sampling factor for Cb or Cr\n");
  if (   (priv->component_infos[cY].Vfactor < priv->component_infos[cCb].Vfactor)
      || (priv->component_infos[cY].Vfactor < priv->component_infos[cCr].Vfactor))
    error("Vertical sampling factor for Y should be greater than vertical sampling factor for Cb or Cr\n");
  if (   (priv->component_infos[cCb].Hfactor!=1) 
      || (priv->component_infos[cCr].Hfactor!=1)
      || (priv->component_infos[cCb].Vfactor!=1)
      || (priv->component_infos[cCr].Vfactor!=1))
    printf("ERROR:Sampling other than 1x1 for Cr and Cb is not supported");
#endif
```

decoder/vaapiDecoderJPEG_unittest.cpp
xuguangxin: Thanks, but ffmpeg generated file is important too. We should support it in driver. If it's not hardware limitation

xxd can do file to array dump too

```
echo Hello  > temp
xxd -i temp
```

xuguangxin:
close this since we have #576

decoder/vaapiDecoderJPEG_unittest.cpp
uartie: Cool, I did not know about `xxd`... command line tools are better and faster.  I will be using that for sure ;)

Perhaps we should add the ffmpeg generated data to the driver test suite.

PR #573, opened 2016-09-08
Isolate TR1

### LINKED ISSUE #571: TR1 vs C++11
We need to isolate the usage of tr1 and C++11 in the entire project.  We need a common interface that can be used that transparently uses either C++11 or TR1 APIs depending on compilation mode (i.e. ANDRIOD vs. everything else).  Currently, this is partially done in [VideoCommonDefs.h](https://github.com/01org/libyami/blob/apache/interface/VideoCommonDefs.h) and [Functional.h](https://github.com/01org/libyami/blob/apache/common/Functional.h).  However, there  are still a lot of files that include tr1 headers instead of the aforementioned headers.

Also, the `#define Bind ...` definition conflicts with gtest's definition of Bind... so this will need resolved.  We should try to avoid the `#define`  pattern, too.

@scott-d-phillips suggested that we might put `std::tr1::bind`, etc.. into the `std` namespace, for example:

```
namespace std {
  using std::tr1::bind;
};
```

... then we  could use `std::bind`, etc. throughout the code.

uartie:
Fixes #571


interface/VideoCommonDefs.h
```diff
@@ -21,30 +21,31 @@
...
-
-}
-#endif
+#if __cplusplus > 199711L
+    #include <memory>
+#else
+    #include <tr1/memory>
+    namespace std {
+
+        #if __cplusplus < 201103L
```
d-scott-phillips: This is already in the else arm of __cplusplus > 199711L, meaning we already know __cplusplus <= 199711L. Checking for < 201103L just seems to be confusing.

uartie: This is necessary.   Without it GCC >= 6.x preprocesses it even if this branch is not enabled.  If this check is not here, the compilation fails.  I will add a comment here in the source.

uartie: Hmmm... odd.  I removed them and now it compiles fine on 6.x.  I must have been in an intermediate change state when I encountered problems.  I will remove this check in an update.

interface/VideoCommonDefs.h
```diff
@@ -21,30 +21,27 @@
...
-    #define DynamicPointerCast std::tr1::dynamic_pointer_cast
-    #define StaticPointerCast std::tr1::static_pointer_cast
-    #define EnableSharedFromThis std::tr1::enable_shared_from_this
-#endif
 
-#define SharedFromThis shared_from_this
-
-}
-#endif
+#if __cplusplus > 199711L
```
xuguangxin, line 25: Will this introduce problem, when libyami build with `__cplusplus <= 199711L` and caller build with`__cplusplus > 199711L`?

uartie: Can you be more specific with a use-case when this could happen?

uartie: So on GCC that defaults to `-std=c++98` (GCC 4.9.2), I built libyami with default (i.e. TR1 is used).  Then I set `CXXFLAGS=-std=c++11` and compiled libyami-utils (i.e. TR1 is not used).  Everything seemed to work fine.

Conversely, I did similar on GCC where default is -std=C++11 (GCC 6.1.1)... compiled libyami with `CXXFLAGS=-std=c++98` and libyami-utils with default.  No issues there either.

I ran yamidecode with h264 and mjpeg streams for both cases.

Is this what your concern was?

uartie: also tested similar case on GCC 4.8.4... no issues

xuguangxin: Oh, maybe this becuase std::tr1 and std are same header.... have same memory layout ;). 
Ok, thanks for check this

xuguangxin:
It is really an elegant solution, thanks @uartie and @scott-d-phillips

uartie:
FYI, this will conflict with PR #576 . I will rebase this once #576 merges.

uartie:
rebased

PR #574, opened 2016-09-09
vc1parser: add boundary check for every read bits

yizhouwei:
bitreader: add new version of peek that returns a bool value


xuguangxin:
close this since we have #577

PR #575, opened 2016-09-09
do not use m_videoFormatInfo in getOutput

xuguangxin:
getOuptut maybe call after format change. Suppose you have a clips with two frame in different resolution, named AB.
Imaging following call sequence:
1.User sent A to yami
2.User sent B to yami
3.Yami generate format change to user
4.User sent B to yami again.
5.Then user get the two frames.
Since the m_videoFormatInfo already changed after step 3, we will get frame A with B’s resolution.
This definitely wrong. This pull request  will fix the issue


YuJiankang:
lgtm +1

lizhong1008:
LGTM， +1

xuguangxin:
Please help to do full round test @wudping

xuguangxin:
merged as d9e91bacb1e3baa1e8396db489c8c0a7a564d84e

PR #579, opened 2016-09-12
[second implementation] vp9 encoder: enable second reference frame mode

### LINKED ISSUE #554: [vp9 encoder]: implement reference frame scheme with last, gold and alt being last 3 inter frames

chivakker:
This second implementation is given to compare against the original patch

vp9 encoder provides two modes for reference frames

Mode 0: This is using previous frame as last reference and
previous key frame as gold and alternate references.  On this
scheme, the reference_frames array will be cleared by a key frame,
then the key frame will populate all 8 slots and last frame will be
updated on slot 0.

Mode 1: This is using previous frame as last reference, one frame
before that is the gold reference and one more frame before that is the
alternate reference. On this scheme, the reference_frames array will be
cleared by a key frame, then the key frame will populate all 8 slots. Then
the first 3 slots will be used as a circular reference list where last/gold/alt
frames will be kept and will be updated according to the rules given by the
intel driver.

This can be tested with libyami-utils yamiencode patch

Fixes #554

### DIFF
--- a/encoder/vaapiencoder_vp9.cpp
+++ b/encoder/vaapiencoder_vp9.cpp
@@ -157,6 +157,7 @@ YamiStatus VaapiEncoderVP9::doEncode(const SurfacePtr& surface,
     if (picture->m_type == VAAPI_PICTURE_I) {
         codedBuffer->setFlag(ENCODE_BUFFERFLAG_SYNCFRAME);
     }
+
     ret = encodePicture(picture);
     if (ret != YAMI_SUCCESS) {
         return ret;
@@ -181,7 +182,7 @@ bool VaapiEncoderVP9::fill(VAEncSequenceParameterBufferVP9* seqParam) const
 // Fills in VA picture parameter buffer
 bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
                            const PicturePtr& picture,
-                           const SurfacePtr& surface) const
+                           const SurfacePtr& surface)
 {
     picParam->reconstructed_frame = surface->getID();
     picParam->coded_buf = picture->getCodedBufferID();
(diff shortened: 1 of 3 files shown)


encoder/vaapiencoder_vp9.cpp
```diff
@@ -191,23 +192,46 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
...
+        if (getReferenceMode()) {
+            // intel driver updates a new slot with every new frame and keeps
+            // the reference frames in a circular buffer, the buffer is defined
+            // with 8 slots but it is up to the application to use as many as
+            // desired. This scheme implements the use of 3 slots for
+            // last/gold/alt refs
+
+            picParam->refresh_frame_flags = 1 << m_referenceIndex[1];
+            // assign the references on only 3 slots
+            picParam->ref_flags.bits.ref_arf_idx = m_referenceIndex[1];
```
xuguangxin, line 221: L221-L223 is simliar to L231-L233. Could you merge this up? If we always treat ref_last_idx  as m_referenceIndex[0], ref_gf_idx as m_referenceIndex[1], ref_arf_idx as m_referenceIndex[2].
It should simplify our code. We can use following code in any reference mode

```
picParam->ref_flags.bits.ref_last_idx = m_referenceIndex[0];
picParam->ref_flags.bits.ref_gf_idx = m_referenceIndex[1];
picParam->ref_flags.bits.ref_arf_idx` = m_referenceIndex[2];
```

Moreover, I think the `if (getReferenceMode())`,  can be removed. Base on following reason:
1.  The m_referenceIndex and refresh_frame_flags totally decide by the order of m_reference. Therefore, we can handle it at referenceListUpdate.
2.  This make code more clear, only one place need handle the getReferenceMode is referenceListUpdate
What do you think?

chivakker: On Mon, Sep 12, 2016 at 6:47 PM, Xu Guangxin notifications@github.com
wrote:

> In encoder/vaapiencoder_vp9.cpp
> https://github.com/01org/libyami/pull/579#discussion_r78485031:
> 
> > ```
> >      // last/golden/alt is used as reference frame. L0 forward
> >      picParam->ref_flags.bits.ref_frame_ctrl_l0 = 0x7;
> > 
> >      // golden and alt are last KeyFrame
> >      // last is last decoded frame
> > ```
> > -        picParam->ref_flags.bits.ref_last_idx = 0;
> > -        picParam->ref_flags.bits.ref_gf_idx = 1;
> > -        picParam->ref_flags.bits.ref_arf_idx = 2;
> > -        picParam->pic_flags.bits.frame_context_idx = 0;
> >   +
> > -        if (getReferenceMode()) {
> > -            // intel driver updates a new slot with every new frame and keeps
> > -            // the reference frames in a circular buffer, the buffer is defined
> > -            // with 8 slots but it is up to the application to use as many as
> > -            // desired. This scheme implements the use of 3 slots for
> > -            // last/gold/alt refs
> >   +
> > -            picParam->refresh_frame_flags = 1 << m_referenceIndex[1];
> > -            // assign the references on only 3 slots
> > -            picParam->ref_flags.bits.ref_arf_idx = m_referenceIndex[1];
> 
> L221-L223 is simliar to L231-L233. Could you merge this up? If we always
> treat ref_last_idx as m_referenceIndex[0], ref_gf_idx as
> m_referenceIndex[1], ref_arf_idx as m_referenceIndex[2].
> It should simplify our code. We can use following code in any case
> 
> picParam->ref_flags.bits.ref_last_idx = m_referenceIndex[0];
> picParam->ref_flags.bits.ref_gf_idx = m_referenceIndex[1];
> picParam->ref_flags.bits.ref_arf_idx` = m_referenceIndex[2];
> 
> Not quite,  On this second scheme the first inter frame has to be inserted
> on position 1, the next interframe on position 2 and the next wraps around
> to position 0.  First scheme updates only position 0.
> 
>  Moreover, I think the if (getReferenceMode()), can be removed. Base on
> following reason:
> 1. The m_referenceIndex and refresh_frame_flags totally decide by the
> order of m_reference. Therefore, we can handle it at referenceListUpdate.
> 2. This make code more clear, only one place need handle the
> getReferenceMode is referenceListUpdate
> What do you think?

getReferenceMode is a user command (this has to be fixed in a better way by
overriding setParameters from the encoder_base class).  It helps to pick
the reference scheme and it affects how m_reference and m_referenceIndex
will behave.

xuguangxin:
close this since we have https://github.com/01org/libyami/pull/560
Hi @chivakker 
Thanks for second implmentation.

PR #588, opened 2016-09-14
vpx-encoder: doEncode to use forceKeyFrame

### LINKED ISSUE #583: vpx encoders won't use the force keyframe flag
libyami client can set the flag for keyframe at the interval desired.  Implementation should honor those requests.

chivakker:
doEncode has to use forceKeyFrame to request a key frame to the driver. There's
no need for the libyami component to calculate the next key frame itself.

Application will set the key frame flag on its own.

Fixes #583

### DIFF
--- a/encoder/vaapiencoder_vp9.cpp
+++ b/encoder/vaapiencoder_vp9.cpp
@@ -159,8 +159,11 @@ YamiStatus VaapiEncoderVP9::doEncode(const SurfacePtr& surface,
 
     PicturePtr picture(new VaapiEncPictureVP9(m_context, surface, timeStamp));
 
-    m_frameCount %= keyFramePeriod();
-    picture->m_type = (m_frameCount ? VAAPI_PICTURE_P : VAAPI_PICTURE_I);
+    if (!(m_frameCount % keyFramePeriod()) || forceKeyFrame)
+        picture->m_type = VAAPI_PICTURE_I;
+    else
+        picture->m_type = VAAPI_PICTURE_P;
+
     m_frameCount++;
 
     CodedBufferPtr codedBuffer
--- a/encoder/vaapiencoder_vp8.cpp
+++ b/encoder/vaapiencoder_vp8.cpp
@@ -136,8 +136,11 @@ YamiStatus VaapiEncoderVP8::doEncode(const SurfacePtr& surface, uint64_t timeSta
 
     PicturePtr picture(new VaapiEncPictureVP8(m_context, surface, timeStamp));
 
-    m_frameCount %= keyFramePeriod();
-    picture->m_type = (m_frameCount ? VAAPI_PICTURE_P : VAAPI_PICTURE_I);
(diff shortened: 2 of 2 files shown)


encoder/vaapiencoder_vp9.cpp
```diff
@@ -159,9 +157,7 @@ YamiStatus VaapiEncoderVP9::doEncode(const SurfacePtr& surface,
 
     PicturePtr picture(new VaapiEncPictureVP9(m_context, surface, timeStamp));
 
-    m_frameCount %= keyFramePeriod();
-    picture->m_type = (m_frameCount ? VAAPI_PICTURE_P : VAAPI_PICTURE_I);
-    m_frameCount++;
+    picture->m_type = (forceKeyFrame ? VAAPI_PICTURE_I : VAAPI_PICTURE_P);
```
xuguangxin: Force Key frame usually use by realtime communication program. If there are some frame drop between peers, following frames are not decodeable, until peer see a key frame. So If peer detect a frame dropping, It will sent a notification back to yami’s user,  then the user will sent an forceKeyFrame to yami. Therefore, we’d better have two machnism here.
1.  The normal case, the I frame decide by intraPeriod
2.  If user want force key frame, we will start a new gop.
Please refer to https://github.com/01org/libyami/blob/apache/encoder/vaapiencoder_h264.cpp#L1154
And https://github.com/01org/libyami/blob/apache/encoder/vaapiencoder_h264.cpp#L1279

xuguangxin (changes requested):
You may have misunderstand about forceKeyFrame;)

encoder/vaapiencoder_vp9.cpp
chivakker: In the case of VPX there's no need to reset the GOP.  But it is correct that the intraPeriod has to be handled separately from the forceKeyFrame.  The example provided on libyami-utils will force a keyFrame based on intraPeriod but it has to consider the other scenarios and involve other parameters like kf_min_dist kf_max_dist and kf_auto.  I'll submit a better example for this based also on how the driver is supposed to interpret the kf_\* variables.

xuguangxin: Reset gop or not will depends on the implementation. However, here are some think about this:
Suppose key period is 30, User request key frame at 29. If we do not reset the GOP, it will generate a key frame in 29 and 30, this is unnecessary and will make bitrate very high, at this point.
Base on this we choose to reset GOP at h264 encoder

chivakker:
Re-submitted the patch considering the comments form @xuguangxin.  There's still some room for improvement once kf_\* variables are better understood.  There's no need to change anything on libyami-utils.

xuguangxin (approved):

xuguangxin:
lgtm, thanks

xuguangxin:
merged as 0e9c1051495ff460eb5f5946940a682948f3d5cd

PR #589, opened 2016-09-16
RFC vp9encoder: set frame width and height to 8 bit aligned

### LINKED ISSUE #582: vp9 encode fails with special 2K yuv input
$ yamiencode -i somefile_300frames.yuv -s I420 -W 2048 -H 858 -c VP9 --rcmode CQP -o ./samefile_300frames.yuv.vp9
[IMGUTILS @ 0x7ffd372c02e0] Picture size 0x0 is invalid
libva info: VA-API version 0.39.3
libva info: va_getDriverName() returns 0
libva info: Trying to open /opt/yami//vaapi/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_0_39
libva info: va_openDriver() returns 0
libyami error 14466 (vaapipicture.cpp, 61): vaEndPicture(): invalid parameter
libyami error 14466 (encode.cpp, 126): assert fails
yamiencode: encode.cpp:126: int main(int, char**): Assertion `0 && (status == YAMI_SUCCESS)' failed.
Aborted (core dumped)

chivakker:
When encoding yuv inputs that are not 8 bit aligned, the first
call to vaEndPicture will result in a unrecoverable error.  By
setting the resolution on the VAEncPictureParameterBufferVP9 structure
to be 8 pixel aligned then the driver will create a valid output
stream but with a strip on the bottom and right side when either
resolution was re-aligned.

There's no need to allocate surfaces or reconstructed surface
8 bit aligned to be able to avoid the vaEndPicture error

Both src and dst need to be aligned on this structure otherwise the
vaEndPicture will keep crashing.

Investigation started to try to fix: #582 

### DIFF
--- a/encoder/vaapiencoder_vp9.cpp
+++ b/encoder/vaapiencoder_vp9.cpp
@@ -257,10 +257,10 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
         }
     }
 
-    picParam->frame_width_src = width();
-    picParam->frame_height_src = height();
-    picParam->frame_width_dst = width();
-    picParam->frame_height_dst = height();
+    picParam->frame_width_src = ALIGN8(width());
+    picParam->frame_height_src = ALIGN8(height());
+    picParam->frame_width_dst = ALIGN8(width());
+    picParam->frame_height_dst = ALIGN8(height());
 
     picParam->pic_flags.bits.show_frame = 1;
 


encoder/vaapiencoder_vp9.cpp
```diff
@@ -257,10 +257,10 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
         }
     }
 
-    picParam->frame_width_src = width();
-    picParam->frame_height_src = height();
-    picParam->frame_width_dst = width();
-    picParam->frame_height_dst = height();
+    picParam->frame_width_src = ALIGN8(width());
```
xuguangxin, line 260: Hi @chivakker 
I am little confuse here. What if the source is not 8 pixel align? Will this introduce artificial if the video src is not 8 pixel aligned?

chivakker: yes, there's going to be a green horizontal line on the  bottom of the encoded stream for resolution 2048x858 or a green horizontal line on the bottom and a green vertical line on the right of the encoded stream when resolution is for example 198x150.  I reached to the libva-driver guys to understand the behavior

xuguangxin: Ok, do you think it is ok to pending this patch and waiting for your fix?

lizhong1008: Seems need driver to fix it.  frame_height_dst and frame_width_dst  are used to scale not clip. We can create a bug to driver, let yami return a error then close this issue.

xuguangxin:
close this since we have https://github.com/01org/libyami/pull/601

PR #592, opened 2016-09-18
Merge code from APL IPC

zhaobob:
For apache rebase.


encoder/vaapiencoder_base.cpp
```diff
@@ -89,12 +89,12 @@ YamiStatus VaapiEncoderBase::start(void)
 void VaapiEncoderBase::flush(void)
 {
     AutoLock l(m_lock);
-    m_output.clear();
 }
```
chivakker, line 98: why is this line removed? h.264 overrides the flush function just avoid calling the flush on the base class.

encoder/vaapiencoder_base.cpp
```diff
@@ -89,12 +89,12 @@ YamiStatus VaapiEncoderBase::start(void)
...
 
 YamiStatus VaapiEncoderBase::stop(void)
 {
     FUNC_ENTER();
+    m_output.clear();
     cleanupVA();
```
chivakker, line 104: Then why is it needed here?  h.264 stop is calling flush, so the problem might be that your changes to h.264 flush don't need to call  base class flush

chivakker (changes requested):
Concerns on the encoder base class modifications.

xuguangxin:
Hi @zhaobob and @chivakker 
This change so many things, Do you  think it's ok to discuss and merge it after Q3 release?

encoder/vaapiencoder_base.cpp
lizhong1008: @chivakker , h264 flush function will call VaapiEncoderBase::flush:
void VaapiEncoderH264::flush()
{
    FUNC_ENTER();
    resetGopStart();
    m_reorderFrameList.clear();
    referenceListFree();

```
VaapiEncoderBase::flush();
```

}

encoder/vaapiencoder_base.cpp
lizhong1008: @chivakker , this change it not only for h264 encoder. It is relative how to use stop() and flush() interface defined in VideoEncoderInterface.cpp. The right pipeline is : start()->encode()->flush()->getoutput()->stop(). If we clear m_output, we can not get output data when we call getoutput().

zhaobob:
@xuguangxin , sure. We can use the local copy of upstream version before the merging.

encoder/vaapiencoder_h264.cpp
```diff
@@ -1429,7 +1429,14 @@ bool VaapiEncoderH264::fill(VAEncSequenceParameterBufferH264* seqParam) const
     /* VUI parameters are always set, at least for timing_info (framerate) */
     seqParam->vui_parameters_present_flag = TRUE;
     if (seqParam->vui_parameters_present_flag) {
-        seqParam->vui_fields.bits.aspect_ratio_info_present_flag = FALSE;
+        if (m_videoParamAVC.SAR.AspectRatioIdc != 0) {
```
xuguangxin: To simply logical and make caller's life easy. Could we always set aspect_ratio_idc to  `Extended_SAR`. and use
`if (!SarWidth && !SarHeight)` to check user set Sar or not ?

encoder/vaapiencoder_base.cpp
```diff
@@ -190,8 +190,8 @@ YamiStatus VaapiEncoderBase::setParameters(VideoParamConfigType type, Yami_PTR v
         VideoParamsCommon* common = (VideoParamsCommon*)videoEncParams;
         if (common->size == sizeof(VideoParamsCommon)) {
             PARAMETER_ASSIGN(m_videoParamCommon, *common);
-            if(m_videoParamCommon.rcParams.bitRate > 0)
-	         m_videoParamCommon.rcMode = RATE_CONTROL_CBR;
+            //if(m_videoParamCommon.rcParams.bitRate > 0)
```
xuguangxin: Please do not just comment out this just remove it.
Hi @lizhong1008 
Do you think it's ok?

encoder/vaapiencoder_h264.cpp
```diff
@@ -762,6 +762,8 @@ class VaapiEncPictureH264:public VaapiEncPicture
         out.flag = 0;
 
         std::vector<Function> functions;
+        if (m_enableAUD && format == OUTPUT_EVERYTHING)
```
xuguangxin, line 875: Aud indicate frame end, right? Do we need move this after 770?

encoder/vaapiencoder_h264.cpp
```diff
@@ -1313,6 +1334,11 @@ void VaapiEncoderH264::setIdrFrame (const PicturePtr& pic)
     m_idrNum++;
 }
 
+void VaapiEncoderH264::setAUD (const PicturePtr& pic, bool enable)
```
xuguangxin: better change    `VaapiEncPictureH264(const ContextPtr& context, const SurfacePtr& surface, int64_t timeStamp)` to     `VaapiEncPictureH264(const ContextPtr& context, const SurfacePtr& surface,   int64_t timeStamp, bool appendAUD)`

interface/VideoEncoderDefs.h
```diff
@@ -215,6 +235,7 @@ typedef struct VideoParamsCommon {
     uint32_t numRefFrames;
     VideoRateControl rcMode;
     VideoRateControlParams rcParams;
+    VideoParamsROI roiParams;
```
xuguangxin: This should not belong to VideoParamsCommon, you have VideoParamsTypeROI

interface/VideoEncoderDefs.h
```diff
@@ -166,6 +166,25 @@ typedef struct VideoRateControlParams {
     int8_t diffQPIB;// B frame qp minus initQP
 }VideoRateControlParams;
 
+typedef struct VideoROI {
+    int16_t x;
```
xuguangxin, line 171: could we use VideoRect for x, y, width and height?

encoder/vaapiencoder_base.cpp
```diff
@@ -484,13 +539,14 @@ bool VaapiEncoderBase::initVA()
         return false;
     }
 
+    attrib[attribCount].type = VAConfigAttribEncROI;
```
xuguangxin, line 544: Do we really need set this?VAConfigAttribEncROI is read only. I think the modification in initVA is not needed.

encoder/vaapiencpicture.h
```diff
@@ -44,9 +44,8 @@ class VaapiEncPicture:public VaapiPicture {
     template < class T >
     bool newSlice(T * &sliceParam);
 
-
     template < class T >
-    bool newMisc(VAEncMiscParameterType, T * &miscParam);
+    bool newMisc(VAEncMiscParameterType, T * &miscParam, uint32_t extraSize = 0);
```
xuguangxin, line 48: could we split changes in vaapiencpicture.xxx to a standalone commit?

encoder/vaapiencoder_h264.cpp
```diff
@@ -1516,21 +1627,60 @@ bool VaapiEncoderH264::fill(VAEncPictureParameterBufferH264* picParam, const Pic
...
+    const VideoParamsQM *qMatrix = &m_videoParamAVC.qMatrix;
+    const uint8_t *list_present_flag;
+    int i;
+
+    if (qMatrix->picScalingMatrixPresentFlag) {
+        list_present_flag = qMatrix->picScalingListPresentFlag;
+    } else if (qMatrix->seqScalingMatrixPresentFlag) {
+        list_present_flag = qMatrix->seqScalingListPresentFlag;
+    } else {
+        return true;
```
xuguangxin, line 1752: should add assert here

interface/VideoEncoderDefs.h
```diff
@@ -185,6 +185,16 @@ typedef struct VideoParamsROI {
     VideoROI roi[MAX_ROI_NUM];
 } VideoParamsROI;
 
+typedef struct VideoParamsQM {
+    uint8_t scalingList4x4[6][16];  //in raster scan order
+    uint8_t scalingList8x8[2][64];  //in raster scan order
+
+    uint8_t seqScalingMatrixPresentFlag;
+    uint8_t seqScalingListPresentFlag[8];
+    uint8_t picScalingMatrixPresentFlag;
```
xuguangxin: better change this to bool

interface/VideoEncoderDefs.h
```diff
@@ -185,6 +185,16 @@ typedef struct VideoParamsROI {
     VideoROI roi[MAX_ROI_NUM];
 } VideoParamsROI;
 
+typedef struct VideoParamsQM {
+    uint8_t scalingList4x4[6][16];  //in raster scan order
+    uint8_t scalingList8x8[2][64];  //in raster scan order
+
+    uint8_t seqScalingMatrixPresentFlag;
```
xuguangxin: better change this to bool and change name to seqScalingMatrixPresent

encoder/vaapiencoder_h264.cpp
```diff
@@ -1887,6 +2037,25 @@ bool VaapiEncoderH264::ensurePicture (const PicturePtr& picture, const SurfacePt
     return true;
 }
 
+bool VaapiEncoderH264::ensureQMatrix(const PicturePtr& picture)
+{
+    assert (picture);
+
+    const VideoParamsQM *qMatrix = &m_videoParamAVC.qMatrix;
+    if (qMatrix->picScalingMatrixPresentFlag == 0 &&
```
xuguangxin: We following webkit code style, it perfer `if (!qMatrix->picScalingMatrixPresent)`  see https://webkit.org/code-style-guidelines/#zero-comparison for details

encoder/vaapiencoder_h264.cpp
```diff
@@ -1516,21 +1627,60 @@ bool VaapiEncoderH264::fill(VAEncPictureParameterBufferH264* picParam, const Pic
...
+
+    if (qMatrix->picScalingMatrixPresentFlag) {
+        list_present_flag = qMatrix->picScalingListPresentFlag;
+    } else if (qMatrix->seqScalingMatrixPresentFlag) {
+        list_present_flag = qMatrix->seqScalingListPresentFlag;
+    } else {
+        return true;
+    }
+
+    for (i = 0; i < 8; i++) {
```
xuguangxin: Could we define a macro for 8 and 6?

encoder/vaapiencoder_h264.cpp
```diff
@@ -383,21 +482,21 @@ bit_writer_write_sps(BitWriter* bitwriter,
...
-
-    #if 0
         if (seq->seq_fields.bits.seq_scaling_matrix_present_flag) {
-          for (i = 0; i < (seq->seq_fields.bits.chroma_format_idc != 3 ? 8 : 12); i++) {
-            bit_writer_put_bits_uint8(bitwriter, seq->seq_fields.bits.seq_scaling_list_present_flag, 1);
-            if (seq->seq_fields.bits.seq_scaling_list_present_flag) {
-              assert(0);
-              /* FIXME, need write scaling list if seq_scaling_matrix_present_flag ==1*/
+            for (i = 0; i < 8; i++) {
+                bitwriter->writeBits(qMatrix->seqScalingListPresentFlag[i], 1);
```
xuguangxin: duplicate code between L678 and L685, could we add a function for this

encoder/vaapiencoder_base.cpp
```diff
@@ -430,6 +474,17 @@ bool VaapiEncoderBase::ensureMiscParams (VaapiEncPicture* picture)
         if (frameRate)
             fill(frameRate);
     }
+
+    uint32_t numROI = m_videoParamCommon.roiParams.numROI;
+    if (numROI != 0) {
```
xuguangxin: we perfer if (!numROI) see https://webkit.org/code-style-guidelines/#zero-comparison for details

xuguangxin (changes requested):
1. Please help add more description on each commit. 
2. Please add some gtest for new functions. you can refer https://github.com/01org/libyami/blob/apache/encoder/vaapiencoder_jpeg_unittest.cpp for example

xuguangxin:
@lizhong1008 , Please help to review

encoder/vaapiencoder_h264.cpp
lizhong1008: Yeah, strongly agree with Guangxin.  If aspect_ratio_idc is not equal to Extended_SAR, no need to set SarWidth and SarHeight. We can remove AspectRatioIdc to avoid parameters conflict.

encoder/vaapiencoder_h264.cpp
```diff
@@ -1429,7 +1429,14 @@ bool VaapiEncoderH264::fill(VAEncSequenceParameterBufferH264* seqParam) const
     /* VUI parameters are always set, at least for timing_info (framerate) */
     seqParam->vui_parameters_present_flag = TRUE;
     if (seqParam->vui_parameters_present_flag) {
-        seqParam->vui_fields.bits.aspect_ratio_info_present_flag = FALSE;
+        if (m_videoParamAVC.SAR.AspectRatioIdc != 0) {
+            seqParam->vui_fields.bits.aspect_ratio_info_present_flag = TRUE;
+            seqParam->aspect_ratio_idc = m_videoParamAVC.SAR.AspectRatioIdc;
+            seqParam->sar_width = m_videoParamAVC.SAR.SarWidth;
```
lizhong1008, line 1676: Only if( aspect_ratio_idc = = Extended_SAR ), sar_width and sar_height  need to be set as spec E.1.1

encoder/vaapiencoder_h264.cpp
```diff
@@ -1429,7 +1429,14 @@ bool VaapiEncoderH264::fill(VAEncSequenceParameterBufferH264* seqParam) const
...
+            seqParam->sar_height = m_videoParamAVC.SAR.SarHeight;
+        } else {
+            seqParam->vui_fields.bits.aspect_ratio_info_present_flag = FALSE;
```
lizhong1008, line 1679: Seems the line is redundant because the seqParam is memset when initialization.

interface/VideoEncoderDefs.h
```diff
@@ -173,6 +173,7 @@ typedef struct SliceNum {
 typedef struct SamplingAspectRatio {
     uint16_t SarWidth;
     uint16_t SarHeight;
+    uint8_t  AspectRatioIdc;
```
lizhong1008: No need to add AspectRatioIdc.

encoder/vaapiencoder_base.cpp
lizhong1008: @xuguangxin , I prefer to change to
  if(m_videoParamCommon.rcParams.bitRate > 0 &&m_videoParamCommon.rcMode ==  RATE_CONTROL_CQP)
  m_videoParamCommon.rcMode = RATE_CONTROL_CBR;
User may set bitrate but forget to set bitrate control mode, our encoder should can work at this case

encoder/vaapiencoder_h264.cpp
```diff
@@ -783,13 +785,29 @@ class VaapiEncPictureH264:public VaapiEncPicture
...
     }
 
     bool isIdr() const {
         return m_type == VAAPI_PICTURE_I && !m_frameNum;
     }
 
+    static YamiStatus getOutputAUD(VideoEncOutputBuffer* outBuffer)
+    {
+        const uint8_t aud[] = {0x00, 0x00, 0x00, 0x01, 0x09, 0xf0};
+        const uint32_t AUD_SIZE = 6;
```
lizhong1008: would better set AUD_SIZE = sizeof (aud).

encoder/vaapiencoder_h264.cpp
```diff
@@ -783,13 +785,29 @@ class VaapiEncPictureH264:public VaapiEncPicture
...
+        ASSERT(outBuffer);
```
lizhong1008: I think code scan tool will report warning. ASSERT(outBuffer && outBuffer->data) would be better.

encoder/vaapiencoder_h264.cpp
lizhong1008: VaapiEncoderH264::reorder will be called for every frame. we just need remove enableAUD and add a function in vaapiencoderh264.h:     bool enableAUD() const {
        return m_videoParamAVC.enableAUD;
    }

encoder/vaapiencoder_base.cpp
```diff
@@ -190,10 +190,9 @@ YamiStatus VaapiEncoderBase::setParameters(VideoParamConfigType type, Yami_PTR v
         VideoParamsCommon* common = (VideoParamsCommon*)videoEncParams;
         if (common->size == sizeof(VideoParamsCommon)) {
             PARAMETER_ASSIGN(m_videoParamCommon, *common);
-            //if(m_videoParamCommon.rcParams.bitRate > 0)
-	        // m_videoParamCommon.rcMode = RATE_CONTROL_CBR;
-	     // Only support CQP and CBR mode now
-            if (m_videoParamCommon.rcMode != RATE_CONTROL_CBR)
+            if (m_videoParamCommon.rcMode != RATE_CONTROL_CBR &&
+                m_videoParamCommon.rcMode != RATE_CONTROL_VBR &&
+                m_videoParamCommon.rcMode != RATE_CONTROL_CVBR)
```
lizhong1008, line 211: How about VA_RC_VBR_CONSTRAINED?

lizhong1008: And I prefer squash this commit with 8fe2b89a1dc0e7b3d95d45b5260fe40bf863106a.

encoder/vaapiencoder_base.cpp
lizhong1008: Would better set it here, if driver can not support VAConfigAttribEncROI, we can get error status from vaCreateConfig.

lizhong1008 (changes requested):

lizhong1008:
One more thing, it would better if we can use a short commit title and add more detail description in the commit message.

xuguangxin:
Hi @zhaobob 
any update on this?

zhaobob:
@xuguangxin @lizhong1008 , I really appreciate your comment. I will update it in two weeks for that I have something others to handle. I hope this will not delay your schedule.

encoder/vaapiencoder_h264.cpp
zhaobob: Actually, AUD is the start of frame.

encoder/vaapiencoder_h264.cpp
zhaobob: Agree

encoder/vaapiencoder_h264.cpp
zhaobob: @lizhong1008 , I think I have to follow the first solution because I can't get the pointer of VaapiEncoderH264 in VaapiEncPictureH264::getOutput().

encoder/vaapiencoder_h264.cpp
zhaobob: Only L680 to L684 is duplicated code. Others are different for SPS and PPS.

encoder/vaapiencoder_h264.cpp
zhaobob: Yes, I have to check the dest buffer is valid or not, right?

encoder/vaapiencoder_h264.cpp
zhaobob: Of course

encoder/vaapiencoder_base.cpp
zhaobob: @lizhong1008 , I got your point.

interface/VideoEncoderDefs.h
zhaobob: Of course.

interface/VideoEncoderDefs.h
zhaobob: Yes

interface/VideoEncoderDefs.h
zhaobob: Yes

interface/VideoEncoderDefs.h
zhaobob: Are you sure? But I see that bitrate is handled in this way.

encoder/vaapiencpicture.h
zhaobob: Yes. I will do that.

encoder/vaapiencoder_h264.cpp
zhaobob: Yes

encoder/vaapiencoder_h264.cpp
zhaobob: You're right.

encoder/vaapiencoder_h264.cpp
zhaobob: Yes

encoder/vaapiencoder_base.cpp
xuguangxin: Little confuse here, if the value is readonly, driver should not allow us set it right?

encoder/vaapiencoder_h264.cpp
xuguangxin: Could we write a function for this.like
writeScalingList(bitwriter, flag, index)

encoder/vaapiencoder_h264.cpp
xuguangxin: ok

encoder/vaapiencoder_h264.cpp
xuguangxin: Bob is right, instead of keep a point to VaapiEncoderH264, it's better to keep a bool in VaapiencPictureH264

encoder/vaapiencoder_h264.cpp
xuguangxin: This will send random value to driver since you did not fill the buffer.right?So this should not happen. Since you have checked value in 1251, you have two option
1. add a list_present_flag param to VaapiEncoderH264::fill(VAIQMatrixBufferH264\* buffer), I have kind of prefer this.
2.  assert at 1754.  it will make code more clearer.

interface/VideoEncoderDefs.h
xuguangxin: Yes. Not everybody care about the ROI. Everybody care about bitrate

xuguangxin (changes requested):

encoder/vaapiencoder_base.cpp
zhaobob: @lizhong1008, I think we should not to bring confusion here. If user set bitrate and qp but forget the mode, which mode we should choose?

encoder/vaapiencoder_base.cpp
zhaobob: @lizhong1008 , The other commit 8fe2b89 is for fixing the confusion of mode. This commit is for add VBR and CVBR. I prefer not to squash them.

zhaobob: I define RATE_CONTROL_CVBR = VA_RC_VBR_CONSTRAINED in header file.

encoder/vaapiencoder_base.cpp
zhaobob: done

encoder/vaapiencoder_base.cpp
zhaobob: This will enable the ROI for intel va driver.

interface/VideoEncoderDefs.h
zhaobob: I prefer not to change it. 
First, I saw that all configuration in vaapiencoder_base.cpp belong to VideoParamsCommon, VideoParamsCommon, VideoParamsAVC, VideoParamsVP9 and VideoParamsHRD are on the same layer. VideoFrameRate, VideoRateControlParams, VideoParamsROI should be on the other layer.
Second, this may influence the interface of gstyami.
Third, the cost for anyone that doesn't care about ROI is some bytes not used in structure VideoParamsCommon.
How about you think it?

encoder/vaapiencoder_h264.cpp
zhaobob: This will not happen because I have checked this before calling fill():
    if (!qMatrix->picScalingMatrixPresent &&
        !qMatrix->seqScalingMatrixPresent ) {
        return true;
    }
    VAIQMatrixBufferH264 *buffer;
    if (!picture->editQMatrix(buffer) || !fill(buffer)) {
        ERROR("failed to create qMatrix");
        return false;
    }

uartie (changes requested):
Build check failures caused by compilation error:

```
16:48:45,687 INFO  - vaapiencoder_base.cpp: In member function 'void YamiMediaCodec::VaapiEncoderBase::fill(VAEncMiscParameterRateControl*) const':
16:48:45,687 INFO  - vaapiencoder_base.cpp:425:18: error: 'VAEncMiscParameterRateControl {aka struct _VAEncMiscParameterRateControl}' has no member named 'max_qp'
16:48:45,687 INFO  - rateControl->max_qp = m_videoParamCommon.rcParams.maxQP;
16:48:45,687 INFO  - ^
16:48:45,736 INFO  - mv -f .deps/libyami_encoder_la-vaapicodedbuffer.Tpo .deps/libyami_encoder_la-vaapicodedbuffer.Plo
16:48:45,757 INFO  - Makefile:624: recipe for target 'libyami_encoder_la-vaapiencoder_base.lo' failed
16:48:45,757 INFO  - make[2]: *** [libyami_encoder_la-vaapiencoder_base.lo] Error 1
```

uartie:
Closing due to inactivity and broken builds.  Feel free to reopen request when you plan to address these issues.

PR #596, opened 2016-09-20
configure: fix the h265dec config inconformity issue.

mypopydev:
in the config help, default is enable h265dec,  in fact
is disable the h265dec by default, need to used the --enable-h265dec
enable h265 decoder. Change the h265enc default setting with
"yes" now.

Signed-off-by: Jun Zhao jun.zhao@intel.com


xuguangxin:
+1

PR #600, opened 2016-09-22
Flush B frame for 264 and 265, to fix #598

### LINKED ISSUE #598: [HEVC][encode] lose frame when set '--ipperiod' as 2.
$ yamiencode -W xxx -H xxx -c HEVC -i input.I420 -o out.h265 --qp xxx --ipperiod 2
$ yamidecode -i out.h265 -m 0 -f I420

Then the dumped out.h265.I420 file size is less than input.I420. For the correct result, the file size should be same.

wudping:


xuguangxin:
@lizhong1008 , help to review
thanks

xuguangxin:
@lizhong1008 , please help to verify it too. thanks

wudping:
libyami-tuils, use:
https://github.com/01org/libyami-utils/pull/39

encoder/vaapiencoder_h264.cpp
```diff
@@ -1232,6 +1244,26 @@ YamiStatus VaapiEncoderH264::doEncode(const SurfacePtr& surface, uint64_t timeSt
...
+    ret = reorder(surface, timeStamp, forceKeyFrame);
+    if (ret != YAMI_SUCCESS)
+        return ret;
+
+    ret = encodeAllFrames();
+    if (ret != YAMI_SUCCESS) {
+        return ret;
+    }
+
+    INFO();
```
lizhong1008: Why a NULL INFO here?

encoder/vaapiencoder_hevc.cpp
```diff
@@ -1135,6 +1156,26 @@ YamiStatus VaapiEncoderHEVC::doEncode(const SurfacePtr& surface, uint64_t timeSt
...
+    ret = reorder(surface, timeStamp, forceKeyFrame);
+    if (ret != YAMI_SUCCESS)
+        return ret;
+
+    ret = encodeAllFrames();
+    if (ret != YAMI_SUCCESS) {
+        return ret;
+    }
+
+    INFO();
```
lizhong1008: a NULL INFO

lizhong1008 (commented):
LGTM except some NULL INFO.

lizhong1008:
One more thing,   please don't forget to change the commet in VideoEncoderInterface.h:
///obsolete, discard cached data (input data or encoded video frames), not sure why an encoder need this
    virtual void flush(void) = 0;

you can change it to flush cached data (input data or encoded video frames)

encoder/vaapiencoder_h264.cpp
wudping: thanks @lizhong1008 , I will delete it.

encoder/vaapiencoder_hevc.cpp
wudping: I will delete it.

wudping:
thanks @lizhong1008 for your constructive advise.

wudping:
updated according to comments.

lizhong1008:
+1

lizhong1008:
@wang-fei , please help to verify it. Don't forget to check h264 also.

encoder/vaapiencoder_base.cpp
```diff
@@ -89,12 +89,12 @@ YamiStatus VaapiEncoderBase::start(void)
 void VaapiEncoderBase::flush(void)
 {
     AutoLock l(m_lock);
```
chivakker, line 91: The line above can be removed.

chivakker (changes requested):
This patch was merged with obvious problems especially on the encoder base class.  flush on the base class is doing nothing now.  So the proper change would be to remove all calls to base::flush in all codecs. 

Please fix it.

encoder/vaapiencoder_base.cpp
lizhong1008: @chivakker , yes, it can be removed.  Thanks for indicate it.

lizhong1008: @chivakker , "This patch was merged with obvious problems especially on the encoder base class. flush on the base class is doing nothing now. So the proper change would be to remove all calls to base::flush in all codecs. Please fix it."   Maybe we can fix h264 and h265 encoder in this patch series， and send a separated patch to fix other codecs?   We just to fix h264 and h265 b frame issue here,  what do you think?

lizhong1008:
@xuguangxin , seems this patch has been merged without "verify +1", do we need wangfei's verification before merging it?

encoder/vaapiencoder_base.cpp
xuguangxin: Hi @chivakker 
Agree with Lizhong, we should focus on h264, h265 in this patch. I think we had better not change every codec for this base on following reason:
1.  Currently, we do not have any thing in VaapiEncoderBase::flush, it do not mean we will not add something in future. 
2.  Very codec have right to choose call base class flush or not.
3.  If it is no functional or readability benefit. We had better not change the code. If we change vp9encoder, author need do test for vp9encoder. This is an extra effort for author and it's not introduce benefit 
Do you think it is make sense?

xuguangxin: Hi @chivakker 
the flush only be call one time. remove it will not improve performance or readbility. Do we really need  remove it?

xuguangxin:
@lizhong1008 , QA is not suppose verify our pull requst  unlease it's a big update. I veirfied it.It's ok.

lizhong1008:
@xuguangxin , so this it our alignment with QA team, right? I am ok if we get alignment.

xuguangxin:
I remember we sync with QA before, but you can double confirm it

PR #601, opened 2016-09-22
vp9enc: return error on unsupported resolutions

chivakker:
non 8-bit aligned resolutions are not supported

Fix for #582, after this was agreed for the current release.


xuguangxin:
+1

PR #603, opened 2016-09-26
h264dec:Load more bytes when meeting epb

LindaYu17:
Reloaded data might be less than time_scale when epb exists, then
an error reports. This only occurs in 32bit device but not 64bit device.
The solution is to load more byte(s) when meeting epb.

### DIFF
--- a/codecparsers/nalReader.cpp
+++ b/codecparsers/nalReader.cpp
@@ -45,21 +45,26 @@ inline bool NalReader::isEmulationBytes(const uint8_t *p) const
 void NalReader::loadDataToCache(uint32_t nbytes)
 {
     const uint8_t *pStart = m_stream + m_loadBytes;
+    const uint8_t *p;
+    const uint8_t *pEnd = m_stream + m_size;
     /*the numbers of emulation prevention three byte in current load block*/
     uint32_t epb = 0;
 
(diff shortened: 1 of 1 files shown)


xuguangxin:
Haha, this means we need run our daily regression on 32 bits platfrom too.

Hi @LindaYu17 
Although this is an important change, but it will change so many things on 64 bits platform. All h264 and h265 test case need rerun. Since we are very close to 1.0.0 release. Do you think it is ok if we merge this after it?
thanks

xuguangxin:
@wudping , Please help to do test for h264 and h265.
thanks

wudping:
okay, @xuguangxin , you are welcome.

wudping:
hi @xuguangxin , do I have to finish that only on 32-bit machine?

xuguangxin:
No, Linda have tested all h264 clips on 32bit platform. You only need test this on 64 bit platform. But in long term, I hope we have 32bit platform to run daily regression too.Ubuntu still ship 32 bit for 16.04...

wudping:
okay, I get it.

wudping:
verify + 1

xuguangxin:
merged as 77ba612d93c1d004fff85c0c6c841ce555317d5b. Thanks Linda

PR #605, opened 2016-09-28
V4l2 update

halleyzhao:


xuguangxin:
Hi @halleyzhao , we are busy with 1.0 release. We will review this after PRC notional holiday.

v4l2/v4l2_decode.cpp
```diff
@@ -39,12 +39,12 @@
...
 #define INT64_TO_TIMEVAL(i64, time_val) do {            \
-        time_val.tv_sec = (int32_t)(i64 >> 32);         \
-        time_val.tv_usec = (int32_t)(i64 & 0xffffffff); \
+        time_val.tv_sec = (int32_t)(i64 >> 31);         \
+        time_val.tv_usec = (int32_t)(i64 & 0x7fffffff); \
     } while (0)
 #define TIMEVAL_TO_INT64(i64, time_val) do {            \
         i64 = time_val.tv_sec;                          \
-        i64 = (i64 << 32) + time_val.tv_usec;           \
+        i64 = (i64 << 31)  + time_val.tv_usec;  \
```
xuguangxin: If we have overflow when we shift 32 , it's highly possible overflow when we shift 31. since time_val.tv_usec is range from 0 to 999,999, why not make i64 = time_val.tv_sec*1000000 + time_val.tv_usec

v4l2/v4l2_codecbase.cpp
```diff
@@ -660,6 +660,23 @@ const char* mimeFromV4l2PixelFormat(uint32_t pixelFormat)
...
+public:
+    VideoFrameDeleter(VADisplay display)
+        : m_display(display)
+    {
+    }
+    void operator()(VideoFrame* frame)
+    {
+        VASurfaceID id = (VASurfaceID)frame->surface;
+        DEBUG("destroy VASurface 0x%x", id);
+        vaDestroySurfaces(m_display, &id, 1);
```
xuguangxin, line 673: need delete frame here

xuguangxin (changes requested):

v4l2/v4l2_decode.cpp
halleyzhao: time_val is fake here.
timestamp is usually int64_t from demuxer, but v4l2 uses time_val. so I simply split int64 to two int32 of time_val.
when we feed the lowest 32bit of int64 to a int32, it may overflow int32. so I just feed the lowest 31 bit to the first int32; the reset bits to the second int32. 
(potential issue is that valid bits in int64 is 62 bits, not 63bits as before.  however, it's big enough; and much bigger than the capacity of time_val)

v4l2/v4l2_codecbase.cpp
halleyzhao: update patch. thanks

v4l2/v4l2_decode.cpp
halleyzhao: either client (player/demuxer) or libyami uses int64 for pts, but v4l2 uses time_val/2xInt32.
in fact, libyami doesn't care of the real value of pts, but uses POC; ;libyami simply return timestamp back on getOutput.

xuguangxin: 31 is magic number, people who do not blame the code will guess, why not use 32 etc.
Since manual said time_val.tv_usec never larger than 1000000, Do you think it's ok to use  `i64 = time_val.tv_sec*1000000 + time_val.tv_usec`, and `time_val.tv_sec = i64/1000000; time_val.tv_usec = i64%1000000`

halleyzhao: it is doable, but I think shift 31 is better.
1. shift 31 has a bigger data range than time_val. in fact time_val lose precision compare to int64 pts, the unit of pts needn't stick to usec.
2. shift 31 is more efficient since there is no divide and multiply

v4l2/v4l2_decode.cpp
```diff
@@ -352,33 +361,42 @@ int32_t V4l2Decoder::ioctl(int command, void* arg)
...
-            if (bufferCount == m_reqBuffCnt)
+            m_outputBufferCountOnInit++;
+            if (m_outputBufferCountOnInit == m_reqBuffCnt)
                 mapVideoFrames(m_videoWidth, m_videoHeight);
         }
 #elif __ENABLE_WAYLAND__
-        struct v4l2_buffer* qbuf = static_cast<struct v4l2_buffer*>(arg);
-        static uint32_t bufferCount = 0;
+        // FIXME, m_outputBufferCountOnInit should be reset on output buffer change (for example: resolution change)
+        // it is not must to init video frame here since we don't accepted external for wayland yet. however, external buffer may be used in the future
```
chivakker: File a bug for this.

v4l2/v4l2_decode.cpp
```diff
@@ -518,9 +536,14 @@ int32_t V4l2Decoder::ioctl(int command, void* arg)
             m_videoHeight = outFormat->height;
             m_maxBufferSize[OUTPUT] = m_videoWidth * m_videoHeight + ((m_videoWidth +1)/2*2) * ((m_videoHeight+1)/2);
         } else {
+        #ifdef YUNOS
```
chivakker: do not use #ifdef YUNOS. In these lines remove all else case.

vaapi/VaapiUtils.cpp
```diff
@@ -67,4 +67,68 @@ uint32_t getRtFormat(uint32_t fourcc)
...
+        WARNING("invalid surface/image width: %d, height: %d", image.width, image.height);
+        return false;
+    }
+
+    if (!fp) {
+        memset(fileName, 0, sizeof(fileName));
+        if (useSingleFile) {
+            sprintf(fileName, "%s/%c%c%c%c_%dx%d", "/tmp/yami", ptrC[0], ptrC[1], ptrC[2], ptrC[3], image.width, image.height);
+        } else {
+            sprintf(fileName, "%s/%04d_%c%c%c%c_%dx%d", "/tmp/yami", _frameCount, ptrC[0], ptrC[1], ptrC[2], ptrC[3], image.width, image.height);
```
chivakker, line 101: This line requires Clang format.

chivakker (changes requested):
Do not use YUNOS as a Macro.

v4l2/v4l2_decode.cpp
halleyzhao: I'm glad to do so.
then you'd update the glue layer for chromeos, to accept EAGAIN, instead of errno of EINVAL.

v4l2/v4l2_decode.cpp
halleyzhao: https://github.com/01org/libyami/issues/618

vaapi/VaapiUtils.cpp
halleyzhao: update his patch by git-clang-format --style=WebKit HEAD^

xuguangxin (approved):

xuguangxin:
Wating for @chivakker  feedback

v4l2/v4l2_decode.cpp
chivakker: I am curious why clang format didn't fix the comment.  Can you comment on that

chivakker (changes requested):
I think there's still some problems with clang format.  Otherwise it looks good.  Thanks for removing the YUNOS, we will change the wrapper code.

xuguangxin:
It do have some format issue but it's not include daniel spotted one
Hi Halley:

```
git-clang-format --style=WebKit --binary=clang-format-3.7 b256a6a6c9cba869b9f0eb01eb992c991bc8ff65
changed files:
    v4l2/v4l2_codecbase.cpp
    v4l2/v4l2_codecbase.h
    v4l2/v4l2_decode.cpp
```

halleyzhao:
update each change set with clang format.  

as to daniel's comments, I'm not clear his points. 
"git-clang-format --style=WebKit --binary=clang-format-3.7 HEAD^" doesn't change anything for that patch.
 anyway, I split the comments to 3 sentences manually.

xuguangxin:
Thanks halley.

xuguangxin:
Hi @chivakker 
Is it ok to merge this?
thanks

halleyzhao:
@chivakker Daniel: is it ok now?

halleyzhao:
@seanvk what's your opinion

chivakker (approved):

chivakker:
Change looks good now.  I know clang is kind of tricky especially when wrapping lines.  I am sorry I couldn't get back to this patch earlier.

xuguangxin:
@wudping , please help to review this. it  should work with https://github.com/01org/libyami-utils/pull/45/files, v4l2decoder/v4l2encoder should work.
thanks

xuguangxin:
@halleyzhao ,sorry,  since we are close to 1.0.1 release (it will include SVC-T), we need delay this merge about 2 weeks.

xuguangxin:
merged as ef7953bc7196c959051a368b5f2c0a71ad4a0a37.
thanks hally

PR #609, opened 2016-09-30
baseenc: make flush as empty function, add instructions for derive class

xuguangxin:
Currently we do not have any thin in baseenc::flush, but in future we may add something,
So best have some message for derive class developer

### DIFF
--- a/encoder/vaapiencoder_base.cpp
+++ b/encoder/vaapiencoder_base.cpp
@@ -88,7 +88,9 @@ YamiStatus VaapiEncoderBase::start(void)
 
 void VaapiEncoderBase::flush(void)
 {
-    AutoLock l(m_lock);
+    /* Current version of VaapiEncoderBase::flush is empty fucntion
+     * But we may add something in future.All derive class need call this in derive::flush()
+     */
 }
 
 YamiStatus VaapiEncoderBase::stop(void)


xuguangxin:
this will fix https://github.com/01org/libyami/issues/610

xuguangxin:
@chivakker , Please help to review.thanks

lizhong1008:
+1

xuguangxin:
Hi @chivakker:
Do you think it's ok to check in this?
thanks

chivakker (approved):
lgtm :+1:

xuguangxin:
merged as cc34729a0e998ba03335a04b84a4fe5d3eadd143

encoder/vaapiencoder_base.cpp
```diff
@@ -88,7 +88,9 @@ YamiStatus VaapiEncoderBase::start(void)
 
 void VaapiEncoderBase::flush(void)
 {
-    AutoLock l(m_lock);
+    /* Current version of VaapiEncoderBase::flush is empty fucntion
+     * But we may add something in future.All derive class need call this in derive::flush()
+     */
```
uartie, line 93: One way to enforce the `base::flush()` to be called is to "not allow" the derived class to override it.  Add an empty "protected" `virtual doFlush() { }` to the base class and derived classes can implement that if they want, instead.  Then, in the "public" `base::flush()` method we just call `doFlush()`.  This way, the derived class does not have to remember to call `base::flush()`.

xuguangxin: It is good suggestion, but it is somewhat complicate, I hope we can find something for flush in near future. ;)

PR #614, opened 2016-10-10
egl: add __ENABLE_EGL__ to on/off egl compilation

LindaYu17:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -104,6 +104,18 @@ fi
 AM_CONDITIONAL(ENABLE_X11,
     [test "x$enable_x11" = "xyes"])
 
+AC_ARG_ENABLE(egl,
+    [AC_HELP_STRING([--enable-egl],
+        [enable egl @<:@default=yes@:>@])],
+    [], [enable_egl="yes"])
+
+if test "$enable_egl" = "yes"; then
+    AC_DEFINE([__ENABLE_EGL__], [1],
+        [Defined to 1 if --enable-egl="yes"])
+fi
+AM_CONDITIONAL(ENABLE_EGL,
+    [test "x$enable_egl" = "xyes"])
+
 AC_ARG_ENABLE(wayland,
     [AC_HELP_STRING([--enable-wayland],
         [enable wayland @<:@default=no@:>@])],
--- a/v4l2/v4l2_codecbase.h
+++ b/v4l2/v4l2_codecbase.h
@@ -31,9 +31,11 @@
     #if __ENABLE_X11__
     #include <X11/Xlib.h>
     #endif
+    #if __ENABLE_EGL__
     #include <EGL/egl.h>
     #define EGL_EGLEXT_PROTOTYPES
     #include "EGL/eglext.h"
+    #endif
 #endif
 #include "VideoCommonDefs.h"
 #include "v4l2codec_device_ops.h"
@@ -101,7 +103,9 @@ class V4l2CodecBase {
     };
     virtual int32_t usePixmap(uint32_t bufferIndex, Pixmap pixmap) {return 0;};
     #endif
+    #if __ENABLE_EGL__
     virtual int32_t useEglImage(EGLDisplay eglDisplay, EGLContext eglContext, uint32_t buffer_index, void* egl_image) {return 0;};
+    #endif
 #endif
     bool setDrmFd(int drm_fd)
     {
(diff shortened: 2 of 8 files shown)


configure.ac
```diff
@@ -104,6 +104,18 @@ fi
...
+        [enable egl @<:@default=yes@:>@])],
+    [], [enable_egl="yes"])
+
+if test "$enable_egl" = "yes"; then
+    AC_DEFINE([__ENABLE_EGL__], [1],
+        [Defined to 1 if --enable-egl="yes"])
+fi
+AM_CONDITIONAL(ENABLE_EGL,
+    [test "x$enable_egl" = "xyes"])
+
```
chivakker, line 118: What is that you want to accomplish here?  How is your environment failing without this?

egl/egl_util.c
```diff
@@ -97,3 +98,4 @@ EGLImageKHR createEglImageFromHandle(EGLDisplay eglDisplay, EGLContext eglContex
 
     return eglImage;
 }
+#endif
```
chivakker: In principle, we should not have a ENABLE_EGL within a EGL file.  If we're compiling this on non-egl scenarios that has to be fixed.

egl/egl_util.h
```diff
@@ -33,3 +34,4 @@ EGLImageKHR createEglImageFromHandle(EGLDisplay eglDisplay, EGLContext eglContex
 #endif /* __cplusplus */
 
 #endif
+#endif
```
chivakker: same as in egl_util.c file comment.

egl/egl_vaapi_image.cpp
```diff
@@ -164,3 +165,4 @@ bool getVaFormat(VADisplay display, VAImageFormat& format)
 }
 
 }//namespace YamiMediaCodec
+#endif
```
chivakker: same as in egl_util.c file comment

egl/egl_vaapi_image.h
```diff
@@ -54,3 +55,4 @@ class EglVaapiImage
 }//namespace YamiMediaCodec
 
 #endif
+#endif
```
chivakker: same as in egl_util.c file comment

v4l2/v4l2_codecbase.h
```diff
@@ -101,7 +103,9 @@ class V4l2CodecBase {
     };
     virtual int32_t usePixmap(uint32_t bufferIndex, Pixmap pixmap) {return 0;};
     #endif
+    #if __ENABLE_EGL__
     virtual int32_t useEglImage(EGLDisplay eglDisplay, EGLContext eglContext, uint32_t buffer_index, void* egl_image) {return 0;};
+    #endif
```
chivakker, line 108: This virtual method can be defined regardless of configure flag enable_egl.  Then the implementations should consider adding the enable_egl if required.

chivakker (commented):
Can you please elaborate on the problem that you want to resolve with this patch? Thanks.

xuguangxin:
Hi @LindaYu17 , is any update for Daniel's comments

configure.ac
LindaYu17: just want to use ENABLE_EGL to simply disable/enable it because in many scenarios we don't need to compile egl related code to the library.

egl/egl_util.c
LindaYu17: thanks for the comments.
updated.
decide to compile it or not in v4l2/makefile.am

v4l2/v4l2_codecbase.h
chivakker: Nevermind the above comment.  EGLDisplay, EGLContext are only defined with ENABLE_EGL so this is ok

configure.ac
chivakker: Ok, it makes sense, thanks for the explanation.

chivakker (approved):
The change looks good to me now.  Please have someone verify it independently before merging.

xuguangxin:
Hi Dongping:
Could help to verify, it should be buildable when we enable or disable egl. 
thanks

wudping:
verify + 1

xuguangxin:
seems related to https://github.com/01org/libyami-utils/pull/47, will merge this after https://github.com/01org/libyami-utils/pull/47 fixed

xuguangxin:
@LindaYu17 , this will conflict with apache branch, please help to rebase it.
thanks

LindaYu17:
@xuguangxin rebase done.

xuguangxin:
merged as 18864d6a9b8748b8ae64abb32ad0a80886ecd11b thanks

PR #615, opened 2016-10-11
v4l2:call setVaDisplay() and createVpp() in decode

LindaYu17:
Move above two functions from V4l2CodecBase to V4l2Decoder to
avoid being called in encode path.

### DIFF
--- a/v4l2/v4l2_codecbase.cpp
+++ b/v4l2/v4l2_codecbase.cpp
@@ -337,23 +337,6 @@ int32_t V4l2CodecBase::ioctl(int command, void* arg)
             ASSERT(ret != -1);
             if (port == INPUT) {
                 DEBUG("start decoding/encoding");
-#if (defined(ANDROID) || defined(__ENABLE_WAYLAND__))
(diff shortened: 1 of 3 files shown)


v4l2/v4l2_codecbase.cpp
```diff
@@ -337,23 +337,6 @@ int32_t V4l2CodecBase::ioctl(int command, void* arg)
...
-                        ret = -1;
-                    }
-
-                    if (!createVpp()) {
-                        ERROR("fail to set up VPP");
-                        ret = -1;
-                    }
-                }
-#endif
                 boolRet = start();
```
xuguangxin, line 340: This conflict with your change log, Move code to start still called by decode thread.

LindaYu17: it should be called in decode path, but not in encode path.

xuguangxin:
ok, thanks for explain, +1

xuguangxin:
merged as 39c05bc616387407d78160796ee8821cddc79965, thank Linda

PR #616, opened 2016-10-11
scale_vpp: check out odd resolution for VPP

wudping:
check out odd resolution for scaler VPP, if so print error, then return.

Signed-off-by: wudping dongpingx.wu@intel.com

### DIFF
--- a/vpp/vaapipostprocess_scaler.cpp
+++ b/vpp/vaapipostprocess_scaler.cpp
@@ -93,6 +93,11 @@ VaapiPostProcessScaler::process(const SharedPtr<VideoFrame>& src,
     if (!src || !dest) {
         return YAMI_INVALID_PARAM;
     }
+    if(src->crop.width & 0x01 || src->crop.height & 0x01){
+        ERROR("unsupported odd resolution");
+        return YAMI_FAIL;
+    }
+
     copyVideoFrameMeta(src, dest);
     SurfacePtr surface(new VaapiSurface(dest));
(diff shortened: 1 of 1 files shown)


vpp/vaapipostprocess_scaler.cpp
```diff
@@ -93,6 +93,12 @@ VaapiPostProcessScaler::process(const SharedPtr<VideoFrame>& src,
     if (!src || !dest) {
         return YAMI_INVALID_PARAM;
     }
+    if(dest->crop.width & 0x01 || dest->crop.height & 0x01 ||
```
xuguangxin: Do not check dest->crop.width and dest->crop.height. When you do composite,it's very likely have odd resolution in dest,

xuguangxin (changes requested):

vpp/vaapipostprocess_scaler.cpp
wudping: thanks @xuguangxin , you mean that I need only to check the source crop.
YAMI can support output odd resolution, is right?

xuguangxin: Yes,driver can support this. But you can double confirm it

wudping: thanks @xuguangxin , I have updated it.

xuguangxin (approved):

xuguangxin:
merged as 08e2403eec7212bfff8b8928d7702ef1935d5bcc

PR #619, opened 2016-10-13
vaapidecoder_h265 : refresh fourcc information in vaapidecoder_h265.

popkartyeah:
intel_driver will change fourcc to VA_FOURCC_P010 when bit_depth_chroma_minus8 or
bit_depth_luma_minus8 value is 2 , so did libyami.

Signed-off-by: Yun Zhou yunx.z.zhou@intel.com


xuguangxin:
This should be an issue for VA driver, driver should not change surface format and reallocate surface. The surfaces we give to driver maybe from display. If driver reallocate it we cannot display it any more.
I have a branch for 10 bit hevc, but it just basically work, we did not do conformance test base on it.

You can refer to https://github.com/xuguangxin/libyami/tree/hevc10

popkartyeah:
ffmpeg_libyami need to get a right fourcc which libyami decoder can support after meet DECODE_FORMAT_CHANGE state, and display the right fourcc output yuv image.

xuguangxin:
For fourcc issue, please try my branch, if it have problem. I can fix it

PR #621, opened 2016-10-17
Vp8Parser: Zero-initialize segment feature data when required

### LINKED ISSUE #620: Decode vp8 stream from libvpx results in video with macroblocks
the vp8 decoder, when decoding attached vp8 stream(unzip to output.ivf), will show many macroblocks and unexpected color blocks. This does not happen if we decode using libvpx.

[output.zip](https://github.com/01org/libyami/files/532372/output.zip)

chivakker:
When segmentation is enabled and update_segment_feature_data flag is
true, but no actual update value is provided in the stream, the value
should be assigned the default "0", instead of what was carried over
from previous frame. This applies to both quantizer and loop filter
update values.

Patch taken from Chromium and reviewed.
Review-Url: https://codereview.chromium.org/1992663002

Original autor is mentioned on the link above and it was adapted
to libyami

Fixes #620

### DIFF
--- a/codecparsers/vp8_parser.cpp
+++ b/codecparsers/vp8_parser.cpp
@@ -274,6 +274,8 @@ bool Vp8Parser::ParseSegmentationHeader(bool keyframe) {
       BD_READ_BOOL_OR_RETURN(&quantizer_update);
       if (quantizer_update)
         BD_READ_SIGNED_OR_RETURN(7, &shdr->quantizer_update_value[i]);
+      else
+	  shdr->quantizer_update_value[i] = 0;
     }
 
     for (size_t i = 0; i < kMaxMBSegments; ++i) {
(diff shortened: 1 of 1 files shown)


xuguangxin:
Hi @chivakker 
thanks for quick response. It really helpful. 

Hi @wudping 
Please help verify #620 and do full round test on vp8.

Hi @taste1981
You can try it if you like.

thanks

taste1981:
[after_fix.zip](https://github.com/01org/libyami/files/535136/after_fix.zip)

@xuguangxin with attached vp8 stream, i'm still getting the issue even after the fix is in.

xuguangxin:
Hi @taste1981 
I can't reproduce you issue, maybe because I have other two patch to skip standalone P frame. Please try apply https://github.com/01org/libyami/pull/623 too.

taste1981:
Applied all the patches. Now VP8 decoding works for me. I double checked, the previous patch should work if I repacked everything. So the patch in effect should be the previous one.

xuguangxin:
Glad to see this.

xuguangxin:
+1 for patch.
Hi @wudping 
Please test this and #623 together when #623 is +1ed
Thanks

wudping:
okay, I will verify once  #623 is +1ed, you are welcome.

lizhong1008:
LGTM, +1. Thanks Daniel.

wudping:
verify + 1

xuguangxin:
merged as e2a9e079fb52be69f7e7f3bb074ee132c1f0bac5

PR #623, opened 2016-10-18
fix vp8 decoder for networking streams.

xuguangxin:
two fixes here:
1. do not treat parser error as fatal error, in network condition, it always have error bitstream, we'd better not treat it as faltal error just skip current frame.
2. skip P frames if we do not see key frame. networking stream many  not start from a key frame, we should skip all P frame until we got a key

### DIFF
--- a/decoder/vaapidecoder_vp8.cpp
+++ b/decoder/vaapidecoder_vp8.cpp
@@ -452,6 +452,7 @@ VaapiDecoderVP8::VaapiDecoderVP8()
     // m_uvModeProbs[3];
     m_sizeChanged = 0;
     m_hasContext = false;
+    m_gotKeyFrame = false;
 }
 
 VaapiDecoderVP8::~VaapiDecoderVP8()
@@ -508,6 +509,7 @@ void VaapiDecoderVP8::flush(void)
     m_lastPicture.reset();
     m_goldenRefPicture.reset();
     m_altRefPicture.reset();
+    m_gotKeyFrame = false;
 
     VaapiDecoderBase::flush();
 }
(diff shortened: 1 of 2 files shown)


xuguangxin:
Hi @chivakker 
Please help to review.thanks

decoder/vaapidecoder_vp8.cpp
```diff
@@ -542,6 +544,13 @@ YamiStatus VaapiDecoderVP8::decode(VideoDecodeBuffer* buffer)
             status = ensureContext();
             if (status != YAMI_SUCCESS)
                 return status;
+            m_gotKeyFrame = true;
+        }
+        else {
+            if (!m_gotKeyFrame) {
+                //we can't decode frame without key
```
lizhong1008: Would better add some debug information here, such as WARNING.

lizhong1008: Seems other codec may have similar issue, do we need to extend m_gotKeyFrame to other codecs?

xuguangxin: 1. I can add warning,
2. It's not needed by h264, h265. The parser will return error if no sps https://github.com/01org/libyami/blob/apache/codecparsers/h264Parser.cpp#L521
   for vp9, I do not have enough time to fix it, maybe we can file a bug to track it.

xuguangxin:
@lizhong1008 updated and create #637 for vp9

xuguangxin:
@lizhong1008 , is it ok to start test this?

xuguangxin:
Hi @wudping 
Seems lizhong take vocation now, Could you start verification this, you can test h264/h265 flush issue with this in one round

wudping:
okay @xuguangxin , I will do at once.

wudping:
verify + 1

lizhong1008:
Thanks for the update. Review +1

PR #649, opened 2016-11-04
fix all unit failed in vaapi directory.

### LINKED ISSUE #648: 3 unittest failed for vaapi direcotry.
[==========] Running 9 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 9 tests from VaapiDisplayTest
[ RUN      ] VaapiDisplayTest.CreateInvalidHandleX11
vaapidisplay_unittest.cpp:106: Failure
Failed
[  FAILED  ] VaapiDisplayTest.CreateInvalidHandleX11 (0 ms)
[ RUN      ] VaapiDisplayTest.CreateInvalidHandleDRM
libva info: VA-API version 0.39.3
libva info: va_getDriverName() returns -1
libva error: va_getDriverName() failed with unknown libva error,driver_name=(null)
libyami error 4590 (vaapidisplay.cpp, 54): vaInitialize: unknown libva error
libyami error 4590 (vaapidisplay.cpp, 320): vaGetDisplay failed.
[       OK ] VaapiDisplayTest.CreateInvalidHandleDRM (0 ms)
[ RUN      ] VaapiDisplayTest.CreateInvalidHandleVA
libyami error 4590 (vaapidisplay.cpp, 320): vaGetDisplay failed.
libyami error 4590 (vaapidisplay.cpp, 320): vaGetDisplay failed.
[       OK ] VaapiDisplayTest.CreateInvalidHandleVA (0 ms)
[ RUN      ] VaapiDisplayTest.CreateX11
libva info: VA-API version 0.39.3
libva info: va_getDriverName() returns 0
libva info: Trying to open /myopt/yami/vaapi/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_0_39
libva info: va_openDriver() returns 0
vaapidisplay_unittest.cpp:183: Failure
Value of: display.get() == display2.get()
  Actual: true
Expected: false
vaapidisplay_unittest.cpp:184: Failure
Expected: (display->getID()) != (display2->getID()), actual: 0x1c8ff00 vs 0x1c8ff00
libva info: VA-API version 0.39.3
libva info: va_getDriverName() returns 0
libva info: Trying to open /myopt/yami/vaapi/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_0_39
libva info: va_openDriver() returns 0
[  FAILED  ] VaapiDisplayTest.CreateX11 (4 ms)
[ RUN      ] VaapiDisplayTest.CreateDRM
libva info: VA-API version 0.39.3
libva info: va_getDriverName() returns 0
libva info: Trying to open /myopt/yami/vaapi/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_0_39
libva info: va_openDriver() returns 0
libva info: VA-API version 0.39.3
libva info: va_getDriverName() returns 0
libva info: Trying to open /myopt/yami/vaapi/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_0_39
libva info: va_openDriver() returns 0
[       OK ] VaapiDisplayTest.CreateDRM (0 ms)
[ RUN      ] VaapiDisplayTest.CreateVA
[       OK ] VaapiDisplayTest.CreateVA (0 ms)
[ RUN      ] VaapiDisplayTest.CompatibleX11
libva info: VA-API version 0.39.3
libva info: va_getDriverName() returns 0
libva info: Trying to open /myopt/yami/vaapi/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_0_39
libva info: va_openDriver() returns 0
vaapidisplay_unittest.cpp:278: Failure
Value of: isCompatible(display, native)
  Actual: true
Expected: false
[  FAILED  ] VaapiDisplayTest.CompatibleX11 (1 ms)
[ RUN      ] VaapiDisplayTest.CompatibleDRM
libva info: VA-API version 0.39.3
libva info: va_getDriverName() returns 0
libva info: Trying to open /myopt/yami/vaapi/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_0_39
libva info: va_openDriver() returns 0
[       OK ] VaapiDisplayTest.CompatibleDRM (1 ms)
[ RUN      ] VaapiDisplayTest.CompatibleVA
[       OK ] VaapiDisplayTest.CompatibleVA (0 ms)
[----------] 9 tests from VaapiDisplayTest (6 ms total)

[----------] Global test environment tear-down
[==========] 9 tests from 1 test case ran. (6 ms total)
[  PASSED  ] 6 tests.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] VaapiDisplayTest.CreateInvalidHandleX11
[  FAILED  ] VaapiDisplayTest.CreateX11
[  FAILED  ] VaapiDisplayTest.CompatibleX11

 3 FAILED TESTS

xuguangxin:
There are vaapi display compatible  issues. This will fix https://github.com/01org/libyami/issues/648

### DIFF
--- a/vaapi/vaapidisplay_unittest.cpp
+++ b/vaapi/vaapidisplay_unittest.cpp
@@ -103,14 +103,14 @@ protected:
 
 #if __ENABLE_X11__
 VAAPI_DISPLAY_TEST(CreateInvalidHandleX11) {
-    FAIL(); // remove when fixed;
+    //this turn off gest warning about multithread
+    //you can check details at
+    //https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md
+    ::testing::FLAGS_gtest_death_test_style = "threadsafe";
 
-    // FIXME: This causes segmentation fault.
-    // VaapiDisplay should be able to handle this gracefully
     NativeDisplay native = {1, NATIVE_DISPLAY_X11};
-    DisplayPtr display = VaapiDisplay::create(native);
-
-    EXPECT_FALSE(display.get());
+    //this will crash vaGetDisplay
+    ASSERT_DEATH({ DisplayPtr display = VaapiDisplay::create(native); }, "");
 }
 #endif
 
--- a/vaapi/vaapidisplay_unittest.cpp
+++ b/vaapi/vaapidisplay_unittest.cpp
@@ -308,6 +308,9 @@ VAAPI_DISPLAY_TEST(CompatibleDRM) {
     native.type = NATIVE_DISPLAY_DRM;
     EXPECT_TRUE(isCompatible(display, native));
 
+    native.type = NATIVE_DISPLAY_AUTO;
+    EXPECT_TRUE(isCompatible(display, native));
+
     native.handle = 0;
     native.type = NATIVE_DISPLAY_VA;
     EXPECT_FALSE(isCompatible(display, native));
@@ -344,6 +347,9 @@ VAAPI_DISPLAY_TEST(CompatibleVA) {
     native.handle = 1;
     native.type = NATIVE_DISPLAY_X11;
     EXPECT_FALSE(isCompatible(display, native));
+
+    native.type = NATIVE_DISPLAY_AUTO;
+    EXPECT_TRUE(isCompatible(display, native));
 }
 
 } //namespace YamiMediaCodec
--- a/vaapi/vaapidisplay_unittest.cpp
+++ b/vaapi/vaapidisplay_unittest.cpp
@@ -140,11 +140,16 @@ VAAPI_DISPLAY_TEST(CreateInvalidHandleVA) {
 
     EXPECT_FALSE(display.get());
 
+    //this turn off gest warning about multithread
+    //you can check details at
+    //https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md
+    ::testing::FLAGS_gtest_death_test_style = "threadsafe";
+
     native.handle = -1;
     native.type = NATIVE_DISPLAY_VA;
-    display = VaapiDisplay::create(native);
 
-    EXPECT_FALSE(display.get());
+    //This will crash in vaDisplayIsValid
+    ASSERT_DEATH({ display = VaapiDisplay::create(native); }, "");
 }
 
 #if __ENABLE_X11__
--- a/vaapi/vaapidisplay.cpp
+++ b/vaapi/vaapidisplay.cpp
@@ -138,6 +138,8 @@ class NativeDisplayDrm : public NativeDisplayBase{
     };
 
     bool isCompatible(const NativeDisplay& display) {
+        if (display.type == NATIVE_DISPLAY_AUTO)
+            return true;
         if (display.type != NATIVE_DISPLAY_DRM)
             return false;
         if (isInvalidDrmHandle(display.handle) || display.handle == m_handle)
@@ -157,6 +159,8 @@ class NativeDisplayVADisplay : public NativeDisplayBase{
     };
 
     bool isCompatible(const NativeDisplay& display) {
+        if (display.type == NATIVE_DISPLAY_AUTO)
+            return true;
         return display.type == NATIVE_DISPLAY_VA && display.handle == m_handle;
     }
 };
--- a/vaapi/vaapidisplay.cpp
+++ b/vaapi/vaapidisplay.cpp
@@ -155,7 +155,9 @@ class NativeDisplayVADisplay : public NativeDisplayBase{
     virtual bool initialize (const NativeDisplay& display) {
         ASSERT(display.type == NATIVE_DISPLAY_VA);
 
-        return acceptValidExternalHandle(display);
+        if (acceptValidExternalHandle(display))
+            return true;
+        return vaDisplayIsValid((VADisplay)display.handle);
     };
 
     bool isCompatible(const NativeDisplay& display) {
--- a/vaapi/vaapidisplay.cpp
+++ b/vaapi/vaapidisplay.cpp
@@ -78,6 +78,11 @@ class NativeDisplayBase {
     DISALLOW_COPY_AND_ASSIGN(NativeDisplayBase);
 };
 
+bool isInvalidDrmHandle(int handle)
+{
+    return handle == 0 || handle == -1;
+}
+
 #if __ENABLE_X11__
 class NativeDisplayX11 : public NativeDisplayBase{
   public:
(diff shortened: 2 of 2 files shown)


xuguangxin:
@uartie , Please help to review if you have time.
thanks

uartie (commented):
Why are you removing these tests?  It is checking that yami does the right thing when we give it an invalid handle.

vaapi/vaapidisplay_unittest.cpp
```diff
@@ -101,19 +101,6 @@ class VaapiDisplayTest : public ::testing::Test {
...
-
-    // FIXME: This causes segmentation fault.
-    // VaapiDisplay should be able to handle this gracefully
-    NativeDisplay native = {1, NATIVE_DISPLAY_X11};
-    DisplayPtr display = VaapiDisplay::create(native);
-
-    EXPECT_FALSE(display.get());
-}
-#endif
-
```
uartie, line 116: Don't remove the test!  Fix VaapiDisplay::create!

vaapi/vaapidisplay_unittest.cpp
```diff
@@ -134,19 +121,6 @@ VAAPI_DISPLAY_TEST(CreateInvalidHandleDRM) {
...
-    EXPECT_FALSE(display.get());
-
-    native.handle = -1;
-    native.type = NATIVE_DISPLAY_VA;
-    display = VaapiDisplay::create(native);
-
-    EXPECT_FALSE(display.get());
-}
-
```
uartie, line 149: Don't remove test!

vaapi/vaapidisplay_unittest.cpp
uartie: Ah yes, I see.  There's no way to determine if a pointer address is valid or not.  In that case, we should turn this into a DEATH test, instead, since we "expect" it to crash.

uartie: See "Death Tests" in the docs: https://github.com/google/googletest/blob/master/googletest/docs/V1_7_AdvancedGuide.md#death-tests

vaapi/vaapidisplay_unittest.cpp
uartie: This test passes for me

xuguangxin: It's passed only because it's a special value, 0 and -1, if you send -2 or 1 to create it will failed. Do we really need this test?

vaapi/vaapidisplay_unittest.cpp
xuguangxin: I think Death Tests used for internal state checking.  In our case, we will crash in vaGetDisplay. It should not apply to death test, right?

uartie: It doesn't matter where it crashes.  The point is we are testing that invalid input to our library function produces the "expected" outcome... in this case, we expect it to crash to tell the library user that they are doing something very wrong.  And a test such as this not only exercises our "negative" code path, it documents it too.

vaapi/vaapidisplay_unittest.cpp
uartie: There is nothing wrong with testing "special" values, especially for the "negative" path.  No matter how silly it may seem, there will always be someone who tries to do it.  We don't always want to test only the "good" (or "happy") path.  You never know when someone will come along and "accidentally" mess up the internal logic that handles 0 and -1 cases.  This test will catch that if no one else does.

IMO, any test is better than "no" test, no matter how trivial or inane it might seem.

BTW, if I use -2 or 1, I get a valid DisplayPtr... that is probably not good.  Seems to me that our intention is that a valid DisplayPtr means it's okay to use it.  In these cases, it is not okay.  Seems to me that we should be checking vaDisplayIsValid during the create function... and write separate DEATH tests to verify the expectation of such.

xuguangxin: Ok, thanks for detailed explain, it make sense to me. I will change it.

vaapi/vaapidisplay_unittest.cpp
xuguangxin: Ok, I can change it.thanks for explain.

xuguangxin:
@uartie , fixed, please help to review.
thanks

vaapi/vaapidisplay_unittest.cpp
```diff
@@ -103,14 +103,9 @@ class VaapiDisplayTest : public ::testing::Test {
 
 #if __ENABLE_X11__
 VAAPI_DISPLAY_TEST(CreateInvalidHandleX11) {
-    FAIL(); // remove when fixed;
-
-    // FIXME: This causes segmentation fault.
-    // VaapiDisplay should be able to handle this gracefully
     NativeDisplay native = {1, NATIVE_DISPLAY_X11};
```
uartie, line 111: Add `::testing::FLAGS_gtest_death_test_style = "threadsafe";` here to eliminate the gtest warning mentioned in your commit message

vaapi/vaapidisplay_unittest.cpp
```diff
@@ -136,15 +131,16 @@ VAAPI_DISPLAY_TEST(CreateInvalidHandleDRM) {
 
 VAAPI_DISPLAY_TEST(CreateInvalidHandleVA) {
     NativeDisplay native = {0, NATIVE_DISPLAY_VA};
-    DisplayPtr display = VaapiDisplay::create(native);
```
uartie, line 140: Add ::testing::FLAGS_gtest_death_test_style = "threadsafe"; here.

uartie (changes requested):

vaapi/vaapidisplay_unittest.cpp
xuguangxin: Little confused about this, we did not create new thread in unit test or library. Why gtest warning us about multithread things

uartie: Gtest uses fork for death tests by default which it claims is unsafe if we are using threads in the library.  For some reason, gtest is "unable to detect" whether we are using threads in the library so it warns us.

AFAICT, instructing gtest to just use a thread for death tests instead of fork is harmless and better.  That is what the above code statement does.

xuguangxin: thanks for explain, updated.

uartie (approved):

uartie:
LGTM... but needs verified.  Thanks.

xuguangxin:
thanks you for reivew, we will start an internal test next week.  I will merge this and your factory refact code after this.

wudping:
verify + 1

xuguangxin:
merged as da4d1a527576d4eeb57ddfd92972acdc7a714e09. thanks U.Arite for review

PR #654, opened 2016-11-09
Factory registration fixes, unittests and other programmer logic fixes

uartie:
Move factory registration for enc/dec/vpp to the host files.  This ensures the static registration is invoked before any public host interface method calls are executed by library users.

Define host (i.e. host interface) tests for encode, decode, and vpp to demonstrate the issue.  The tests are compiled into a separate unittest_host binary for each codec type.

Also fixes for a few other programmer errors.

### DIFF
--- a/decoder/vaapidecoder_host_unittest.cpp
+++ b/decoder/vaapidecoder_host_unittest.cpp
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2016 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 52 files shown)


xuguangxin:
great! thanks for refact this

decoder/vaapidecoder_host_unittest.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    EXPECT_EQ( expect, actual );
+}
+
+INSTANTIATE_TEST_CASE_P(
+    MimeType, VaapiDecoderHostTest,
+    ::testing::Values(
+        YAMI_MIME_H264, YAMI_MIME_AVC, YAMI_MIME_H265, YAMI_MIME_HEVC,
+        YAMI_MIME_MPEG2, YAMI_MIME_VC1, YAMI_MIME_VP8, YAMI_MIME_VP9,
+        YAMI_MIME_JPEG));
```
chivakker, line 115: This could be moved to a const vector.

chivakker (approved):
lgtm :+1:.  I only have a non-functional comment that can be modified later.

wudping:
verify + 1

xuguangxin:
merged as 225aec2e6bfd0505b8474dba1145d99e88548ee2

PR #655, opened 2016-11-10
vaapic: memset misc object

lizhong1008:

### DIFF
--- a/encoder/vaapiencpicture.h
+++ b/encoder/vaapiencpicture.h
@@ -124,6 +124,7 @@ createMiscObject(VAEncMiscParameterType miscType, T * &bufPtr)
     if (obj) {
         misc->type = miscType;
         bufPtr = (T *) (misc->data);
+        memset(bufPtr, 0, sizeof(T));
     }
     return obj;
 }


chivakker (changes requested):
The change makes sense but the commit message is rather vague.  There's no such a thing as vaapic, vaapi-encpicture has to be cleared to 0 when creating a miscellaneous object.

encoder/vaapiencpicture.h
```diff
@@ -124,6 +124,8 @@ createMiscObject(VAEncMiscParameterType miscType, T * &bufPtr)
     if (obj) {
         misc->type = miscType;
         bufPtr = (T *) (misc->data);
+        if (bufPtr)
```
xuguangxin: This if is not needed, data is an array in misc, if we can access misc->type, misc->data never be NULL.

lizhong1008: @xuguangxin misc is an input of createBufferObject,  and drive will init misc->data, may change it to NULL.  So we should check bufPtr

xuguangxin: The data's prototype is array.  misc->data will get the array address. if misc is not null, misc->data always be a valid value.

lizhong1008:
@chivakker  yeah, your comments make sense, it is a typo. I'll update it.

chivakker (approved):

wudping:
verify + 1

xuguangxin:
merged as 0ee515f10124740a8f0d9830210be5d8d2142387

PR #656, opened 2016-11-16
common: Add pooled video frame allocator

LindaYu17:

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -3,6 +3,7 @@ libyami_common_source_c = \
 	utils.cpp \
 	nalreader.cpp \
 	surfacepool.cpp \
+	PooledFrameAllocator.cpp \
 	YamiVersion.cpp \
 	$(NULL)
 
--- a/common/PooledFrameAllocator.cpp
+++ b/common/PooledFrameAllocator.cpp
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "PooledFrameAllocator.h"
+#include <common/log.h>
+
+namespace YamiMediaCodec {
+
+struct SurfaceDestoryer
+{
+private:
+    SharedPtr<VADisplay> m_display;
+    std::vector<VASurfaceID> m_surfaces;
+public:
+    SurfaceDestoryer(const SharedPtr<VADisplay>& display, std::vector<VASurfaceID>& surfaces)
+        :m_display(display)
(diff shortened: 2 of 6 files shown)


common/videopool.h
```diff
@@ -25,11 +25,12 @@ template <class T>
 class VideoPool : public EnableSharedFromThis<VideoPool<T> >
 {
 public:
-    static SharedPtr<VideoPool<T> >
-    create(std::deque<SharedPtr<T> >& buffers)
+    VideoPool(std::deque<SharedPtr<T> >& buffers)
```
xuguangxin, line 28: Please split this to other commit

common/Makefile.am
```diff
@@ -3,10 +3,12 @@ libyami_common_source_c = \
 	utils.cpp \
 	nalreader.cpp \
 	surfacepool.cpp \
+	pooledframeallocator.cpp \
```
xuguangxin: Please use PooledFrameAllocator.cpp. We should follow webkit style https://webkit.org/code-style-guidelines/ for new file

common/pooledframeallocator.h
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef POOLED_FRAME_ALLOCATOR_H_
+#define POOLED_FRAME_ALLOCATOR_H_
```
xuguangxin: use PooledFrameAllocator_h, it's webkit style

xuguangxin (changes requested):

LindaYu17:
change done @xuguangxin

xuguangxin (approved):

common/PooledFrameAllocator.cpp
```diff
@@ -0,0 +1,88 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <PooledFrameAllocator.h>
```
xuguangxin: Sorry I missed this, This should be "" instead of <>

LindaYu17:
done @xuguangxin

common/PooledFrameAllocator.cpp
```diff
@@ -0,0 +1,88 @@
...
+            &attrib, 1);
+
+    if (status != VA_STATUS_SUCCESS) {
+        ERROR("create surface failed, %s", vaErrorStr(status));
+        return false;
+    }
+    std::deque<SharedPtr<VideoFrame> > buffers;
+    for (size_t i = 0;  i < surfaces.size(); i++) {
+        SharedPtr<VideoFrame> f(new VideoFrame);
+        memset(f.get(), 0, sizeof(VideoFrame));
```
chivakker, line 74: This memset can be moved to the constructor of VideoFrame.  There's a similar problem with the MiscBuffer that is fixed in that way.

chivakker (approved):
I only have one comment but it minor at this point.  If submitter sees more benefit on pushing the memset to the constructor please do.

common/PooledFrameAllocator.cpp
xuguangxin: VideoFrame do not have constructor. It is a pod type; it will be use in capi too.

xuguangxin: Could give more detail description for MiscBuffer thing?

xuguangxin:
merged as 9c7bb50639233a505599a17d74caa4a0dd0ee0ad

PR #658, opened 2016-11-17
Add decode api test framework for all codecs 

xuguangxin:
This patch set not intend to merge. I just use it for preview and collect patch set from different component owner.

Hi @chivakker 
Could you help to fix mpeg2 patch? It will break the yamidecode. The dumped md5 is totally wrong.
Please also help to add mpeg2 releated unit test in DecoderApi_unittest.cpp too.


vaapi/vaapidisplay.cpp
```diff
@@ -128,6 +128,8 @@ class NativeDisplayDrm : public NativeDisplayBase{
     };
```
xuguangxin, line 138: Please ignore this patch, I keep it here just make sure the decode api unit test runnable. I have other pull request to address this problem. https://github.com/01org/libyami/pull/649. I will rebase this after it merged

xuguangxin:
This is a fix for https://github.com/01org/libyami/issues/661, one important feature missed here, flush. We will add this later.

xuguangxin:
@lizhong1008 , Please helpto review h264 patch, please give some suggestion for the failed case mentioned in the commit. thanks

decoder/vaapidecoder_mpeg2.cpp
```diff
@@ -517,65 +517,18 @@ YamiStatus VaapiDecoderMPEG2::fillConfigBuffer()
...
-            m_videoFormatInfo.height = m_configBuffer.height;
-            m_videoFormatInfo.surfaceWidth = m_configBuffer.width;
-            m_videoFormatInfo.surfaceHeight = m_configBuffer.height;
-            m_configBuffer.surfaceWidth = m_configBuffer.width;
-            m_configBuffer.surfaceHeight = m_configBuffer.height;
-            status = YAMI_DECODE_FORMAT_CHANGE;
-        }
+    if (setFormat(width, height, width, height, kMinSurfaces + 1)) {
+        m_VAStart = false;
+        return YAMI_DECODE_FORMAT_CHANGE;
```
chivakker, line 528: setFormat doesn't check for width and height being greater than m_videoFormatInfo. As such it could be that the information doesn't get updated and VA is not restarted as it should.  It looked strange that passing same width and height twice to setFormat will produce a format change.  Remember that VA doesn't need to be restarted when surfaces can be re-used.

decoder/vaapidecoder_mpeg2.cpp
```diff
@@ -908,8 +861,7 @@ YamiStatus VaapiDecoderMPEG2::decode(VideoDecodeBuffer* buffer)
             DEBUG("decode tries for a sequence_header");
             if (!m_VAStart) {
                 status = processConfigBuffer();
-                if (status != YAMI_SUCCESS
-                    && status != YAMI_DECODE_FORMAT_CHANGE) {
+                if (status != YAMI_SUCCESS) {
```
chivakker, line 864: don't modify this.

decoder/vaapidecoder_mpeg2.cpp
```diff
@@ -934,8 +886,7 @@ YamiStatus VaapiDecoderMPEG2::decode(VideoDecodeBuffer* buffer)
             if (m_previousStartCode
                 == YamiParser::MPEG2::MPEG2_SEQUENCE_HEADER_CODE) {
                 status = processConfigBuffer();
-                if (status != YAMI_SUCCESS
-                    && status != YAMI_DECODE_FORMAT_CHANGE) {
+                if (status != YAMI_SUCCESS) {
```
chivakker, line 889: don't modify this, there's no need to send the extra format change when sequence header code came.

decoder/vaapidecoder_mpeg2.cpp
xuguangxin: This checked by isSurfaceGeometryChanged. Here is basic logical
1. Format change return for any width, height, susrfaceWidth, surfaceHeight, fourcc, pdb size change. This give user a chance to reconfig the renderer
2. We need flush dpb and output all previous frames since format changed.
3. We only reallocate surface pool when surface geometry changed (this make sure by decoder/vaapidecoder_base.cpp:L328)
Do you think this make sense?

decoder/vaapidecoder_mpeg2.cpp
xuguangxin: We need return user for every format change right?

decoder/vaapidecoder_mpeg2.cpp
xuguangxin: So we need move setFormat to processSlice, just like we did in VaapiDecoderH264::decodeSlice?

xuguangxin:
close this since we have #695

PR #662, opened 2016-11-18
h264enc: fix baseline profile decoding failure bug

lizhong1008:
svc-t requires to set slice header by yami instead of driver,
but yami always pads some 1/0 bits to make sure slice header is bytes
aligned. Actually, we need to pad some bits 1 to cabac mode, but no need to
pad 0 to cavlc mode, getBitWriterData() may pad some bits 0 when fluch
cached data. That is why
main/high profile work well because cabac is default mode, and baseline
failed due to cabac mode.

### DIFF
--- a/encoder/vaapiencoder_h264.cpp
+++ b/encoder/vaapiencoder_h264.cpp
@@ -1661,11 +1661,12 @@ bool VaapiEncoderH264::addPackedPrefixNalUnit(const PicturePtr& picture) const
 
     bit_writer_write_trailing_bits(&bs);
 
+    uint32_t codedBits = bs.getCodedBitsCount();
     uint8_t* codedData = bs.getBitWriterData();
-    ASSERT(codedData);
+    ASSERT(codedData && codedBits);
 
     if (!picture->addPackedHeader(VAEncPackedHeaderRawData, codedData,
-                                  bs.getCodedBitsCount())) {
+                                  codedBits)) {
         ret = false;
     }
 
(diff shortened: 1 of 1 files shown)


lizhong1008:
It is fix https://github.com/01org/libyami/issues/663

lizhong1008:
@LindaYu17 please help to verify it. Thanks.

LindaYu17:
verified +1

encoder/vaapiencoder_h264.cpp
```diff
@@ -1661,11 +1661,12 @@ bool VaapiEncoderH264::addPackedPrefixNalUnit(const PicturePtr& picture) const
 
     bit_writer_write_trailing_bits(&bs);
 
+    uint32_t codedBits = bs.getCodedBitsCount();
```
xuguangxin, line 1664: could we have some comments on code, for the pad thing?  With time pass, not everybody will read the commit log. 
Do we have problem in hevc and jpeg? since it use addPackedHeader

xuguangxin: @lizhong1008 , any update on this?

lizhong1008: @xuguangxin , jpeg has not called bit writer, such as getCodedBitsCount. For HEVC encoder,     bit_writer_write_trailing_bits(&bs) will pad some zeros, so it is not a issue.

lizhong1008: Maybe I can send a separated patch to add some comments of bitwritter.

xuguangxin: ok.

xuguangxin:
merged as 71ec0184157d3535f87999d85216e9301fa324be

PR #669, opened 2016-11-24
dec_vp9: support vp9 10bit depth odd resolution

wudping:
Change the way to calculate the size of P010;
for the odd resolution, need to change the chrominance's measurements to
even number, but the luminance's need not.
For example, P010 179x223, the size of luminance plan should be:
179x223x2
the size of chrominance plan should be:
180x224
So the size of P010 179x223 is:
179x223x2 + 180x224.

But for the even resolution, the size of P010 is:
Width * Hieght * 3
For example P010 1960x1080, the size is:
1960x1080x3


xuguangxin:
Please remove this line  { YAMI_FOURCC_P010, 2, { 4, 4 }, { 2, 1 } },

wudping:
okay thanks @xuguangxin

wudping:
updated, please help to review.

xuguangxin:
merged as 5bd0b21323f93723509aeaf34bb5b8ad875b21fc

PR #671, opened 2016-11-28
vp9_dec: modify the way to align resolution with 64, fix VIZ-8822

wudping:
In vp9 decoder, replace "SB_ALIGN(w) (((w | 0x3f) + 1) >> 6)" with
"#define SB_ALIGN(w) (((w) + 63) >> 6)".
Because when "w" is 0, the result of aligning should be 0, but (0|0x3f + 1)>>6 is
1. So when the low six bits of "w" is 0, the result will be wrong.

to fix VIZ-8822

### DIFF
--- a/codecparsers/vp9parser.cpp
+++ b/codecparsers/vp9parser.cpp
@@ -302,8 +302,8 @@ uint32_t get_min_lb_tile_cols(uint32_t sb_cols)
     return log2;
 }
 
-/* align to 64 */
-#define SB_ALIGN(w) (((w | 0x3f) + 1) >> 6)
+/* align to 64, follow specification 6.2.6 Compute image size syntax */
+#define SB_ALIGN(w) (((w) + 63) >> 6)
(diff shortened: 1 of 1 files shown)


codecparsers/vp9parser.cpp
```diff
@@ -303,7 +303,7 @@ uint32_t get_min_lb_tile_cols(uint32_t sb_cols)
 }
 
 /* align to 64 */
-#define SB_ALIGN(w) (((w | 0x3f) + 1) >> 6)
+#define SB_ALIGN(w) (((w) + 63) >> 6)
```
xuguangxin, line 306: Please add this to comments
/*6.2.6 Compute image size syntax */

wudping: thanks for your help @xuguangxin .

wudping:
updated, please help review again.

xuguangxin:
merged as 2ff190f56644b64d62e323ab92399296a50b045a

PR #672, opened 2016-11-29
common: init width and height

wudping:
assign the initial values into width and height in function setFormat().


wudping:
please help to review, thanks in advance.

xuguangxin:
please add some comments on this like fix what problem.
thanks

wudping:
thanks @xuguangxin , 
If not, the size of decoded YUV file will be 0.

wudping:
added comment in commit message, thanks @xuguangxin .

PR #674, opened 2016-12-02
common: change return value of guessFourcc

wudping:
Replace return value YAMI_FOURCC_I420 with 0 of funciton guessFourcc().
Because we don't know if the return value YAMI_FOURCC_I420 is the default
value or the actual. If we can't guess the fourcc from the file name,
the default value should be set by user according to different use case.

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -56,7 +56,7 @@ uint32_t guessFourcc(const char* fileName)
         }
     }
 
-    return YAMI_FOURCC_I420;
+    return 0;
 }
 
 bool guessResolution(const char* filename, int& w, int& h)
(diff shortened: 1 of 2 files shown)


wudping:
libyami-utils pull-request [#65](https://github.com/01org/libyami-utils/pull/65#674) has changed according to this modify, please help review, thanks in advance.

common/utils.cpp
```diff
@@ -56,7 +56,7 @@ uint32_t guessFourcc(const char* fileName)
         }
     }
 
-    return YAMI_FOURCC_I420;
+    return 0;
```
xuguangxin, line 59: We treat i420 as default yuv output.
This will break yamidecode -i xx.264 -o aaa.yuv.

wudping: okay, I got it, I will delete this patch.

wudping: thanks @xuguangxin .

wudping:
close this pull-request, because we not need this any more.

PR #676, opened 2016-12-06
bitWriter: don't return false when overflow and add some comments

lizhong1008:


codecparsers/bitWriter.cpp
```diff
@@ -71,8 +71,7 @@ void BitWriter::flushCache()
 bool BitWriter::writeBits(uint32_t value, uint32_t numBits)
 {
     if (value >= (uint64_t)1 << numBits) {
-        ERROR("Write Bits Error: value overflow");
-        return false;
+        WARNING("Write Bits: value overflow");
```
xuguangxin, line 74: little confused by this,why upper layer do not clip the value. 
If we write(1024, 5) this will silently write a zero for us?

lizhong1008: 1. In many case, upper layer don't know it is overflow. If we check it in upper layer, we need check every time before call writBits and many code lines need to be changed. Return false is not a idea as my comments in the commit log.
2. If we write(1024, 5), it will report a warning.

lizhong1008:
@wudping , please help to verify these three patches. Thanks

wudping:
okay, you are welcome.

xuguangxin (approved):

wudping:
verify + 1

PR #680, opened 2016-12-13
interface/VideoCommonDefs.h: include stdbool.h

zhijianli88:
stdbool.h defines 'bool' and it's compatible with C/C++.
Previously, we define bool as int for C compiler, which
is not identical with C++. The former occupies 4 bytes, and
the later occupies 1 bytes at x86_64 platform, which will
cause some unexpected behaviours at using capi.

Reported-by: Zhao, Xinda <xinda.zhao@intel.com>
Signed-off-by: Li Zhijian <zhijianx.li@intel.com>

### DIFF
--- a/interface/VideoCommonDefs.h
+++ b/interface/VideoCommonDefs.h
@@ -19,6 +19,7 @@
 // config.h should NOT be included in header file, especially for the header file used by external
 
 #include <stdint.h>
+#include <stdbool.h>
 
 #ifdef __cplusplus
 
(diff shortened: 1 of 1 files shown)


xuguangxin:
Hi @wudping ,
Please help to check why our regression test did not capture this issue.
thanks

xuguangxin:
thanks @zhijianli88 for patch and thanks @jsorg71  for feedback.

PR #687, opened 2016-12-21
VaapiSurface: compare crop size with original size

LindaYu17:
we need to compare crop width and height with original surface's width
and height, but not with cropped width and height, because it might be
cropped already before.


xuguangxin:
close this since it duplicate with #688

PR #688, opened 2016-12-21
VaapiSurface: compare crop size with original size

LindaYu17:
we need to compare crop width and height with original surface's width
and height, but not with cropped width and height, because it might be
cropped already before.


xuguangxin:
@taste1981 this will fix the vp8 crop issue, could you try it?

taste1981 (commented):
this fixes my issue on our conference server. Please help to merge this. Thanks!

xuguangxin:
merged as 26604c0184ccb479f4157207fab68af1c7988b96, thanks @taste1981  and @LindaYu17

PR #690, opened 2016-12-22
[Decode][h264/h265]Return YAMI_DECODE_INVALID_DATA instead of YAMI_SUCCESS if we see bitstream error

### LINKED ISSUE #684: [avc decoder]decoder does not report error when bitstream does not start with sps/pps
When decoding an h.264 bitstream that does not start with sps/pps, decode() does not report error when decoding nals before the first SPS/PPS.

We would expect decode call return error before the first SPS/PPS.

xuguangxin:
this give caller a notification about the bitstream error, so he can ask peer resend the key frame. It will fix https://github.com/01org/libyami/issues/684. @taste1981 please help to verify.
thanks.

### DIFF
--- a/decoder/vaapidecoder_h264.cpp
+++ b/decoder/vaapidecoder_h264.cpp
@@ -1137,7 +1137,8 @@ bool VaapiDecoderH264::decodeAvcRecordData(uint8_t* buf, int32_t bufSize)
         memset(&buffer, 0, sizeof(buffer));
         buffer.data = buf;
         buffer.size = bufSize;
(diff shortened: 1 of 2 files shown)


taste1981:
I'm not sure if this is what we want.... In your patch you return fail if you fail decoding any slice... but we're actually requesting returning failure if no ref.

taste1981 (approved):
This is verified to fix my issue.

lizhong1008:
+1

xuguangxin:
merged as 2f813e2314f661d9f68e0f1e51140894dd7e313d, thanks @taste1981

PR #691, opened 2016-12-23
V4L2-INPUT thread crashed after seeking

LindaYu17:
m_prevPic may be NULL if NalUnitType is NAL_SEI. Add a condition check
to make sure m_prevPic is not NULL before calling m_dpb.init.

Change-Id: Ibde719dac254c04e0b5348bffc1c47870bd087be
Signed-off-by: Matthew Lai <matthew.lai@intel.com>

### DIFF
--- a/decoder/vaapidecoder_h264.cpp
+++ b/decoder/vaapidecoder_h264.cpp
@@ -1683,6 +1683,10 @@ YamiStatus VaapiDecoderH264::createPicture(const SliceHeader* const slice,
     VaapiPictureType picStructure;
     bool isSecondField = false;
 
+    /* skip all non-idr slices if m_prevPic is a NULL */
+    if(!m_prevPic && !nalu->m_idrPicFlag)
+        return YAMI_DECODE_INVALID_DATA;
+
     if (slice->field_pic_flag) {
         if (slice->bottom_field_flag)
             picStructure = VAAPI_PICTURE_BOTTOM_FIELD;


decoder/vaapidecoder_h264.cpp
```diff
@@ -1760,7 +1760,7 @@ YamiStatus VaapiDecoderH264::decodeSlice(NalUnit* nalu)
         status = createPicture(slice, nalu);
         if (status != YAMI_SUCCESS)
             return status;
-        if (!m_currPic
+        if (!m_currPic || !m_prevPic
```
xuguangxin: This mean after seek, the first frame is not IDR? If it is an IDRit should set by https://github.com/01org/libyami/blob/apache/decoder/vaapidecoder_h264.cpp#L1721

laimatth: No, the first frame may not be an IDR.  Based on our testing, the first frame could be a SEI if the video contains SEI frames.  This is probably a mediaplayer's flaw, but it will be better if yami can handle this kind of defects.

xuguangxin: Agree with you, but sei will drop by decodeNalu and never goes here. So the root cause should be no key frame after seek, right?

xuguangxin:
hi @LindaYu17 
Why this not catch by our unit test, is anything special?

LindaYu17:
@xuguangxin we don't have this kind of test case. It is a special case which only exist in some video.

xuguangxin:
Hi @LindaYu17 
We have flush unit test here https://github.com/xuguangxin/libyami/commit/fb1f88770302a6edad2cf36f7022aee7ac92d7c9, it will cover seek to keyframe, but it never covers seek to no keyframe. 
Do we need to add this?

decoder/vaapidecoder_h264.cpp
laimatth: Yes, the sei will be dropped by decodeNalu but the next NAL_SLICE_NONIDR will cause the V4L2_INPUT thread to crash. 
By applying this modification, yami can skip all the NONIDR until reach the next IDR for start playing. This shall be a better result instead of thread crash, I think.

xuguangxin: Agree with you, but we'd better wait for Linda's unit test. I still wonder we have other issue or not. Another thing we should notice is, the player need fix the key frame issue too, if no, t the seek may take more time then user expected. one extream case is, you have 1-hour video,  but it only has 1 key frame, in this case, the user never sees any video after seek.

wudping:
verify+1

decoder/vaapidecoder_h264.cpp
lizhong1008: @laimatth , this patch is not to fix the SEI NAL issue, it is to fix the first slice is not IDR slice issue, right? 
1. I think you can refer  https://github.com/01org/libyami/pull/623/commits/4dca2e071159bf9a28581bbc83c56f460667a420 to fix such issue, because actually we don't need Previous Frame if current frame is IDR frame.  Yeah, I know we reset previous frame when IDR but it is not necessary in fact.
2. Would better to change the commit title and message. It is to fix h264 decoder issue not V4L2 issue. Maybe you can change it to "h264dec: don't start decode unit IDR slice", and create a V4L2 decoder issue then link it to the issue in the commit message.

lizhong1008 (changes requested):

xuguangxin:
Hi @lizhong1008 
Please help reiview again.
thanks

lizhong1008:
+1

wudping:
verify +　１

xuguangxin:
merged as b744e9b6df2a8a921f4d32975ce0f58edf127186, thanks Matthew and Linda

PR #692, opened 2016-12-27
add thread abstract for libyami

xuguangxin:
This will use java's executor pattern, but it only supports one thread. http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html
It will post jobs to the work thread, the work thread does not need care about lock and condition, it single thread. this will make multithread programer's life easy, we will use this in v4l2decoder/v4l2encoder in future.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -5,6 +5,7 @@ libyami_common_source_c = \
 	surfacepool.cpp \
 	PooledFrameAllocator.cpp \
 	YamiVersion.cpp \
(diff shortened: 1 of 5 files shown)


xuguangxin:
this will fix https://github.com/01org/libyami/issues/693

YuJiankang:
+1, lgtm

xuguangxin:
merged as 29b4f976a40305f6c68891380b12779b949b8b4a

Issue #696, opened 2016-12-30
[VP9][Encoder] '--qp' and '-f' does not work.


ghost:
1. yamiencode -i /media/encoderbitstreams/720x480p.yuv -s I420 -W 720 -H 480 --qp 100 -f 40 -c VP9 --rcmode CQP -o 720x480p.yuv_10.vp9
2 .mediainfo 720x480p.yuv_10.vp9
can get fps is 30.
3. set another qp value, and out stream is the same as 720x480p.yuv_10.vp9.

wudping:
code for this issue is merged.

ghost:
verify OK in 1.1.0 RC2 test.

PR #698, opened 2017-01-06
vp9_qp: setting qp to vp9 stream

wudping:
Setting the qp to luma_ac_qindex, replace the default value.

to fix issue #696 

### DIFF
--- a/encoder/vaapiencoder_vp9.cpp
+++ b/encoder/vaapiencoder_vp9.cpp
@@ -278,7 +278,8 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
 
     picParam->pic_flags.bits.show_frame = 1;
 
-    picParam->luma_ac_qindex = kDefaultQPValue;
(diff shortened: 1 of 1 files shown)


wudping:
please help to review.

encoder/vaapiencoder_vp9.cpp
```diff
@@ -278,7 +278,9 @@ bool VaapiEncoderVP9::fill(VAEncPictureParameterBufferVP9* picParam,
 
     picParam->pic_flags.bits.show_frame = 1;
 
-    picParam->luma_ac_qindex = kDefaultQPValue;
+    picParam->luma_ac_qindex =
+        (initQP() > minQP() && initQP() < maxQP()) ? initQP() : kDefaultQPValue;
```
xuguangxin: do we  need  (initQP() >= minQP() && initQP() <= maxQP()) ?

wudping: thanks @xuguangxin , You're right.
In that case, we can set minimum and maximum, I think.

wudping:
updated according to suggestion, please help to review.

ghost:
verify+1

xuguangxin:
merged as 441ee5485f9e045970b9e00faed10151594704ea

Issue #679, opened 2016-12-13
[H264][Encoder] Encoder will generate invalid frame when --ipperiod >= 5 --intraperiod >1500


wudping:
Setting ipperiod with 5 and intraperiod with 2000, the encoded video will be with wrong color from about the 1024th frame on. 
The cmd:
`yamiencode -i 33161F_1920x1080.265_1920x1080.I420 --intraperiod 2000 --ipperiod 5 -N 5000 -o intraperiod2000_ipperiod_5_1920x1080.264`

PR #700, opened 2017-01-13
264_enc: modify max_frame_num

wudping:
According to spec, log2_max_frame_num_minus4 should be in
the range[0, 12], so log2_max_frame_num should be in the
range[4, 16].

So as H265.

To fix issue [#679](https://github.com/01org/libyami/issues/679)

### DIFF
--- a/encoder/vaapiencoder_h264.cpp
+++ b/encoder/vaapiencoder_h264.cpp
@@ -121,7 +121,7 @@ h264_get_slice_type (VaapiPictureType type)
     return -1;
 }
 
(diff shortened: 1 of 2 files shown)


wudping:
to fix issue #679 , please help review.

xuguangxin (changes requested):
this not fix the problem just delay it from 2^10 to 2^16

wudping:
Thanks @xuguangxin , I am going to find out the original cause.

xuguangxin:
close this since it did fix the problem, need new patch to fix this

PR #702, opened 2017-02-08
265dec: stop decoding current picture when SEI

wudping:
In some bitsstreams, SEI NAL units are inserted between VCL NAL
units which belong to the one picture. If decoding the current
picture when meeting a SEI NAL unit, the VCL units after the SEI
will not be decoded.

to fix [VIZ-6406](https://jira01.devtools.intel.com/browse/VIZ-6406)


wudping:
please help to review.

xuguangxin (approved):

PR #705, opened 2017-02-10
265parser: cancel some special value checking

wudping:
As the specification, the value of delta_poc_s0_minus1 should
be in [0, 2^15], but some clip is with -3, so we have to cancel
this value checking to decode it completely.

to fix the decoding error of the one h265 container clip.


wudping:
please help to review.

xuguangxin (changes requested):
If delta_poc_s0_minus1 have problem, delta_poc_s1_minus1 will same problem.

wudping:
updated according to comment, please help to review,
Thanks @xuguangxin  for your suggestion.

xuguangxin (approved):

PR #717, opened 2017-03-03
add get and put surface api to libyami

xuguangxin:
add a callback to libyami, help user control which surface can be decode to. This is a critical interface for buffer sharing between decoder and display. We can't decode to screen surface.


YuJiankang (approved):
lgtm

wudping:
verify + 1

PR #720, opened 2017-03-16
Use zero copy api in v4l2 decoder.

xuguangxin:
The main target of this pull request is enabling zero copy feature in v4l2decoder. But it does more things beyond this. Current v4l2decoder is a little mess. It full of duplicate and full of #ifdef. It's hard to make it even buildable. So before I add zero copy patch. I use some patch to refact the v4l2decoder to make it more readable.


YuJiankang (approved):
lgtm

PR #722, opened 2017-03-17
travis: enable all codec and v4l2 compiling

wudping:
add "--enable-xxx" flag into compiling command line,
to compile all the codecs and v4l2 for every commit.

to fix [VIZ-9125](https://jira01.devtools.intel.com/browse/VIZ-9125).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,21 +27,20 @@ before_install:
   - sudo apt-get install -y libxext-dev
   - sudo apt-get install -y libxfixes-dev
   - sudo apt-get install -y intel-gpu-tools
+  - sudo apt-get install -y libv4l-0
+  - sudo apt-get install -y libv4l-dev
+  - sudo apt-get install -y libcogl-gles2-dev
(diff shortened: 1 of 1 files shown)


uartie:
Good call.  All codecs are enabled on intel/ci already... but we don't have v4l2 enabled.  We will be sure to enable v4l2 on intel/ci too ;)

uartie:
... we also have --enable-gtest option enabled during compilation on intel/ci to ensure it doesn't break compilation on new PR's.  Perhaps we could do that for travis-ci too after we bring gtest library directly into the libyami source tree to simplify gtest dependency.  For now, it should be ok to omit from travis-ci since it's covered by intel/ci.

xuguangxin:
Hi uarite,
Thanks for pay attention to this. v4l2 is important since we have customers use it.
I hope we can fix all gtest issues in next quarter. So we can enforce every pull request pass gest in intel/ci.

xuguangxin (approved):

PR #725, opened 2017-03-24
Jpeg unit

wudping:
implement JPEG decoderAPI unittest.

to figure out [VIZ-9126](https://jira01.devtools.intel.com/browse/VIZ-9126)

### DIFF
--- a/decoder/vaapiDecoderJPEG.h
+++ b/decoder/vaapiDecoderJPEG.h
@@ -47,6 +47,8 @@ private:
 
     YamiStatus finish();
 
+    YamiStatus ensureContext();
+
     SharedPtr<VaapiDecoderJPEG::Impl> m_impl;
     PicturePtr m_picture;
 
--- a/decoder/vaapiDecoderJPEG_unittest.cpp
+++ b/decoder/vaapiDecoderJPEG_unittest.cpp
@@ -345,11 +345,10 @@ TEST_P(JPEGTest, Decode_SimpleMulti)
 
     ASSERT_EQ(YAMI_SUCCESS, decoder.start(&config));
 
-    // Decode returns format change after it decodes the SOF segment.
+    // Decode returns format change for the first frame decoding at first time.
     ASSERT_EQ(YAMI_DECODE_FORMAT_CHANGE, decoder.decode(&buffer));
 
-    // Resume decoding.  The decoder will continue where it left off (after the
-    // SOF segment) when we pass the same buffer as previous call.
+    // Decode has been initiated by the last "decoder.decode(&buffer)" call.
     ASSERT_EQ(YAMI_SUCCESS, decoder.decode(&buffer));
 
     output = decoder.getOutput();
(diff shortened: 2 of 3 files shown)


wudping:

xuguangxin:
Hi @uartie , please help to review.
thanks

decoder/vaapiDecoderJPEG.h
```diff
@@ -33,6 +33,7 @@ class VaapiDecoderJPEG
     virtual YamiStatus start(VideoConfigBuffer*);
     virtual YamiStatus reset(VideoConfigBuffer*);
     virtual YamiStatus decode(VideoDecodeBuffer*);
+    YamiStatus ensureContext();
```
uartie: this should be a private method

decoder/vaapiDecoderJPEG_unittest.cpp
```diff
@@ -345,27 +345,12 @@ TEST_P(JPEGTest, Decode_SimpleMulti)
...
-
-    output = decoder.getOutput();
-    ASSERT_TRUE(bool(output));
-    EXPECT_EQ(GetParam()->getFourcc(), output->fourcc);
-
-    // Set buffer at second jpeg image
-    buffer.data = const_cast<uint8_t*>(mjpeg.data() + paramSize);
-    buffer.size = paramSize; // Length of second jpeg image data
-    buffer.timeStamp = 1;
-
```
uartie: We should keep this check for the second image buffer decode.  The only difference now is that we don't expect to get a format change on the next decode call.  That is, notice we are testing the decode for different image in the mjpeg sequence `buffer.data = const_cast<uint8_t*>(mjpeg.data() + paramSize);`

uartie (changes requested):

uartie:
The main difference I see in this change is that now we are parsing the jpeg header twice for every image.  Is that enough to notice any performance hit?

uartie:
previous comment edited

uartie:
ok, maybe just the first image header is parsed twice

decoder/vaapiDecoderJPEG_unittest.cpp
wudping: start() registered for M_SOF0 is empty now, and no YAMI_DECODE_FORMAT_CHANGE signal is triggered excepted for the current resolution is really different from the previous.

wudping: These two frames are the same.

wudping (commented):
Agree with you @uartie, just the first image header is parse twice.
We don't need to call VaapiDecoderJPEG::decode() twice for every JPEG frame.

wudping:
Updated according comment, please help to review. Thanks a lot.

decoder/vaapiDecoderJPEG_unittest.cpp
uartie: They have the same bit values, but the two frames are different data pointers.  Notice one buffer starts at "mjpeg.data()" offset and the other at "mjpeg.data() + paramSize" offset.  This verifies that the single decoder instance properly handles/decodes multiple images.  This simulates what the api user would do for mjpeg decoding.

uartie (changes requested):

decoder/vaapiDecoderJPEG_unittest.cpp
uartie: That is, please decode the second image in this test and verify the expectations.

wudping:
Thanks @uartie , the second JPEG frame in the same buffer decoding is restored to the previous.
Please help to review. Thanks again.

wudping:
Any misunderstanding I make, please let me know, I will modify it. Thanks.

uartie (approved):

xuguangxin:
thanks U.Arite

PR #730, opened 2017-04-01
Common unittest

### LINKED ISSUE #728: common/unittest with failed case
libyami]$./common/unittest
...................
[  FAILED  ] UtilsTest.guessResolutionOverflow (0 ms)
[ RUN      ] UtilsTest.getPlaneResolution
libyami error 10944 (utils.cpp, 193): do not support this format, fourcc
libyami error 10944 (utils.cpp, 193): do not support this format, fourcc Y800
libyami error 10944 (utils.cpp, 193): do not support this format, fourcc 411P
libyami error 10944 (utils.cpp, 193): do not support this format, fourcc ARGB
[       OK ] UtilsTest.getPlaneResolution (0 ms)
[----------] 4 tests from UtilsTest (0 ms total)

[----------] 1 test from ThreadTest
[ RUN      ] ThreadTest.FullTest
this may take some time
libyami error 10944 (Thread.cpp, 110): : sent job after stop()
[       OK ] ThreadTest.FullTest (10003 ms)
[----------] 1 test from ThreadTest (10003 ms total)

[----------] Global test environment tear-down
[==========] 11 tests from 4 test cases ran. (10005 ms total)
[  PASSED  ] 8 tests.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] NalReaderTest.ReadAsUnits
[  FAILED  ] NalReaderTest.ReadAsWhole
[  FAILED  ] UtilsTest.guessResolutionOverflow

wudping:
There are some error in libyami/common/unittest, this patches will fix them.

to fix #728

### DIFF
--- a/common/nalreader_unittest.cpp
+++ b/common/nalreader_unittest.cpp
@@ -53,20 +53,26 @@ const std::array<uint8_t, 39> g_data = {
     0x12
 };
 
-const std::array<int32_t, 5> g_nsizes = {9, 4, 6, 12, 7};
+const std::array<int32_t, 5> g_nsizes = { 6, 1, 3, 9, 4 };
 
 NALREADER_TEST(ReadAsUnits) {
     const uint8_t* nal;
-    int32_t offset(1), size;
+    const uint32_t start(3); //start code size
+    int32_t size; //NAL unit size
+    int32_t offset(1); //to skip 1 junk byte in g_data
 
-    NalReader reader(&g_data[0], g_data.size(), false);
+    NalReader reader(&g_data[0], g_data.size(), 0, false);
 
     for (unsigned n(0); n < g_nsizes.size(); ++n) {
+        //For unit reads, the NalReader returns buffer contents between each
+        //start code prefix, start code prefix is not included in read result;
+        //"nal": buffer contents;
+        //"size": buffer length;
         EXPECT_TRUE(reader.read(nal, size));
         EXPECT_EQ(size, g_nsizes[n]);
         for (int32_t i(0); i < size; ++i)
-            EXPECT_EQ(nal[i], g_data[i+offset]);
-        offset += size;
+            EXPECT_EQ(nal[i], g_data[i + offset + start]);
+        offset += (size + start);
     }
 
     EXPECT_FALSE(reader.read(nal, size));
@@ -75,14 +81,17 @@ NALREADER_TEST(ReadAsUnits) {
 NALREADER_TEST(ReadAsWhole) {
     const uint8_t* nal;
     int32_t size;
+    const int32_t offset = 4; // 1 initial junk byte + first start code
 
-    NalReader reader(&g_data[0], g_data.size(), true);
-
+    NalReader reader(&g_data[0], g_data.size(), 0, true);
+    //For whole read, the NalReader returns entire buffer contents after
+    //first start code. That is, the first start code prefix is not included
+    //in the result.
     EXPECT_TRUE(reader.read(nal, size));
-    EXPECT_EQ((size_t)size, g_data.size());
+    EXPECT_EQ((size_t)(size + offset), g_data.size());
 
     for (int32_t i(0); i < size; ++i)
-        EXPECT_EQ(nal[i], g_data[i]);
+        EXPECT_EQ(nal[i], g_data[i + offset]);
 
     EXPECT_FALSE(reader.read(nal, size));
 }
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -27,9 +27,12 @@
 
 #include <assert.h>
 #include <ctype.h>
+#include <sstream>
 #include <string.h>
 #include <sys/time.h>
 #include <va/va.h>
+#include <cstdlib>
+#include <limits>
 
 namespace YamiMediaCodec{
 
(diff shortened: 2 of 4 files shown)


common/nalreader_unittest.cpp
```diff
@@ -53,20 +53,20 @@ const std::array<uint8_t, 39> g_data = {
...
-    NalReader reader(&g_data[0], g_data.size(), false);
+    NalReader reader(&g_data[0], g_data.size(), 0, false);
 
     for (unsigned n(0); n < g_nsizes.size(); ++n) {
         EXPECT_TRUE(reader.read(nal, size));
         EXPECT_EQ(size, g_nsizes[n]);
         for (int32_t i(0); i < size; ++i)
-            EXPECT_EQ(nal[i], g_data[i+offset]);
-        offset += size;
+            EXPECT_EQ(nal[i], g_data[i + offset + 4]);
```
xuguangxin: @uartie ,may have more comments, but I think, 4, 3 are magic number, nobody know what it is could we use same macro for this?

common/utils_unittest.cpp
```diff
@@ -126,9 +126,9 @@ UTILS_TEST(guessResolutionOverflow) {
             << "x"
             << std::numeric_limits<long>::max();
 
-    EXPECT_FALSE(guessResolution(sstream.str().c_str(), w, h));
-    EXPECT_EQ(w, 0);
-    EXPECT_EQ(h, 0);
+    EXPECT_TRUE(guessResolution(sstream.str().c_str(), w, h));
```
xuguangxin: could we make w, h = 0 in this case? Since we will return 0 for "w3840h2160",  better return 0 if we can't get enough room for long.

xuguangxin (dismissed):

common/utils_unittest.cpp
wudping: hi @xuguangxin , do you mean that we should assign 0 to width, when it is greater than 3840, so as height ?

common/nalreader_unittest.cpp
wudping: hi @xuguangxin , @uartie, as you see, 4 is consist of 1 junk byte, and 3 start code bytes.

uartie: To make it more clear to reader... keep offset initialization at 1 (i.e. to skip junk byte) and add comment to this effect.  Define a constant start code size = 3.  Then use i + offset + start code size for indexing.  Finally, increment offset by size + start code size.

Add comment to test case about expectation... something like: "For unit reads, the NalReader returns buffer contents between each start code prefix; start code prefix is not included in read result."

common/nalreader_unittest.cpp
```diff
@@ -76,13 +76,13 @@ NALREADER_TEST(ReadAsWhole) {
...
     EXPECT_TRUE(reader.read(nal, size));
-    EXPECT_EQ((size_t)size, g_data.size());
+    EXPECT_EQ((size_t)(size + 4), g_data.size());
 
     for (int32_t i(0); i < size; ++i)
-        EXPECT_EQ(nal[i], g_data[i]);
+        EXPECT_EQ(nal[i], g_data[i + 4]);
 
     EXPECT_FALSE(reader.read(nal, size));
 }
```
uartie: Define `const int32_t offset = 4; // 1 initial junk byte + first start code`
Then replace 4's above with `offset`.

Add comment to test case about expectation, for example: "For whole read, the NalReader returns entire buffer contents after first start code.  That is, the first start code prefix is not included in the result."

common/utils_unittest.cpp
uartie: This test case should not be changed.  It exposes a bug in `guessResolution`.  Please fix `guessResolution` instead.  That is, `guessResolution` overflows the **int** output parameters (w,h) if it parses a width or height from `filename` param that is larger than what can fit in an `int` value.  This is the bug.

As @xuguangxin pointed out, `guessResolution` should avoid overflow and just make w,h = 0 instead.

uartie (changes requested):

wudping:
Updated according to the detail comments. Thanks.
Please help to review again.

common/utils.cpp
```diff
@@ -121,6 +121,11 @@ bool guessResolution(const char* filename, int& w, int& h)
             sscanf(tokStart, "%d", &h);
         }
     }
+    //"w" or "h" overflows the int, that is dealt with as a bug.
+    if (w < 0 || h < 0) {
```
uartie: What happens when you `sscanf` values in the range `long(std::numeric_limits<int>::max()) * 2 + 3`.. `long(std::numeric_limits<int>::max()) * 3 + 2` into an `int`?  This might be worth writing another unittest for.

Perhaps it's better to rewrite this function to use `std::strtol`, validate `errno` afterwards, then verify the result is in the range `0`..`std::numeric_limits<int>::max()` before assigning to w,h.

uartie (changes requested):
Perhaps split the `guessResolution` fixes into a separate pull-request until it's satisfactory.  This way we can merge `nalreader` fixes sooner.

wudping:
okay, @uartie , If guessResultion() pull-request is still with some problem, I will split it into other pull-request.

wudping:
Updated according to detail comments. Thanks @uartie .
Please help to review again, Thanks.

common/Makefile.am
```diff
@@ -30,6 +30,7 @@ libyami_common_ldflags = \
 	$(LIBVA_LIBS) \
 	$(LIBVA_DRM_LIBS) \
 	-lpthread \
+	-std=c++11 \
```
uartie: Remove this.  No need for c++11 in this patch series.  I'm pretty sure the libyami project does not approve c++11 yet since some current users may not have it available.

uartie (changes requested):

common/Makefile.am
wudping: Okay, @uartie.
But as I see it,  std::strtol()  requires c++11. 
If so, I will replace std::strtol with atol().

uartie: std::strtol does not requires c++11.  Only std::strtoll does.

http://en.cppreference.com/w/cpp/string/byte/strtol

wudping:
Separated unittest guessResolution() from this patch series. Please help to approve. Thanks.

wudping:
Re-integrate guessResolution() into this patch series, because there are not too many update. Thanks @uartie  for your suggestion.
Please help to review again, Thanks.

uartie (approved):

xuguangxin:
thansk @uartie for review

PR #737, opened 2017-04-27
vp8dec: enable svc-t decoding

wudping:
Support three layers vp8 svc-t decoding.

### DIFF
--- a/decoder/vaapidecoder_vp8.cpp
+++ b/decoder/vaapidecoder_vp8.cpp
@@ -513,6 +513,9 @@ YamiStatus VaapiDecoderVP8::decode(VideoDecodeBuffer* buffer)
             break;
         }
 
+        if (!targetTemporalFrame())
+            return YAMI_SUCCESS;
+
         if (m_frameHdr.key_frame == Vp8FrameHeader::KEYFRAME) {
             status = ensureContext();
             if (status != YAMI_SUCCESS)
@@ -564,4 +567,27 @@ YamiStatus VaapiDecoderVP8::decode(VideoDecodeBuffer* buffer)
     return status;
 }
 
+bool VaapiDecoderVP8::targetTemporalFrame()
+{
+    switch (m_configBuffer.temporalLayer) {
+    case 0: //decode all layers.
+        return true;
+    case 1:
+        if ((m_frameHdr.key_frame == Vp8FrameHeader::KEYFRAME) ||
+            m_frameHdr.refresh_last)
+            return true;
+        break;
+    case 2:
+        if ((m_frameHdr.key_frame == Vp8FrameHeader::KEYFRAME) ||
+            m_frameHdr.refresh_last || m_frameHdr.refresh_golden_frame)
+            return true;
+        break;
+    case 3:
+        return true;
+    default: //decode all layers.
+        return true;
+    }
+
+    return false;
+}
 }
--- a/interface/VideoDecoderDefs.h
+++ b/interface/VideoDecoderDefs.h
@@ -57,6 +57,10 @@ typedef struct {
     VAProfile profile;
     uint32_t flag;
     uint32_t fourcc;
+    //xxxLayer - how many layers to decode; if 0, decode all layers.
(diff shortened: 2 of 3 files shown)


wudping:

decoder/vaapidecoder_base.h
```diff
@@ -85,6 +85,8 @@ class VaapiDecoderBase:public IVideoDecoder {
 
     //do not use this, we will remove this in near future
     virtual VADisplay getDisplayID();
+
+    virtual bool isSVCTFrame() { return true; }
```
xuguangxin: Why do we need add this to the base?

decoder/vaapidecoder_vp8.cpp
```diff
@@ -564,4 +567,28 @@ YamiStatus VaapiDecoderVP8::decode(VideoDecodeBuffer* buffer)
     return status;
 }
 
+bool VaapiDecoderVP8::isSVCTFrame()
+{
+    switch (m_configBuffer.temporalLayer) {
+    case 0:
```
xuguangxin: Please comments about 0, why we need return true for 0 layer

decoder/vaapidecoder_vp8.cpp
```diff
@@ -564,4 +567,28 @@ YamiStatus VaapiDecoderVP8::decode(VideoDecodeBuffer* buffer)
...
+            return true;
+        break;
+    case 2:
+        if ((m_frameHdr.key_frame == Vp8FrameHeader::KEYFRAME) ||
+            m_frameHdr.refresh_last || m_frameHdr.refresh_golden_frame)
+            return true;
+        break;
+    case 3:
+        return true;
+    default:
```
xuguangxin: Please return true for default, if you have 3 layers, user want 4 layers, you need decode all 3 for them

interface/VideoDecoderDefs.h
```diff
@@ -57,6 +57,13 @@ typedef struct {
     VAProfile profile;
     uint32_t flag;
     uint32_t fourcc;
+    //xxxLayer - layer number, 0: decode all layers, >0: decode up to target layer.
```
xuguangxin: Please remove unsupport things. vp8 not support it, not mean all codecs not support it.

xuguangxin: decode up to the target layer.-> How many layers you want to decode.

xuguangxin (changes requested):

decoder/vaapidecoder_base.h
wudping: Because other codecs may support SVC-T.

wudping:
updated, thanks.

decoder/vaapidecoder_base.h
xuguangxin: Different decoder may handle SVC in different way, this not needed in base class

decoder/vaapidecoder_vp8.cpp
```diff
@@ -564,4 +567,27 @@ YamiStatus VaapiDecoderVP8::decode(VideoDecodeBuffer* buffer)
     return status;
 }
 
+bool VaapiDecoderVP8::isSVCTFrame()
```
xuguangxin: you can change this to isTargetFrame

wudping:
updated, thanks @xuguangxin .

decoder/vaapidecoder_base.h
```diff
@@ -85,6 +85,8 @@ class VaapiDecoderBase:public IVideoDecoder {
 
     //do not use this, we will remove this in near future
     virtual VADisplay getDisplayID();
+
+    virtual bool targetTemporalFrame() { return true; }
```
xuguangxin: Do not put this in base, it's codec specific feature

xuguangxin (changes requested):

wudping:
Updated, thanks @xuguangxin .

xuguangxin (approved):

xuguangxin (approved):

PR #752, opened 2017-06-28
accept size 0 byte stream in BitReader and NalReader

xuguangxin:
When we have some stream from the network, it may have 0 sizes for some chunk. We'd better support 0 sizes and report error later.

### DIFF
--- a/codecparsers/bitReader_unittest.cpp
+++ b/codecparsers/bitReader_unittest.cpp
@@ -73,4 +73,27 @@ BITREADER_TEST(NoOverflow)
     EXPECT_EQ(0u, reader.getRemainingBitsCount());
 }
 
+void checkBitreadEmpty(BitReader& reader)
+{
+    EXPECT_TRUE(reader.end());
+    EXPECT_EQ(0u, reader.getPos());
+    EXPECT_EQ(0u,
+        reader.getRemainingBitsCount());
+    bool b;
+    EXPECT_FALSE(reader.readT(b));
+    EXPECT_TRUE(reader.end());
+}
+
+BITREADER_TEST(NullInit)
+{
+    uint8_t data = 0;
+    BitReader r1(&data, 0);
+    checkBitreadEmpty(r1);
+
+    BitReader r2(NULL, 0);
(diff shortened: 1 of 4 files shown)


xuguangxin:
@uartie , please help review this. thanks

xuguangxin:
Hi @uartie . any update on this.
thanks

codecparsers/bitReader_unittest.cpp
```diff
@@ -73,4 +73,27 @@ BITREADER_TEST(NoOverflow)
     EXPECT_EQ(0u, reader.getRemainingBitsCount());
 }
 
+void checkBitread(BitReader& reader)
```
uartie: It looks like this function is only relevant for checking a reader with a null/empty bitstream.  Perhaps rename it to checkBitreadEmpty or checkBitreadNull for clarity.

codecparsers/nalReader_unittest.cpp
```diff
@@ -55,4 +55,31 @@ NALREADER_TEST(ReadBeyondBoundary)
     EXPECT_EQ(0, reader.readSe());
 }
 
+void checkBitread(NalReader& reader)
```
uartie: It looks like this function is only relevant for checking a reader with a null/empty bitstream.  Perhaps rename it to checkBitreadEmpty or checkBitreadNull for clarity.

uartie (commented):
Does the jpeg decoder handle an empty/null bitstream appropriately with this change?  Perhaps we need a unit test for that too, if we don't have one already.

uartie:
Also, I think it might be appropriate in this case to squash these commits to avoid introducing an intermediate unit test regression commit.

xuguangxin:
Thanks for the review.I will change the name

- Construct_InvalidParams will check the empty/null bitstream.
- Squash commits may introduce trouble when user just want cherry-pick some patch to the reversion. They usually not care about unit test. But they want changes in bit reader.

xuguangxin:
@uartie done

uartie (approved):
LGTM

Issue #707, opened 2017-02-17
libyami should pull in gtest source into a sub-folder


seanvk:
As we've learned from libva-utils and intel-vaapi-driver, it is far easier to simply include the gtest source in the project itself for purposes of the actual testing.

xuguangxin:
Thanks for suggestion. I will check libva-utils and see how to do this

xuguangxin:
I will send pull request about decoder API for vc1 and hevc. This change will conflict with the patch. I will do this change after decoder unit test merged.

Zhziyao:
Hi, @uartie @seanvk 
According to your suggestion, I have moved the unittest files(gtest source) from encoder directory to test directory. Please help review pull request #753. If it is OK, I will move all the other subdirs' unittest files of libyami to test directory.

uartie:
#753 does nothing to address this issue.  This issue is about adding the third-party google test source code (https://github.com/google/googletest) into this project.  See https://github.com/01org/intel-vaapi-driver/commit/1746adbb9adabd9fbee417ae3db67da599cc84e9 for an example.

Zhziyao:
Hi, @uartie 
Thank you for your kindness advice. I have added google test source code to this project. 
Please help review pull request #774 .  Thanks.

xuguangxin:
close this since #774 merged, thanks U.Artie and Sean

PR #753, opened 2017-06-28
unittest: move unittest files to test directory

Zhziyao:
make test directory under libyami
move unittest files from encoder directory to test directory

Signed-off-by: Zhziyao <ziyao.zhang@intel.com>


uartie (changes requested):
Why do you want to do this?  This has nothing to do with #707 ... #707 is about adding the third-party google test source code into the project.

uartie:
Test files do not need to be moved... just add the gtest source code and compile it as a convenience library.  Then link each unittest binary to that library.

xuguangxin:
@Zhziyao ,any update on this?

Issue #741, opened 2017-05-31
Need fix all gtest failure things.


xuguangxin:
Currently, we have 7 issues on SKL, we need to fix them.
CompatibleX11  
CreateX11  
Format_Change/0  
Format_Change/1  
ReadAsUnits  
ReadAsWhole 
guessResolutionOverflow

uartie:
For the X11 based tests, we should allow those to be disabled/enabled at runtime since not all test environments will have an X server running (e.g. servers).  We could do that by prefixing the test names with DISABLED_ (i.e. gtest framework skips [DISABLED_](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#temporarily-disabling-tests) prefixed tests by default, but can be enabled with command line)... or we can check an environment variable set by the user (e.g. YAMI_ENABLE_X11_TESTS=[0|1]).

The ReadAsUnits, ReadAsWhole and guessResolutionOverflow test cases are already fixed by #730... it just needs final review and merged.

xuguangxin:
Thanks for suggestion, we will look it.

after #730 merge.
We have 4 issue left

CompatibleX11  
CreateX11  
Format_Change/0  
Format_Change/1

uartie:
On SKL and BSW, we also have these failures in decode unittest:

Flush/3
Format_Change/3

...which aren't supported on these platforms and should be skipped at runtime.  This means we need a runtime support check feature.  Currently, those tests segfault on these platforms.

xuguangxin:
@uartie , it still needs more work to add run-time check in yami. Let use add some check in gtest first. I will send a pull request this week.

xuguangxin:
with #754, all issues fixed

PR #754, opened 2017-06-28
dec_unittest: fix vp9 decode unittest failure on BSW/SKL

wudping:
BSW/SKL do not support VP9 decoding, so vp9 decoding leads to core-dump
in decode unittest.
This patch will skip the unittest when detect that the platform dosen't
support VP9.

to fix issue [741](https://github.com/01org/libyami/issues/741)

### DIFF
--- a/decoder/DecoderApi_unittest.cpp
+++ b/decoder/DecoderApi_unittest.cpp
@@ -106,6 +106,14 @@ static FrameData g_jpegdata[] = {
     g_EOF,
 };
 
+static const std::string getFullTestName()
+{
+    const ::testing::TestInfo* const info = ::testing::UnitTest::GetInstance()->current_test_info();
+    return std::string(info->test_case_name())
+        + std::string(".")
+        + std::string(info->name());
+}
+
 class TestDecodeFrames {
 public:
     typedef SharedPtr<TestDecodeFrames> Shared;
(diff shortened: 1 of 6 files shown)


wudping:
please help to review.

decoder/DecoderApi_unittest.cpp
```diff
@@ -283,6 +291,12 @@ TEST_P(DecodeApiTest, Format_Change)
     while (frames.getFrame(buffer, info)) {
 
         YamiStatus status = decoder->decode(&buffer);
+        if (YAMI_UNSUPPORTED == status) {
```
xuguangxin: duplicate code here. 
please use
bool checkStatus(YamiStatus status)
to  remove all duplicates

decoder/DecoderApi_unittest.cpp
```diff
@@ -283,6 +291,12 @@ TEST_P(DecodeApiTest, Format_Change)
...
+            RecordProperty("skipped", true);
+            std::cout << "[  SKIPPED ] " << getFullTestName()
+                      << " We can't test this in terminal, the stream is unsupported." << std::endl;
```
xuguangxin: change this to "hw do not support this decoder"

decoder/DecoderApi_unittest.cpp
```diff
@@ -297,7 +311,14 @@ TEST_P(DecodeApiTest, Format_Change)
             allocator->onFormatChange(format);
 
             //send buffer again
-            EXPECT_EQ(YAMI_SUCCESS, decoder->decode(&buffer));
+            status = decoder->decode(&buffer);
+            if (YAMI_UNSUPPORTED == status) {
```
xuguangxin, line 309: no call decode call need this, we just need check it in first decode call

xuguangxin (changes requested):
Please split core changes and unit test changes

decoder/DecoderApi_unittest.cpp
wudping: hi @xuguangxin ,
This is necessary actually. Because Yami will return YAMI_DECODE_FORMAT_CHANGE when decoding the first frame.

xuguangxin (approved):

PR #778, opened 2017-08-01
Svc

LindaYu17:


xuguangxin:
@wudping , please help review this.
Could we add some test case in daily regression to capture svct issues?

v4l2/v4l2_encode.cpp
```diff
@@ -58,6 +58,14 @@ bool V4l2Encoder::start()
     streamFormat.streamFormat = AVC_STREAM_FORMAT_ANNEXB;
     m_encoder->setParameters(VideoConfigTypeAVCStreamFormat, &streamFormat);
 
+    if (m_svct) {
+        VideoTemporalLayers layers;
+        layers.numLayersMinus1 = 1; // It is two layers actually
+        layers.bitRate[0] = 3000000;
+        layers.bitRate[1] = 4000000;
```
wudping: The highest layer's bitrate should been set via "VideoParamsCommon.rcParams.bitRate"; So this line should be replaced with:
VideoParamsCommon.rcParams.bitRate = 4000000;

LindaYu17:
@wudping updated!

v4l2/v4l2_encode.cpp
```diff
@@ -195,6 +203,16 @@ bool V4l2Encoder::giveOutputBuffer(struct v4l2_buffer *dqbuf)
     if (outputBuffer->flag & ENCODE_BUFFERFLAG_SYNCFRAME)
         dqbuf->flags = V4L2_BUF_FLAG_KEYFRAME;
 
+
+    printf("outputBuffer->temporalId = %d\n", outputBuffer->temporalID);
```
xuguangxin: please remove this print?

wudping:
+1

wudping:
+1

v4l2/v4l2_encode.cpp
```diff
@@ -58,6 +58,14 @@ bool V4l2Encoder::start()
     streamFormat.streamFormat = AVC_STREAM_FORMAT_ANNEXB;
     m_encoder->setParameters(VideoConfigTypeAVCStreamFormat, &streamFormat);
 
+    if (m_svct) {
+        VideoTemporalLayers layers;
+        layers.numLayersMinus1 = 1; // It is two layers actually
+        layers.bitRate[0] = 3000000;
```
xuguangxin: This hard coded bitrate may introduce issue on dynamic bitrate changes. Could we change logical like this:
if total birate is a, the zero layer bitrate is a/2 etc.

xuguangxin (changes requested):

v4l2/v4l2_encode.cpp
LindaYu17: do you mean this?
layers.bitRate[0] = m_videoParams.rcParams.bitRate / 2;
layers.bitRate[1] = m_videoParams.rcParams.bitRate;

but if we don't set m_videoParams.rcParams.bitRate, how to set the layer bitrate？

wudping: In my opinion, m_videoParams.rcParams.bitRate is what we need to set here.

wudping: How about set a default bitrate according to the "m_videoParams.resolution" in case m_videoParams.rcParams.bitRate == 0. As far as I know, 4 Mb/s for 1080P.

v4l2/v4l2_encode.cpp
```diff
@@ -58,6 +58,15 @@ bool V4l2Encoder::start()
     streamFormat.streamFormat = AVC_STREAM_FORMAT_ANNEXB;
     m_encoder->setParameters(VideoConfigTypeAVCStreamFormat, &streamFormat);
 
+    if (m_svct) {
+        VideoTemporalLayers layers;
+        layers.numLayersMinus1 = 1; // It is two layers actually
+        layers.bitRate[0] = m_videoParams.rcParams.bitRate / 2;
+        layers.bitRate[1] = m_videoParams.rcParams.bitRate;
```
wudping, line 65: Hi @LindaYu17 , 
We can delete the line "layers.bitRate[1] = m_videoParams.rcParams.bitRate;", because the second layer is the highest layer in this case, and whose bitrate is assigned by "m_videoParams.rcParams.bitRate" in vaapiencoder_xxx.cpp file.

LindaYu17: layers.bitRate[1] = m_videoParams.rcParams.bitRate;
this assignment make the code more clear and it is not conflict with that in libyami either, so suggest to keep this line.

xuguangxin:
@wudping , is it ok to merge this?

wudping:
@xuguangxin , it's okay for me.

xuguangxin:
@wudping this will break yami-utils since a API change, right? could you prepare a patch for yami-utils?

wudping:
@xuguangxin , okay, I am planing to add a patch for this update next week. But SVC-T regression will be latter.

wudping:
hi @xuguangxin, The commit of interface modifying the typo "VideoTemproalLayers" does not impact libyami-utils.

xuguangxin:
@wudping , why? this means the temporal layer never set in yami-utils?

wudping:
Hi @xuguangxin , 
typedef struct VideoParamsCommon {
    uint32_t size;
    VAProfile profile;
    uint8_t level;
    VideoResolution resolution;
    VideoFrameRate frameRate;
    **VideoTemproalLayers temporalLayers;**
libyami-itils sets temporalLayers via VideoParamsCommon, and yami-utils needn't to define a new variable with VideoTemproalLayers type, so we just need to modify the file VideoEncoderDefs.h in libyami.

Anyway, I am sorry for this typo. It's my fault.

xuguangxin:
ok thanks for explain

xuguangxin (approved):

PR #781, opened 2017-08-08
Enhance short term refs set update for multiple refs hevc encoding

### LINKED ISSUE #760: Implement the Multi Reference Frame for H265 encoder

pengche1:
Fixes #760

Signed-off-by: peng.chen <peng.c.chen@intel.com>

### DIFF
--- a/encoder/vaapiencoder_hevc.cpp
+++ b/encoder/vaapiencoder_hevc.cpp
@@ -1290,7 +1290,7 @@ void VaapiEncoderHEVC::shortRfsUpdate(const PicturePtr& picture)
     m_shortRFS.inter_ref_pic_set_prediction_flag = 0;
 
     if (intraPeriod() > 1 && m_refList0.size()) {
-        m_shortRFS.num_negative_pics         = 1;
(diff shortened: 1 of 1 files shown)


encoder/vaapiencoder_hevc.cpp
```diff
@@ -1290,7 +1290,7 @@ void VaapiEncoderHEVC::shortRfsUpdate(const PicturePtr& picture)
     m_shortRFS.inter_ref_pic_set_prediction_flag = 0;
 
     if (intraPeriod() > 1 && m_refList0.size()) {
-        m_shortRFS.num_negative_pics         = 1;
+        m_shortRFS.num_negative_pics         = m_refList0.size();
```
xuguangxin, line 1293: this wired, why we need set the reference frame both a negative and positive.  could we just set it as positive?

pengche1: yes, it is my fault, it should be only set for negative pics. You can discard my patch and fix it with another patch

xuguangxin: ok, please help send other patch

PR #783, opened 2017-08-17
vp8: add virutal dtor to avoid memory leak in destruction

LindaYu17:


xuguangxin (approved):

chivakker (commented):
:+1:

PR #784, opened 2017-08-18
JPEG:exit parse when no more data

LindaYu17:

### DIFF
--- a/codecparsers/jpegParser.cpp
+++ b/codecparsers/jpegParser.cpp
@@ -308,6 +308,9 @@ bool Parser::firstMarker()
 
 bool Parser::nextMarker()
 {
+    if (m_input.getRemainingBitsCount() == 0)
+        return false;
+
     const uint8_t* const end = m_data + m_size - 1;
     const uint8_t* const start = m_data + currentBytePosition();
     const uint8_t* current = start;


LindaYu17:
@xuguangxin , please help to review this patch.
I found in Android env, 
320         match = std::find(current, end, 0xFF);
coredump happens when end < current, so add check before std::find().

codecparsers/jpegParser.cpp
```diff
@@ -313,6 +313,9 @@ bool Parser::nextMarker()
     const uint8_t* current = start;
     const uint8_t* match;
```
uartie: Maybe just assign`const uint8_t* const start =  m_data + std::min(currentBytePosition(), m_size - 1);` at line 312 instead?

xuguangxin: hi @uartie , if m_size = 0, the start will be m_data - 1. 
Could we adjust code just like
>const uint8_t* const end = m_data + m_size;

we only need change following lines
>if (match == end || (match +1) == end)
>      return false;

uartie: @xuguangxin good point.

Unfortunately, @LindaYu17 says the coredump happens before this conditional when "end < current"... so this check would come too late.

@LindaYu17, does the issue only happen when m_size == 0?  That seems to be the only way "end < current" can occur (assuming currentBytePosition() is always < m_size).  If so, perhaps the following will fix it at the source of the problem:

```
const uint32_t epos = (m_size > 0 ? m_size - 1 : 0);
const uint8_t* const end = m_data + epos;
```

@LindaYu17 does Construct_InvalidParams unit test in jpegParser_unittest.cpp not detect your issue?  If not, can you write a unit test to catch this bug, please?

LindaYu17: hi @uartie this is not corner case in Android, it happens in any jpeg decode.
when the issue happens, currentBytePosition() = m_size = jpeg size.

uartie: OK, I see now... `currentBytePosition() < m_size` is a bad assumption.

So all we need is
```
if (m_input.getRemainingBitsCount() < 8)
  return false;
```
at top of function.

xuguangxin: yeah, this good.

LindaYu17:
updated!

uartie (approved):

xuguangxin:
thanks, Linda and U.Artie

PR #788, opened 2017-09-05
vaapi: ERROR if vaCreateConfig fails

### LINKED ISSUE #443: provide an error message when Profile or Entrypoint are not supported
libyami client should have a meaningful error message displayed when either the Profile or Entrypoint are not supported.  User should realize that the particular generation of hardware doesn't support a Profile or Entrypoint that is going to be used. .

stripes416:
Also correct grammar in nearby error message.

Fixes #443
Signed-off-by: Kelly Ledford <kelly.ledford@intel.com>

### DIFF
--- a/vaapi/vaapicontext.cpp
+++ b/vaapi/vaapicontext.cpp
@@ -24,6 +24,7 @@
 #include "common/common_def.h"
 #include "vaapi/vaapidisplay.h"
 #include "vaapi/VaapiUtils.h"
+#include "vaapi/vaapistreamable.h"
 #include <algorithm>
 #include <vector>
 
@@ -84,14 +85,17 @@ YamiStatus VaapiConfig::create(const DisplayPtr& display,
     VAConfigID config;
 
     if (!checkProfileCompatible(display, profile)){
-        ERROR("Unsupport profile");
+        ERROR("Unsupported profile %s.\n", toString(profile).c_str());
         return YAMI_UNSUPPORTED;
     }
 
     vaStatus = vaCreateConfig(display->getID(), profile, entry, attribList, numAttribs, &config);
 
-    if (!checkVaapiStatus(vaStatus, "vaCreateConfig "))
+    if (!checkVaapiStatus(vaStatus, "vaCreateConfig ")) {
+        ERROR("Unable to create config for profile: %s and entrypoint: %s.\n",
+              toString(profile).c_str(), toString(entry).c_str());
         return YAMI_FAIL;
+    }
     confg.reset(new VaapiConfig(display, config));
     return YAMI_SUCCESS;
 }
--- a/vaapi/vaapistreamable.h
+++ b/vaapi/vaapistreamable.h
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2017 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef vaapistreamable_h
+#define vaapistreamable_h
+#include <va/va.h>
+#include <sstream>
+
+template <typename T>
+const std::string toString(const T& t)
+{
+    std::ostringstream os;
+    os << t;
+    return os.str();
+}
+
+
+inline std::ostream&
+operator<<(std::ostream& os, const VAProfile& profile)
+{
+    switch(profile) {
+    case VAProfileNone:
+        return os << "VAProfileNone";
+    case VAProfileMPEG2Simple:
+        return os << "VAProfileMPEG2Simple";
+    case VAProfileMPEG2Main:
+        return os << "VAProfileMPEG2Main";
+    case VAProfileMPEG4Simple:
+        return os << "VAProfileMPEG4Simple";
+    case VAProfileMPEG4AdvancedSimple:
+        return os << "VAProfileMPEG4AdvancedSimple";
+    case VAProfileMPEG4Main:
+        return os << "VAProfileMPEG4Main";
+    case VAProfileVC1Simple:
+        return os << "VAProfileVC1Simple";
+    case VAProfileVC1Main:
+        return os << "VAProfileVC1Main";
+    case VAProfileVC1Advanced:
+        return os << "VAProfileVC1Advanced";
+    case VAProfileH263Baseline:
+        return os << "VAProfileH263Baseline";
+    case VAProfileJPEGBaseline:
+        return os << "VAProfileJPEGBaseline";
+    case VAProfileVP8Version0_3:
+        return os << "VAProfileVP8Version0_3";
+    case VAProfileHEVCMain:
+        return os << "VAProfileHEVCMain";
+    case VAProfileHEVCMain10:
+        return os << "VAProfileHEVCMain10";
+    case VAProfileVP9Profile0:
+        return os << "VAProfileVP9Profile0";
+    case VAProfileVP9Profile1:
+        return os << "VAProfileVP9Profile1";
+    case VAProfileVP9Profile2:
+        return os << "VAProfileVP9Profile2";
+    case VAProfileVP9Profile3:
+        return os << "VAProfileVP9Profile3";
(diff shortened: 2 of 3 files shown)


vaapi/vaapicontext.cpp
```diff
@@ -90,14 +90,16 @@ YamiStatus VaapiConfig::create(const DisplayPtr& display,
...
-        ERROR("Unsupport profile");
+        ERROR("Unsupported profile.");
         return YAMI_UNSUPPORTED;
     }
 
     vaStatus = vaCreateConfig(display->getID(), profile, entry, attribList, numAttribs, &config);
 
-    if (!checkVaapiStatus(vaStatus, "vaCreateConfig "))
+    if (!checkVaapiStatus(vaStatus, "vaCreateConfig ")) {
+	ERROR("Unable to create config. This may be due to an unsupported entrypoint.");
```
chivakker: What about saying: "Unable to create config for profile %s and entrypoint %s".  

At this point of execution you know which entrypoint and profile were requested. Then user can run vainfo to confirm the codec is supported.

xuguangxin: maybe we can return YAMI_UNSUPPORTED here too

stripes416:
Sounds good. I'll make those changes.

vaapi/vaapicontext.cpp
stripes416: chivakker, I have some questions about how to go about what you requested.

I don't see a good way to converting the profile / entrypoint to a string with libva or libyami. However, libva-utils/vainfo/vainfo.c does contain such code.

I could copy-and-paste the code from vainfo.c into this file or yamiinfo.cpp, but that seems a little distasteful from a maintenance perspective. Alternatively I could move the code to libva/va/va.c and use it from both projects. There's a function with a similar purpose already in va.c: vaErrorStr( ). However, doing so would add another API to libva which we may not want.

What do you think?

chivakker: @stripes416 IIRC, there should be a ProfileToString utility function and EntrypointToString.  I think it might be inside the test folder.  You could probably copy them instead.

uartie: I'd prefer to see C++ style string conversions used... see "toString()" and stream operators from intel-vaapi-driver/test/i965_streamable.h.  It's ok to copy that code into this project (e.g. add "vaapi/vaapistreamable.h").

uartie: @xuguangxin returning YAMI_UNSUPPORTED only makes sense when vaStatus is VA_STATUS_ERROR_UNSUPPORTED_PROFILE or VA_STATUS_ERROR_UNSUPPORTED_ENTRYPOINT, right?  Otherwise, we should return YAMI_FAIL?

vaapi/vaapicontext.cpp
```diff
@@ -90,14 +90,16 @@ YamiStatus VaapiConfig::create(const DisplayPtr& display,
     VAConfigID config;
 
     if (!checkProfileCompatible(display, profile)){
-        ERROR("Unsupport profile");
+        ERROR("Unsupported profile.");
```
uartie: In addition, please print the profile here.

vaapi/vaapicontext.cpp
xuguangxin: @uartie , yeah, you are right

xuguangxin:
@stripes416 any update on this?

uartie:
LibVA 2.0 has been merged to master.  This PR needs to be rebased on top of libyami's libva 2.0 compatibility commits... please rebase to fix intel/ci builds.

vaapi/vaapicontext.cpp
```diff
@@ -24,6 +24,7 @@
 #include "common/common_def.h"
 #include "vaapi/vaapidisplay.h"
 #include "vaapi/VaapiUtils.h"
+#include <va/va_str.h>
```
uartie: We can't include this unconditionally since libyami supports both libva 1.x and libva 2.0

vaapi/vaapicontext.cpp
```diff
@@ -84,14 +85,17 @@ YamiStatus VaapiConfig::create(const DisplayPtr& display,
     VAConfigID config;
 
     if (!checkProfileCompatible(display, profile)){
-        ERROR("Unsupport profile");
+        ERROR("Unsupported profile: %s.", vaProfileStr(profile));
```
uartie: va...Str() methods only available in libva 2.0

vaapi/vaapicontext.cpp
```diff
@@ -84,14 +85,17 @@ YamiStatus VaapiConfig::create(const DisplayPtr& display,
...
         return YAMI_UNSUPPORTED;
     }
 
     vaStatus = vaCreateConfig(display->getID(), profile, entry, attribList, numAttribs, &config);
 
-    if (!checkVaapiStatus(vaStatus, "vaCreateConfig "))
+    if (!checkVaapiStatus(vaStatus, "vaCreateConfig ")) {
+        ERROR("Unable to create config for profile: %s and entrypoint: %s.",
+              vaProfileStr(profile), vaEntrypointStr(entry));
```
uartie: same as above

uartie (changes requested):

vaapi/vaapicontext.cpp
```diff
@@ -24,8 +24,10 @@
 #include "common/common_def.h"
 #include "vaapi/vaapidisplay.h"
 #include "vaapi/VaapiUtils.h"
+#include "vaapi/vaapistreamable.h"
 #include <algorithm>
 #include <vector>
+#include <sstream>
```
uartie: This is already included indirectly via vaapistreamable.h.  It is not even needed in this unit if you use toString(...) instead (which you just added in vaapistreamable.h).

vaapi/vaapicontext.cpp
```diff
@@ -84,14 +86,21 @@ YamiStatus VaapiConfig::create(const DisplayPtr& display,
     VAConfigID config;
 
     if (!checkProfileCompatible(display, profile)){
-        ERROR("Unsupport profile");
+        std::stringstream msg;
+        msg << "Unsupported profile: " << profile << std::endl;
+        ERROR("%s", msg.str().c_str());
```
uartie: `ERROR("Unsupported profile: %s", toString(profile).c_str());` would be simpler.

vaapi/vaapicontext.cpp
```diff
@@ -84,14 +86,21 @@ YamiStatus VaapiConfig::create(const DisplayPtr& display,
...
     }
 
     vaStatus = vaCreateConfig(display->getID(), profile, entry, attribList, numAttribs, &config);
 
-    if (!checkVaapiStatus(vaStatus, "vaCreateConfig "))
+    if (!checkVaapiStatus(vaStatus, "vaCreateConfig ")) {
+        std::stringstream msg;
+        msg << "Unable to create config for profile: " << profile
+            << " and entrypoint: " << entry << std::endl;
+        ERROR("%s", msg.str().c_str());
```
uartie: This is what toString(...) is for... so you don't have to do these extra steps before the logging macros.

vaapi/vaapistreamable.h
```diff
@@ -0,0 +1,124 @@
...
+ * of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT.
+ * IN NO EVENT SHALL INTEL AND/OR ITS SUPPLIERS BE LIABLE FOR
+ * ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
```
uartie: Please use the Apache 2.0 license like the other files.

uartie (changes requested):

vaapi/vaapicontext.cpp
uartie: Since checkVaapiStatus(vaStatus, "vaCreateConfig") already prints "vaCreateConfig: the requested VAEntryPoint is not supported", perhaps just do:

```
ERROR("   VAProfile: %s", toString(profile).c_str());
ERROR("VAEntrypoint: %s", toString(entry).c_str());
```

...to cut down on the error message redundancy.

uartie: or maybe:

```
ERROR("%s, %s", toString(profile).c_str(), toString(entry).c_str());
```

uartie (approved):

xuguangxin:
Hi @stripes416 ,
this great! 

thanks all Daniel and U.Artie for review.

PR #790, opened 2017-09-19
Hevc enc based on media studio VA

wudping:
Enable HEVC encoding based on VPG driver.


wudping:
Please help to review, thanks a lot in advance.

wudping:
error commit

PR #799, opened 2017-09-27
decoder/base: always clear output buffer on destruct

uartie:
Decoded frames in the output buffer hold a reference to Surface
objects.  If the output buffer is not flushed before vaTerminate
is called (which happens because DisplayPtr goes away), then this
causes vaDestroySurface to fail (seg fault) when that output buffer
is flushed and surfaces are destroyed.   This can happen during
decoder destruction where user does not explicitly flush output.


xhaihao (approved):
LGTM

xuguangxin (approved):

xuguangxin:
thanks, U.Artie and Haihao

PR #801, opened 2017-09-29
jpegd: attempt to decode images even if EOI is missing

uartie:
By standard, a missing EOI technically means the image
is corrupt.  However, there are many images in the wild where
a missing EOI is the only corruption to the image (i.e. the
entropy-coded data and headers are still intact).  This can
happen, for example, when the image was encoded by a buggy
application that forgets to append the EOI to the final
image.  If a missing EOI is the only issue, then decoders can
still successfully decode it.  However, a missing EOI does
not necessarily mean the rest of the image is intact.  That
is, there is no guarantee that the final decoded output will
produce the desired result.  Either way, we insert a fake
EOI marker here to give decoders a chance to decode the
image.

VIZ-8219

### DIFF
--- a/codecparsers/jpegParser.cpp
+++ b/codecparsers/jpegParser.cpp
@@ -237,6 +237,7 @@ Parser::Parser(const uint8_t* data, const uint32_t size)
     , m_callbacks()
     , m_sawSOI(false)
     , m_sawEOI(false)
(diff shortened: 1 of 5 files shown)


xuguangxin (approved):

xuguangxin:
thanks U.Artie

PR #807, opened 2017-11-02
configure.ac: bump version to 1.3.0

YuJiankang:
bump version to 1.3.0
Signed-off-by: jkyu <jiankang.yu@intel.com>


uartie (approved):
LGTM

uartie:
In the future, since you're maintaining the releases I don't think it's necessary to do this kind of change via a PR... rather just fix it directly.

YuJiankang:
Artie, thanks

PR #808, opened 2017-11-07
v4l2_decode:alloc 1 frame for generating thumbnail

LindaYu17:
only one surface is needed for generating thumbnail, enable low latency
mode for this purpose.

Change-Id: Ia8c19c1d950706860884ce036c1fdd0e9f353f90


v4l2/v4l2_decode.cpp
```diff
@@ -857,6 +862,12 @@ void V4l2Decoder::getInputJob()
         if (m_output->isAllocationDone()) {
             m_state = kWaitSurface;
             DEBUG("early out, no surface");
+            if (m_isThumbnailMode) {
```
xuguangxin: Why we need this?  In thubmail mode, we will decode the first frame then get output. then it will decode the second frame, and get YAMI_DECODE_NO_SURFACE.  So we do not need do special thing for thumbnail mode here

LindaYu17: Yes, the first frame is decoded and output. but we still do not have chance to take it to v4l2, and not have chance to setDeviceEvent() to notify app either. If it enters kWaitSurface, it will block here.
but yes, consumeInput() is not mandatory. @xuguangxin

xuguangxin (approved):

PR #815, opened 2017-11-15
use env var for device file to open

### LINKED ISSUE #809: Add support of choosing device for executing
libyami doesn't support choosing a device for executing. It uses a default device.

jsorg71:
Fixes #809


uartie (approved):
LGTM. Thanks!

PR #820, opened 2017-12-25
Vpp rotation and fix color balance bug

wudping:
1, add function of YUV image clockwise rotation;
2, fix the color-balance bug of only supporting to set one of those attributes: HUE, SATURATION, BRIGHTNESS, CONTRAST. Now it supports to set several of those attributes.

### DIFF
--- a/vpp/vaapipostprocess_scaler.cpp
+++ b/vpp/vaapipostprocess_scaler.cpp
@@ -65,6 +65,7 @@ VaapiPostProcessScaler::VaapiPostProcessScaler()
 {
     m_denoise.level = DENOISE_LEVEL_NONE;
     m_sharpening.level = SHARPENING_LEVEL_NONE;
+    m_transform = VPP_TRANSFORM_NONE;
 }
 
 bool VaapiPostProcessScaler::getFilters(std::vector<VABufferID>& filters)
@@ -125,8 +126,22 @@ VaapiPostProcessScaler::process(const SharedPtr<VideoFrame>& src,
         vppParam->filters = &filters[0];
         vppParam->num_filters = (unsigned int)filters.size();
     }
+    vppParam->rotation_state = mapToVARotationState(m_transform);
 
     return picture.process() ? YAMI_SUCCESS : YAMI_FAIL;
+}
+uint32_t VaapiPostProcessScaler::mapToVARotationState(VppTransform vppTransform)
+{
+    switch (vppTransform) {
+    case VPP_TRANSFORM_ROT_90:
+        return VA_ROTATION_90;
+    case VPP_TRANSFORM_ROT_180:
+        return VA_ROTATION_180;
+    case VPP_TRANSFORM_ROT_270:
+        return VA_ROTATION_270;
+    default:
+        return VA_ROTATION_NONE;
+    }
 }
 
 bool VaapiPostProcessScaler::mapToRange(
@@ -394,7 +409,15 @@ VaapiPostProcessScaler::setParameters(VppParamType type, void* vppParam)
 
         return setColorBalanceParam(*colorbalance);
     }
+    else if (type == VppParamTypeTransform) {
+        VppParamTransform* param = (VppParamTransform*)vppParam;
+        if (param->size != sizeof(VppParamTransform)) {
+            return YAMI_INVALID_PARAM;
+        }
+        m_transform = (VppTransform)param->transform;
+
+        return YAMI_SUCCESS;
+    }
     return VaapiPostProcessBase::setParameters(type, vppParam);
 }
-
 }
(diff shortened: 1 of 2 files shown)


wudping:

vpp/vaapipostprocess_scaler.cpp
```diff
@@ -87,6 +88,18 @@ bool VaapiPostProcessScaler::getFilters(std::vector<VABufferID>& filters)
     return !filters.empty();
 }
 
+void VaapiPostProcessScaler::setRotationState(VAProcPipelineParameterBuffer* vppParam)
+{
+    if (m_transform != VPP_TRANSFORM_NONE) {
```
xuguangxin: We do not need this function.

L93 to L101 equals to one line
 vppParam->rotation_state = mapToVARotationState(m_transform);
You can add it to L147

vpp/vaapipostprocess_scaler.cpp
```diff
@@ -369,6 +398,15 @@ VaapiPostProcessScaler::setParameters(VppParamType type, void* vppParam)
 
         return setColorBalanceParam(*colorbalance);
     }
+    else if (type == VppParamTypeTransform) {
```
xuguangxin, line 412: Could you add get function too?

vaapi/VaapiBuffer.cpp
```diff
@@ -52,6 +52,27 @@ BufObjectPtr VaapiBuffer::create(const ContextPtr& context,
     return buf;
 }
 
+BufObjectPtr VaapiBuffer::create(const ContextPtr& context,
```
xuguangxin: full of duplicated code. Instead of pass num to this create function, could we just pass size*num to the original version of create? or at least original version of creat function can call this version of create like this:

create(context, type, size, data, 1)

vpp/vaapipostprocess_scaler.h
```diff
@@ -49,16 +49,21 @@ class VaapiPostProcessScaler : public VaapiPostProcessBase {
         BufObjectPtr filter; //send to va;
     };
 
-    struct ColorBalanceParam {
-        BufObjectPtr filter; //send to va;
-        int32_t level; //send to va
+    struct ColorBalanceProperty {
```
xuguangxin: this is useless changes, it will change many lines, but no benefits
Could we just keep the original name? and add m_colorBalanceFilter as a member of VaapiPostProcessScaler

xuguangxin (changes requested):
Hi @wudping , please do not mix two things here. the rotation not related to color balance things. 
Please submit two pull request.
1. enable rotation
2. fix the color balance things.
thanks

vaapi/VaapiBuffer.cpp
wudping: thanks @xuguangxin , I will modify the code and double-check its function.
but here, one more explaination for this piece of code:
I follow the code in [va-uilts](https://github.com/01org/libva-utils/blob/master/videoprocess/vavpp.cpp#L1273)
`va_status = vaCreateBuffer(va_dpy, context_id,
                               VAProcFilterParameterBufferType, sizeof(color_balance_param), **4**,
                               color_balance_param, &color_balance_param_buf_id);`
I am not sure if using **size*num** instead of this piece of code will be a potential bug.

vpp/vaapipostprocess_scaler.h
wudping: thanks @xuguangxin , But if we pick filter out from ColorBalanceParams, that will harm the structure style, Because  the similar structure DeinterlaceParams and ProcParams all contain filter, and there is no single filter member varible in this class too.

vaapi/VaapiBuffer.cpp
wudping: sorry about that @xuguangxin , you are right, I will modify the function like that:
    static BufObjectPtr **create**(const ContextPtr&,
        VABufferType type,
        uint32_t size,
        const void* data = 0,
        void** mapped = 0,
        **uint32_t num = 1);**

**size*num** doesn't work in this case, so add one format parameter **uint32_t num = 1**.

vpp/vaapipostprocess_scaler.cpp
wudping: sorry about that @xuguangxin , I omitted that. 
It should be added with set function, but the interface "IVideoPostProcess" doesn't define it, and the other VPP attributes have not implemented getParameter() function, so how about to add this function in the next pull-request with other VPP attributes?

wudping:
hi @xuguangxin , updated, this pull-request only for rotation. 
please help to review again, thanks.

vpp/vaapipostprocess_scaler.cpp
xuguangxin: then please remove get function....sorry for this

wudping:
hi @xuguangxin, no problem;
updated, getParameters() has been removed, please help to review again, thanks.

xuguangxin (approved):

xuguangxin:
thanks dongping

PR #821, opened 2017-12-29
Vpp color balance multi-parameter setting

wudping:
This pull-request is suppose to fix the bug that vpp can not set one more parameter of HUE, SATURATION, BRIGHTNESS and CONTRAST.


wudping:

PR #835, opened 2018-03-21
Memset fixed during compiling with enabling gtest

wangzj0601:
When enable gtest during compiling libyami on Fedora28, it will also occur memset issue.

### DIFF
--- a/codecparsers/h264Parser_unittest.cpp
+++ b/codecparsers/h264Parser_unittest.cpp
@@ -147,7 +147,6 @@ namespace H264 {
         void checkH264Pps(Parser& parser, const NalUnit* nalu)
         {
             SharedPtr<PPS> pps(new PPS());
-            memset(pps.get(), 0, sizeof(PPS));
 
             ASSERT_EQ(NAL_PPS, nalu->nal_unit_type);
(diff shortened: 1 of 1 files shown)


xuguangxin:
@wangzj0601 ,there are two commits in your pull request.  I think the commit 72322e2 already merged, could you rebase your code?
thanks

codecparsers/h264Parser_unittest.cpp
```diff
@@ -173,7 +172,7 @@ namespace H264 {
         void checkH264SliceHeader(Parser& parser, NalUnit* nalu)
         {
             SharedPtr<SliceHeader> slice(new SliceHeader);
-            memset(slice.get(), 0, sizeof(SliceHeader));
+            *slice = SliceHeader();
```
xuguangxin: why we need this if SliceHeader has a constructor. the member already inited by "new SliceHeader"

xuguangxin (changes requested):

codecparsers/h264Parser_unittest.cpp
wangzj0601: @xuguangxin , I have removed the line of code.

uartie:
Fixes #845

PR #836, opened 2018-03-23
fix bug https://github.com/intel/libyami/issues/833

wangzj0601:
fix overflow issue on 32bit OS

### DIFF
--- a/codecparsers/mpeg2_parser.cpp
+++ b/codecparsers/mpeg2_parser.cpp
@@ -32,6 +32,7 @@
 // config.h defines macros for log.h to define log levels
 #include "common/log.h"
 #include "mpeg2_parser.h"
+#include <inttypes.h>
 
 namespace YamiParser {
 namespace MPEG2 {
@@ -255,7 +256,7 @@ namespace MPEG2 {
             return false;
 
 
-        DEBUG("slice header size                  : %ld",
+        DEBUG("slice header size                  : %" PRIu64 "",
               m_slice.sliceHeaderSize);
         DEBUG("slice number                       : %x",
               m_slice.verticalPosition);
@@ -269,7 +270,7 @@ namespace MPEG2 {
               m_slice.intra_slice);
         DEBUG("extra_bit_slice                    : %x",
               m_slice.extra_bit_slice);
-        DEBUG("slice size                         : %ld",
+        DEBUG("slice size                         : %" PRIu64 "",
               m_slice.sliceDataSize);
         DEBUG("size left on buffer                : %d", nalSize);
         DEBUG("slice data                         : %p",
--- a/decoder/vaapidecoder_mpeg2.cpp
+++ b/decoder/vaapidecoder_mpeg2.cpp
@@ -24,6 +24,7 @@
 #include "vaapidecoder_mpeg2.h"
 
 #include <string.h>
+#include <inttypes.h>
 
 using YamiParser::MPEG2::Slice;
 using YamiParser::MPEG2::PictureCodingExtension;
@@ -134,7 +135,7 @@ VaapiDecoderMPEG2::DPB::insertPictureToReferences(const PicturePtr& picture)
 
 YamiStatus VaapiDecoderMPEG2::DPB::insertPicture(const PicturePtr& picture)
 {
-    INFO("insertPicture to DPB size %lu", m_referencePictures.size());
+    INFO("insertPicture to DPB size %zu", m_referencePictures.size());
 
     YamiStatus status = YAMI_SUCCESS;
 
@@ -172,7 +173,7 @@ YamiStatus VaapiDecoderMPEG2::DPB::insertPicture(const PicturePtr& picture)
         }
     }
 
-    DEBUG("insertPicture returns dpb size %lu", m_referencePictures.size());
+    DEBUG("insertPicture returns dpb size %zu", m_referencePictures.size());
     return status;
 }
 
@@ -887,7 +888,7 @@ YamiStatus VaapiDecoderMPEG2::decode(VideoDecodeBuffer* buffer)
     m_stream->time_stamp = buffer->timeStamp;
     m_currentPTS = buffer->timeStamp;
 
-    DEBUG("decode size %ld timeStamp %ld", m_stream->streamSize,
+    DEBUG("decode size %ld timeStamp %" PRIu64 "", m_stream->streamSize,
           m_stream->time_stamp);
 
     if (m_stream->streamSize < YamiParser::MPEG2::kStartCodeSize) {
--- a/v4l2/v4l2_decode.cpp
+++ b/v4l2/v4l2_decode.cpp
@@ -380,7 +380,7 @@ public:
         m_count = count;
         if (!count) {
             for (size_t i = 0; i < m_surfaces.size(); i++) {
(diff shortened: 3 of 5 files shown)


common/utils.cpp
```diff
@@ -74,8 +74,8 @@ bool guessResolution(const char* filename, int& w, int& h)
     int state = STATE_START;
     std::string strFilename(filename);
     const char* tokStart = NULL;
-    long width = 0;
-    long hight = 0;
+    long long width = 0;
+    long long hight = 0;
```
uartie: Line 97 and line 119 should use `std::strtoll`

uartie: ...and line 132 should use `long long maxInt`

common/utils_unittest.cpp
```diff
@@ -131,9 +131,9 @@ UTILS_TEST(guessResolutionOverflow) {
     EXPECT_EQ(h, 0);
 
     sstream.str("");
-    sstream << long(std::numeric_limits<int>::max()) * 2 + 3
+    sstream << long(std::numeric_limits<int>::max()) * 2ll + 3
             << "x"
-            << long(std::numeric_limits<int>::max()) * 3 + 2;
+            << long(std::numeric_limits<int>::max()) * 3ll + 2;
```
uartie: Lines 122 thru 127 should change `std::numeric_limits<long>::max()` to `std::numeric_limits<long long>::max()`

uartie (changes requested):

common/utils.cpp
xuguangxin: could we jst let width height as int? the width & heigh never greater than INT_MAX even in 32 bit system.

codecparsers/mpeg2_parser.cpp
```diff
@@ -255,8 +255,8 @@ namespace MPEG2 {
             return false;
 
 
-        DEBUG("slice header size                  : %ld",
-              m_slice.sliceHeaderSize);
+        DEBUG("slice header size                  : %lld",
```
xuguangxin: use [PRIu64 ](http://en.cppreference.com/w/cpp/types/integer)in string maybe better

decoder/vaapidecoder_mpeg2.cpp
```diff
@@ -134,7 +134,7 @@ VaapiDecoderMPEG2::DPB::insertPictureToReferences(const PicturePtr& picture)
 
 YamiStatus VaapiDecoderMPEG2::DPB::insertPicture(const PicturePtr& picture)
 {
-    INFO("insertPicture to DPB size %lu", m_referencePictures.size());
+    INFO("insertPicture to DPB size %lu", (unsigned long)m_referencePictures.size());
```
xuguangxin: you can use %zu. https://stackoverflow.com/questions/2524611/how-can-one-print-a-size-t-variable-portably-using-the-printf-family

v4l2/v4l2_decode.cpp
```diff
@@ -954,7 +954,7 @@ int32_t V4l2Decoder::onQueueBuffer(v4l2_buffer* buf)
     CHECK(buf);
     uint32_t type = buf->type;
     CHECK(type == V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE
-        || V4L2_BUF_TYPE_VIDEO_OUTPUT_MPLANE);
+        || (unsigned int)V4L2_BUF_TYPE_VIDEO_OUTPUT_MPLANE);
```
xuguangxin: is this check has logical error?
should be 

CHECK(type == V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE
-        || type == V4L2_BUF_TYPE_VIDEO_OUTPUT_MPLANE));

vaapi/vaapidisplay.cpp
```diff
@@ -304,7 +304,7 @@ DisplayPtr DisplayCache::createDisplay(const NativeDisplay& nativeDisplay)
     vaapiDisplay.reset();
 
     //crate new one
-    DEBUG("nativeDisplay: (type : %d), (handle : %ld)", nativeDisplay.type, nativeDisplay.handle);
+    DEBUG("nativeDisplay: (type : %d), (handle : %ld)", nativeDisplay.type, (long)nativeDisplay.handle);
```
xuguangxin: %p maybe better

v4l2/v4l2_decode.cpp
```diff
@@ -380,7 +380,7 @@ class ExternalBufferOutput : public V4l2Decoder::Output, BaseSurfaceAllocator {
         m_count = count;
         if (!count) {
             for (size_t i = 0; i < m_surfaces.size(); i++) {
-                if (m_surfaces[i] != VA_INVALID_SURFACE) {
+                if ((unsigned int)m_surfaces[i] != VA_INVALID_SURFACE) {
```
xuguangxin: could we convert VA_INVALID_SURFACE to intptr_t

xuguangxin (changes requested):

common/utils.cpp
uartie: @xuguangxin `long` was added originally in a088df6e6f762a27f7b13b8342260ad100920fbf to detect overflow of w,h output vars on 64 bit sys.

uartie: ...that is, if user provided filename has W or H > INT_MAX, then it is invalid.

xuguangxin: but for video encoder or vaCreateSurface, we can't create such surface. 16K may enough for many years. I mean no codec in near feature will support W or H > INT_MAX

uartie: Right.  But this function is only responsible for parsing the width and height from user specified filename string (no knowledge of va capabilities).  We can't assume user always specifies good input filename.  The `long long width` and `long long hight` are only temporary local variables to read the width and height from the filename string for validating.  Consider if we used std::atoi when parsing the width and height from filename string... then if width and height are > INT_MAX, the result would be undefined.  I suppose we could still have the same issue if width and height in filename is > maximum long long.  Nonetheless, I never really liked how this function was implemented and would be nice to refactor to a better solution.

xuguangxin: @uartie , if under layer returns an undefined value for this, no matter how to handle it in up layer. it still has the problem.
instead of use a088df6 maybe b3c3086582b695bd036628fc0f7dc260de10ba3d in #838 ,  it only change few lines and it can catch some exceptions. 
the target of this guessResolution function is provide some handle for us. 
When you do the vpp or decode, you get the yuv, but it's no place to get the original resolution. you will forget it in someday.  With this function, we can write the resolution down in file name. 
The technical in this function called [DFA](http://goldparser.org/articles/dfa.htm), most of the compiler use it to collect tokens

xuguangxin:
@wangzj0601 any update?thanks

xuguangxin:
@wangzj0601 , please help rebase this patch.
thanks

xuguangxin (approved):

xuguangxin:
hi @uartie ,
please help review this patch again.
thanks

xuguangxin:
@uartie , any update?
thanks

xuguangxin:
@uartie , any feedback?

xuguangxin:
@uartie , let us merge it first. We can send an additional request if we find a issue

PR #838, opened 2018-04-05
Fix guess resolution overflow

xuguangxin:
Instead of use a088df6, I think this fix may be better. 
1. It avoids change with and heigh to long long in https://github.com/intel/libyami/pull/836/files#r178894788. 
2. it's inplace operation, avoid the new allocation string object

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -27,6 +27,7 @@
 
 #include <assert.h>
 #include <ctype.h>
+#include <sstream>
 #include <string.h>
 #include <sys/time.h>
 #include <va/va.h>
@@ -59,6 +60,15 @@ uint32_t guessFourcc(const char* fileName)
     return YAMI_FOURCC_I420;
 }
 
+//do not use sscanf to get int, the function will make guessResolutionOverflow failed
+//see http://www.kumobius.com/2013/08/c-string-to-int/ for detials
+static bool getInt(const char* tokStart, int& v)
+{
+    std::istringstream iss(tokStart);
+    iss >> v;
+    return !iss.fail();
+}
+
 bool guessResolution(const char* filename, int& w, int& h)
 {
     enum {
@@ -86,7 +96,9 @@ bool guessResolution(const char* filename, int& w, int& h)
             {
                 if (*p == 'x' || *p == 'X') {
                     state = STATE_X;
-                    sscanf(tokStart, "%d", &w);
+                    if (!getInt(tokStart, w)) {
+                        state = STATE_START;
+                    }
                 } else if (!isdigit(*p)){
                     state = STATE_START;
                 }
@@ -106,7 +118,9 @@ bool guessResolution(const char* filename, int& w, int& h)
             {
                 if (!isdigit(*p)) {
                     state = STATE_END;
-                    sscanf(tokStart, "%d", &h);
+                    if (!getInt(tokStart, h)) {
+                        state = STATE_START;
+                    }
                 }
                 break;
             }
@@ -117,11 +131,10 @@ bool guessResolution(const char* filename, int& w, int& h)
     }
     //conner case
     if (*p == '\0' && state == STATE_HEIGHT) {
-        if (!isdigit(*p)) {
-            sscanf(tokStart, "%d", &h);
-        }
+        if (!getInt(tokStart, h))
+            return false;
     }
-    return w && h;
+    return w > 0 && h > 0;
 }
 
 struct ResolutionEntry {
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -27,12 +27,9 @@
 
 #include <assert.h>
 #include <ctype.h>
-#include <sstream>
 #include <string.h>
 #include <sys/time.h>
 #include <va/va.h>
-#include <cstdlib>
-#include <limits>
 
 namespace YamiMediaCodec{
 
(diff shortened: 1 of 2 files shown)


common/utils_unittest.cpp
```diff
@@ -127,17 +127,6 @@ UTILS_TEST(guessResolutionOverflow) {
             << std::numeric_limits<long>::max();
 
     EXPECT_FALSE(guessResolution(sstream.str().c_str(), w, h));
-    EXPECT_EQ(w, 0);
-    EXPECT_EQ(h, 0);
-
-    sstream.str("");
-    sstream << long(std::numeric_limits<int>::max()) * 2 + 3
-            << "x"
-            << long(std::numeric_limits<int>::max()) * 3 + 2;
```
uartie: This case is meant to test when an int overflow wraps back into a positive value.  On 64 bit system, `long(std::numeric_limits<int>::max()) * 2 + 3` == `4294967297` and `sscanf` in `guessResolution` would result in `1`.

Similarly, for `long(std::numeric_limits<int>::max()) * 3 + 2` == `6442450943` and `sscanf` would result in `2147483647`

Thus, consider the possibility that filename == `4294967297x4294967297`... `guessResolution` would return `true` and set the output vars `w=1` and `h=1`.  Such filename causes yamiencode to core dump.

xuguangxin: Hmm, I checked scanf return 1 for overflowed value. 
It's hard to handle this problem if scanf can't handle it well. 
Since long still have a problem on this. 
Any suggestion?

uartie: `std::istringstream` might be a good solution:
```
std::istringstream iss("6442450943");
int value(0);
iss >> value;
assert !iss.fail()
```

uartie: `std::stoi` in C++11 would be faster than `std::istringstream`

uartie: Here's a good overview of the pros and cons of different techniques: http://www.kumobius.com/2013/08/c-string-to-int/

uartie: For a simple/quick fix, we could just refactor like so:
```
diff --git a/common/utils.cpp b/common/utils.cpp
index 4aeb2da2989f..55c77314e678 100644
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -74,8 +74,8 @@ bool guessResolution(const char* filename, int& w, int& h)
     int state = STATE_START;
     std::string strFilename(filename);
     const char* tokStart = NULL;
-    long width = 0;
-    long hight = 0;
+    int width = 0;
+    int hight = 0;
 
     w = h = 0;
     uint32_t i = 0;
@@ -94,7 +94,10 @@ bool guessResolution(const char* filename, int& w, int& h)
                 if (strFilename[i] == 'x' || strFilename[i] == 'X') {
                     state = STATE_X;
                     strFilename[i] = ' ';
-                    width = std::strtol(tokStart, NULL, 10);
+                    std::istringstream iss(tokStart);
+                    iss >> width;
+                    if (iss.fail())
+                      return false;
                 }
                 else if (!isdigit(strFilename[i])) {
                     state = STATE_START;
@@ -116,7 +119,10 @@ bool guessResolution(const char* filename, int& w, int& h)
                 if (!isdigit(strFilename[i])) {
                     state = STATE_END;
                     strFilename[i] = ' ';
-                    hight = std::strtol(tokStart, NULL, 10);
+                    std::istringstream iss(tokStart);
+                    iss >> hight;
+                    if (iss.fail())
+                      return false;
                 }
                 break;
             }
@@ -126,19 +132,16 @@ bool guessResolution(const char* filename, int& w, int& h)
     }
 
     //conner case
-    if (strFilename.length() == i && state == STATE_HEIGHT)
-        hight = std::strtol(tokStart, NULL, 10);
-
-    long maxInt = std::numeric_limits<int>::max();
-    if (width > maxInt || hight > maxInt) {
-        w = 0;
-        h = 0;
-    }
-    else {
-        w = width;
-        h = hight;
+    if (strFilename.length() == i && state == STATE_HEIGHT) {
+        std::istringstream iss(tokStart);
+        iss >> hight;
+        if (iss.fail())
+          return false;
     }
 
+    w = width;
+    h = hight;
+
     return w && h;
 }
```

...but, maybe it's better to refactor entire function, instead, for better efficiency?  Your choice.

xuguangxin: thanks, @uartie , it's really a good lesson. get so much information from your link. efficiency is not a major problem for this function. it only uses one/two time for one round.

xuguangxin: changed. please help review
thanks

uartie: Since you fixed the overflow issue, please don't remove these checks in this unittest.  We still want to verify that passing > MAX_INT value returns false and does not regress in the future.  To fix 32-bit system where MAX_INT == MAX_LONG, we should use `long long` here instead (or any type that can hold > MAX_INT value).  Removing expectation that w,h == 0 is OK.

uartie (changes requested):

xuguangxin:
@uartie , Sorry it reverted by the c8069f8c2d04b6706b88e963b1694c5c5f339225. I added it back.

common/utils.cpp
```diff
@@ -62,49 +60,55 @@ uint32_t guessFourcc(const char* fileName)
     return YAMI_FOURCC_I420;
 }
 
+//do not use sscanf to get int, the function will make guessResolutionOverflow failed
+//see http://www.kumobius.com/2013/08/c-string-to-int/http://www.kumobius.com/2013/08/c-string-to-int/ for detials
```
uartie: Please fix URL in comment (it's here twice)

common/utils.cpp
```diff
@@ -62,49 +60,55 @@ uint32_t guessFourcc(const char* fileName)
...
+static bool getInt(const char* tokStart, int& v)
+{
+    std::istringstream iss(tokStart);
+    iss >> v;
+    if (iss.fail())
+        return false;
+    return true;
```
uartie: you can just `return !iss.fail()`

uartie (changes requested):

xuguangxin:
@uartie , done, you read code so carefully.
thanks

uartie (approved):

PR #840, opened 2018-04-11
fix coverity bug

wangzj0601:


decoder/DecoderApi_unittest.cpp
```diff
@@ -330,6 +331,8 @@ TEST_P(DecodeApiTest, Format_Change)
     EXPECT_TRUE(inFrames > 0);
     EXPECT_EQ((int)formats.size(), 1);
     EXPECT_EQ(inFrames, outFrames);
+    delete allocator;
```
xuguangxin: allocator is a local variable. nobody will use it after this point. No need set NULL to it

xuguangxin (changes requested):
please help remove allocator = NULL

wangzj0601:
@xuguangxin  it should has a memory leak after "assert", so I delete the variable at a wrong position before, and I change it like this now, what do you think about it?

xuguangxin:
this wired , the assert will abort the program, right? why we need handle memory leak at abort

decoder/DecoderApi_unittest.cpp
```diff
@@ -273,7 +273,10 @@ TEST_P(DecodeApiTest, Format_Change)
     SharedPtr<IVideoDecoder> decoder;
     TestDecodeFrames frames = *GetParam();
     decoder.reset(createVideoDecoder(frames.getMime()), releaseVideoDecoder);
-    ASSERT_TRUE(bool(decoder));
+    if (!bool(decoder)) {
+        delete allocator;
+        ASSERT_TRUE(bool(decoder));
+    }
```
uartie: It would be better to just declare and create the `allocator` after asserting the decoder... that way we can just keep it simple with the original `ASSERT_TRUE(bool(decoder))`.

Also, where does the `allocator` get deleted on successful test?  Does the `decoder` take ownership and delete it?

decoder/DecoderApi_unittest.cpp
```diff
@@ -342,7 +346,10 @@ TEST_P(DecodeApiTest, Flush)
...
```
uartie: see previous comment.

uartie (changes requested):

uartie:
> this wired , the assert will abort the program, right? why we need handle memory leak at abort

@xuguangxin the `ASSERT_TRUE` only aborts the test case, not the entire test program.

decoder/DecoderApi_unittest.cpp
wangzj0601: @uartie there is a line "m_externalAllocator.reset(allocator,unrefAllocator);", it should take ownership the variable, so the system will delete the memory after the tests complete.

xuguangxin (approved):

xuguangxin:
thanks U.Artie for explain

decoder/DecoderApi_unittest.cpp
```diff
@@ -342,15 +343,16 @@ TEST_P(DecodeApiTest, Flush)
     SharedPtr<IVideoDecoder> decoder;
     TestDecodeFrames frames = *GetParam();
     decoder.reset(createVideoDecoder(frames.getMime()), releaseVideoDecoder);
-    ASSERT_TRUE(bool(decoder));
 
     decoder->setAllocator(allocator);
+    ASSERT_TRUE(bool(decoder));
```
uartie: see other comment

decoder/DecoderApi_unittest.cpp
```diff
@@ -273,15 +273,16 @@ TEST_P(DecodeApiTest, Format_Change)
...
```
uartie: Still not quite right... this will crash if decoder is invalid (at `decoder->setAllocator(...)`).  The `allocator` should be created and passed to decoder **_after_** asserting the decoder is valid, like this:

```
decoder.reset(createVideoDecoder(frames.getMime()), releaseVideoDecoder);
ASSERT_TRUE(bool(decoder));
DecodeSurfaceAllocator* allocator = new DecodeSurfaceAllocator(display);
decoder->setAllocator(allocator);
```

uartie (changes requested):

decoder/DecoderApi_unittest.cpp
wangzj0601: yeah, you are right, it should be confirmed whether decoder is invalid or not, then we can use it.

uartie:
@wangzj0601 thanks for fixing and revising the contents of your patch.

However, the commit message is too vague and not useful.  It is not a "coverity bug"... it is a yami test bug.  Please describe in the commit message what it is you are actually doing to the code and why it is necessary.

uartie (approved):
LGTM. Thanks!

PR #843, opened 2018-04-28
fix warnings for v4l2 build.

xuguangxin:


interface/v4l2_wrapper.h
```diff
@@ -44,13 +44,18 @@ int32_t YamiV4L2_Munmap(void* addr, size_t length);
...
+int32_t YamiV4L2_SetXDisplay(int32_t fd, Display *x11Display);
+#endif //defined(__ENABLE_X11__)
+#endif //#else
+
+#if defined(__ENABLE_EGL__)
+int32_t YamiV4L2_UseEglImage(int fd, /*EGLDisplay*/void* eglDisplay, /*EGLContext*/void* eglContext, unsigned int buffer_index, void* egl_image);
 #endif
+
+int32_t YamiV4L2_SetDrmFd(int32_t fd, int drm_fd);
+
```
uartie: The ifdef logic here still does not match the v4l2/v4l2_wrapper.cpp ifdef logic.  That is, in CPP file, YamiV4L2_SetDrmFd and YamiV4L2_UseEglImage functions can only be defined if ```__ENABLE_WAYLAND__``` is not defined, yet they still get declared in header.

__HEADER FILE:__
```
#if defined(__ENABLE_WAYLAND__)
  ... YamiV4L2_SetWaylandDisplay ... // declaration iff __ENABLE_WAYLAND__
#else
  #if defined(__ENABLE_X11__)
     ... YamiV4L2_SetXDisplay ... // declaration iff !__ENABLE_WAYLAND__ and __ENABLE_X11__
  #endif //defined(__ENABLE_X11__)
#endif //#else

#if defined(__ENABLE_EGL__)
  ... YamiV4L2_UseEglImage ... // declaration iff __ENABLE_EGL__
#endif

... YamiV4L2_SetDrmFd ... // declaration ALWAYS 
```
versus

__CPP FILE__
```
#if defined(__ENABLE_WAYLAND__)
  ... YamiV4L2_SetWaylandDisplay ... // definition iff __ENABLE_WAYLAND__
#else
  #if defined(__ENABLE_X11__)
    ... YamiV4L2_SetXDisplay ... // definition iff !__ENABLE_WAYLAND__ && __ENABLE_X11__
  #endif
  #if defined(__ENABLE_EGL__)
    ... YamiV4L2_UseEglImage ... // definition iff !__ENABLE_WAYLAND__ && __ENABLE_EGL__
  #endif
  ... YamiV4L2_SetDrmFd ... // definition only iff !__ENABLE_WAYLAND__
#endif
```

Why don't we just remove the ```#else``` portion and put all `#ifdef` in their own conditional blocks? I mean, wayland and x11 can coexist, so why not the API's? ```v4l2_codecbase.cpp``` will also need fixed for ifdef logic.

uartie (changes requested):

interface/v4l2_wrapper.h
xuguangxin: good suggestion, updated.
Please also help review https://github.com/intel/libyami-utils/pull/124

uartie (approved):
LGTM

PR #848, opened 2018-05-11
enable lowlatency unittest which can get a output frame when give a i…

wangzj0601:
…nput frame immediately

### DIFF
--- a/decoder/DecoderApi_unittest.cpp
+++ b/decoder/DecoderApi_unittest.cpp
@@ -106,6 +106,49 @@ static FrameData g_jpegdata[] = {
     g_EOF,
 };
 
+static FrameData g_h264data_lowlatency[] = {
+    g_avc8x8I,
+    g_avc8x8P,
+    g_avc8x16,
+    g_avc16x16,
+    g_avc8x8I,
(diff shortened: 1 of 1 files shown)


decoder/DecoderApi_unittest.cpp
```diff
@@ -408,6 +456,40 @@ TEST_P(DecodeApiTest, Flush)
...
+
+    VideoDecodeBuffer buffer;
+    memset(&buffer, 0, sizeof(buffer));
+    FrameInfo info;
+    uint32_t inFrames = 0;
+    uint32_t outFrames = 0;
+    //keep previous resolution
+
+    while (frames.getFrame(buffer, info)) {
+        buffer.flag = 1;
```
xuguangxin: much better, only one thing here:
please use VIDEO_DECODE_BUFFER_FLAG_FRAME_END instead of [magic numbe](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)

xuguangxin (changes requested):

xuguangxin:
close this since it included in https://github.com/intel/libyami/pull/849

PR #849, opened 2018-05-15
Fix low latency

### LINKED ISSUE #844: Decoder 'drain' functionality
I'm trying to update from an older libyami and I used the drain parameter of decodeGetOutput.  Setting it to true would get me a frame out for every frame in.  Since that is removed, how do you get similar functionality?

xuguangxin:
two things done to fix low latency mode.
1. output output-able frame immediately when we add a frame to dpb. Do not check "current poc - previous poc == 1", this will introduce problem for poc "0, 2, 4, 6, 8"
2. add a flag VIDEO_DECODE_BUFFER_FLAG_FRAME_END to tell decoder to stop search next frame, so the decoder can decode current frame ASAP
this fixes https://github.com/intel/libyami/issues/844 and https://github.com/intel/libyami/issues/847

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,7 +8,7 @@ m4_define([yami_api_major_version], 0)
 # update this for every release when micro version large than zero
 m4_define([yami_api_minor_version], 5)
 # change this for any api change
-m4_define([yami_api_micro_version], 4)
+m4_define([yami_api_micro_version], 5)
 m4_define([yami_api_version],
     [yami_api_major_version.yami_api_minor_version.yami_api_micro_version])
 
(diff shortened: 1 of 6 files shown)


jsorg71:
It looks like h265 will have this feature too.  Is that correct?

xuguangxin:
@jsorg71 , h265 may not need this.
Bitstream will carry latency indication.  see [checkLatency ](https://github.com/intel/libyami/blob/apache/decoder/vaapidecoder_h265.cpp#L366). h264 does not have this, it only checks dpb [fullness](https://github.com/intel/libyami/blob/apache/decoder/vaapidecoder_h264.cpp#L1090). 
This is why we need a flag for h264.

YuJiankang (commented):
+1, lgtm

uartie:
status?

YuJiankang (approved):

xuguangxin:
Hi @uartie , sorry for late.

PR #858, opened 2018-06-11
add gcc compilation option for SDL325

wangzj0601:


uartie (changes requested):
Invoking gcc before `AC_PROC_CC` check is bad.  Also, compiler might not be gcc.

Even worse, this does not work on all distros/autoconf versions.  Something like `AX_CHECK_COMPILE_FLAG` would be better (https://www.gnu.org/software/autoconf-archive/ax_check_compile_flag.html).  Unfortunately, this `AX_CHECK_COMPILE_FLAG` m4 macro is not built-in and would require us to track it under project `m4` directory or force users to install autoconf-archive distro package.  I believe the license "exception" for that macro would allow us to track it in this project, but you would need to verify that first.  Or maybe you could find a built-in macro (try googling)?  Perhaps some form of `AC_COMPILE_IFELSE` usage would work.

wangzj0601:
I wrote these options into a variable and override it into libyami compilation

Makefile.am
```diff
@@ -37,11 +37,13 @@ libyami_la_LDFLAGS = \
 	$(LIBYAMI_LT_LDFLAGS) \
 	$(LIBVA_LIBS) \
 	$(LIBVA_DRM_LIBS) \
+	$(LDFLAGS) \
```
uartie, line 40: User LDFLAGS, CFLAGS, CXXFLAGS, CPPFLAGS etc. are automatically included already.

autogen.sh
```diff
@@ -22,8 +22,16 @@ else
     autoreconf -v --install || exit $?
 fi
 
+export CFLAGS="-D_FORTIFY_SOURCE=2 -fPIE -fPIC -fstack-protector"
+export CXXFLAGS="-D_FORTIFY_SOURCE=2 -fPIE -fPIC -fstack-protector"
+export LDFLAGS="-z noexecstack -z relro -z now -pie"
+
```
uartie, line 28: NEVER override user flags in toolchain files.

uartie (changes requested):
Please read and understand https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html

Also, why can't you just set custom user flags in your own environment instead?  See configure --help influential environment variables.

uartie:
Rejecting... user flags should be defined by user only.  User flags can negate AM_CFLAGS and AM_CXXFLAGS that are in configure.ac (e.g. `CFLAGS=-Wno-error CXXFLAGS=-Wno-error ./autogen.sh` to turn off AM_ setting of `-Werror`).

PR #859, opened 2018-06-13
Add flags for SDL signoff called SDL325 - Compile With Defenses Enabled

wangzj0601:
Due to the flags SDL required for Linux again, we need add these flags for SDL signoff called SDL325 - Compile With Defenses Enabled

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -405,8 +405,12 @@ if test "$enable_media_studio_va" = "yes"; then
         [Defined to 1 to enable H265 encoding when --enable-media-studio-va="yes"])
 fi
 
-AC_SUBST([AM_CFLAGS], ["-g -O2 -Wall -Wno-unused-function -Wno-cpp -Werror"])
-AC_SUBST([AM_CXXFLAGS], ["-g -O2 -Wall -Wno-unused-function -Wno-cpp -Wno-missing-braces -Werror"])
+AC_SUBST([AM_CFLAGS], ["-g -O2 -fPIE -fPIC -fstack-protector -Wall -Wno-unused-function -Wno-cpp -Werror"])
+AC_SUBST([AM_CXXFLAGS], ["-g -O2 -fPIE -fPIC -fstack-protector -Wall -Wno-unused-function -Wno-cpp -Wno-missing-braces -Werror"])
+
+#Add flags for SDL signoff called SDL325 - Compile With Defenses Enabled
+AC_SUBST([AM_CPPFLAGS], ["-D_FORTIFY_SOURCE=2"])
+AC_SUBST([AM_LDFLAGS], ["-z noexecstack -z relro -z now -pie"])
 
 # Checks for programs.
 AC_DISABLE_STATIC
--- a/gtestsrc/Makefile.am
+++ b/gtestsrc/Makefile.am
@@ -11,6 +11,7 @@ libgtest_la_SOURCES = \
 libgtest_la_CXXFLAGS = \
 	-I$(top_srcdir)/gtestsrc/gtest \
 	-I$(top_srcdir)/gtestsrc/gtest/include \
+	$(AM_CXXFLAGS) \
 	$(NULL)
 
 EXTRA_DIST = \


configure.ac
```diff
@@ -438,6 +438,12 @@ AC_TYPE_UINT8_T
 pkgconfigdir=${libdir}/pkgconfig
 AC_SUBST(pkgconfigdir)
 
+#Add flags for SDL signoff called SDL325 - Compile With Defenses Enabled
+SDL_CFLAGS="-D_FORTIFY_SOURCE=2 -fPIE -fPIC -fstack-protector"
+SDL_LDFLAGS="-z noexecstack -z relro -z now -pie"
+AC_SUBST(SDL_CFLAGS)
+AC_SUBST(SDL_LDFLAGS)
+
```
uartie: It would be better to just AC_SUBST the AM_CPPFLAGS and AM_LDFLAGS.  This way you don't have to modify any makefiles... all makefile targets will capture the flags automatically through AM_ flags (e.g. there are other makefiles/targets that you missed).

uartie (changes requested):

configure.ac
wangzj0601: As your requirement, I add these options into AM_flags, please review, thanks.

configure.ac
```diff
@@ -405,8 +405,8 @@ if test "$enable_media_studio_va" = "yes"; then
         [Defined to 1 to enable H265 encoding when --enable-media-studio-va="yes"])
 fi
 
-AC_SUBST([AM_CFLAGS], ["-g -O2 -Wall -Wno-unused-function -Wno-cpp -Werror"])
-AC_SUBST([AM_CXXFLAGS], ["-g -O2 -Wall -Wno-unused-function -Wno-cpp -Wno-missing-braces -Werror"])
+AC_SUBST([AM_CFLAGS], ["-g -O2 -Wall -Wno-unused-function -Wno-cpp -Werror -D_FORTIFY_SOURCE=2 -fPIE -fPIC -fstack-protector"])
+AC_SUBST([AM_CXXFLAGS], ["-g -O2 -Wall -Wno-unused-function -Wno-cpp -Wno-missing-braces -Werror -z noexecstack -z relro -z now -pie"])
```
uartie: I think you misunderstood me.  Please add AM_LDFLAGS and AM_CPPFLAGS with your new compiler options.  Don't append them to AM_CFLAGS and AM_CXXFLAGS... those are the wrong variables.

uartie (dismissed):

configure.ac
wangzj0601: Sorry for my mistake, I uploaded it wrongly and I have modified it just now.

uartie:
It's looking better.  But it appears someone forgot to add `$(AM_CXXFLAGS)` to the `libgtest_la_CXXFLAGS` target in `gtestsrc/Makefile.am`... can you add that, please, to silence the new warning about "`FORTIFY_SOURCE requires -O`" I'm seeing now?

Also, while at it, please use same format as we did for AM_CFLAGS substitution in configure.ac (i.e. use a one-liner AC_SUBST).

uartie (commented):
LGTM.  @xuguangxin can make the final call to approve/merge.

wangzj0601:
@uartie, the option "FORTIFY_SOURCE" does need "-O", thank you for your reply over and over again very much

xuguangxin:
@uartie ,thanks for reivew.
@wangzj0601 , why we need -O 2 bot in CPP_FLAGS AND CXX_FLAGS.
I remember CPP is for some process, not compile flag,  why it needs -O 2.

wangzj0601:
This -O should be for "-D_FORTIFY_SOURCE=2", only -O is set to 2 or higher, then the option "FORTIFY_SOURCE" can be activated.

xuguangxin:
will this –O2 covered by CPP_FLAGS?
If someone need define –O 0 for debug. Will this build failed?

thanks

wangzj0601:
This -O2 and CPPFLAGS will be compiled together.
For debug, you can modify the makefile and change O2 to O0, it will not build failed.

uartie:
CXXFLAGS get applied to c++ targets.  CFLAGS get applied to c targets.  CPPFLAGS get applied to both c and c++ targets.  Thus, -O in all of these is redundant.

uartie:
Perhaps we should just consolidate all common flags into CPPFLAGS

uartie:
The implicit compiler command in make for c targets is essentially `$(CC) $(AM_CPPFLAGS) $(CPPFLAGS) $(AM_CFLAGS) $(CFLAGS) ...`.  Likewise, for c++ targets.  For the compiler, the last option seen overrides any previous option.  Therefore, if user wants to use `-O0` instead of `-O2`, they can set user `CFLAGS="-O0"` and `CXXFLAGS="-O0"` in environment to override the project AM_ flags.  There is no need for user to modify the makefile or configure.ac by hand.

uartie:
One more lesson ;-) ...

Technically, `CPPFLAGS` are for preprocessor and `CFLAGS/CXXFLAGS` are for compiler.  But in terms of the way `make` assembles the final command, it doesn't make much difference (mostly).

uartie:
...the caveat: `CFLAGS/CXXFLAGS` get applied by `make` during link phase, too.  But, the `CPPFLAGS` do not get applied by `make` during link phase.  Therefore, it would be better to put flags into the proper places.  (see https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html for valid preprocessor flags)

configure.ac
```diff
@@ -408,6 +408,10 @@ fi
 AC_SUBST([AM_CFLAGS], ["-g -O2 -Wall -Wno-unused-function -Wno-cpp -Werror"])
 AC_SUBST([AM_CXXFLAGS], ["-g -O2 -Wall -Wno-unused-function -Wno-cpp -Wno-missing-braces -Werror"])
 
+#Add flags for SDL signoff called SDL325 - Compile With Defenses Enabled
+AC_SUBST([AM_CPPFLAGS], ["-O2 -D_FORTIFY_SOURCE=2 -fPIE -fPIC -fstack-protector"])
```
uartie: `-O2 -fPIE -fPIC -fstack-protector` belong in `AM_CFLAGS` and `AM_CXXFLAGS`

wangzj0601: I have moved these flags into AM_CFLAGS and AM_CXXFLAGS, is it OK?

uartie:
LGTM

xuguangxin (approved):
thanks U.Artie for detailed review

PR #863, opened 2018-08-24
Fix compile issue for Android Celadon project

xuguangxin:
Celadon use clang compiler, the compiler command is strict than gcc default.
we use this serials patches to fix the compile issue on Celadon.

More detail about Celadon you can find on
https://github.com/projectceladon/manifest/wiki


YuJiankang (approved):
lgtm, +1

PR #866, opened 2018-11-10
add two advanced deinterlace mode motion_adaptive and motion_compensated

xuguangxin:


xuguangxin:
@uartie , please help review this.
I found the media-driver and intel-vappi-driver have different behaviors.
media-driver will fallback to spacial deinterlace if we just give him one frame.
Intel -vaapi driver will simply silent failed with no error returned since he needs a forward reference frame
My first patch only tested on media-driver. 
so I write a new one. it set forward reference frame.  should work on both drivers. 
thanks

uartie:
@xuguangxin please fix travis-ci build first.

xuguangxin:
@uartie , done, thanks for reminding.

uartie (approved):
LGTM

PR #868, opened 2018-11-16
add conner test case for YamiParser::NalReader::getPos, fix the issue

xuguangxin:
getPos is complex when we have epb, if we do not reach the epb,
the getPos does not need to include it. 
The first commit will expose the following issue.
The second, commit fixed it

    [ RUN      ] NalReaderTest.GetPosForEPB
    nalReader_unittest.cpp:97: Failure
          Expected: 1u
          Which is: 1
    To be equal to: r.getPos()
          Which is: 9
    nalReader_unittest.cpp:100: Failure
          Expected: 8u
          Which is: 8
    To be equal to: r.getPos()
          Which is: 16
    [  FAILED  ] NalReaderTest.GetPosForEPB (0 ms)

### DIFF
--- a/codecparsers/nalReader.cpp
+++ b/codecparsers/nalReader.cpp
@@ -143,4 +143,20 @@ void NalReader::rbspTrailingBits()
         skip(1); /*rbsp_alignment_zero_bit, equal to 0*/
 }
 
+uint64_t NalReader::getPos() const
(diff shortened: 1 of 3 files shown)


wangzj0601 (approved):
I approve this pull request, thanks

PR #869, opened 2018-12-03
refact vc1 parser to make code more read able.

xuguangxin:
Made following improvement:
1. remove memory copy for rbdu, we only need few headers. And hw can handle emulation bytes by him self.Copy and remove emulation byte in the entire stream is unnecessary. 
2. isolate dpb and reference management logical to DPB class. it will make the code more readable..
3. introduce RDBU class, to hide rdbu parser details.


xuguangxin:
I will rebase this for new Nal reader changes

PR #871, opened 2018-12-26
Remove NalReader::getEpbCnt() to simply the error prone code

xuguangxin:


wangzj0601 (approved):
all tests are same with previous results and has no regression issues, approve this patch.

